PR #41, opened 2023-07-21
fix: use alphabetical as the default sort type in schemas

haocheng6:
### Description

Currently, the default `type` specified in the schema of every rule is `SortType.natural`. This will cause inconsistent behavior. For example,  `"perfectionist/sort-array-includes": "error"` and `"perfectionist/sort-array-includes": ["error", {}]` don't work in the same way. The first one uses alphabetical sorting while the second uses natural sorting.

This PR fixes the default `type` for all rules.

### Additional context

N/A

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/rules/sort-array-includes.ts
+++ b/rules/sort-array-includes.ts
@@ -46,7 +46,7 @@ export default createEslintRule<Options, MESSAGE_ID>({
               SortType.natural,
               SortType['line-length'],
             ],
-            default: SortType.natural,
+            default: SortType.alphabetical,
           },
           order: {
             enum: [SortOrder.asc, SortOrder.desc],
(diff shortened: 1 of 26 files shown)


azat-io:
Nice catch!
Do you mind if I combine this but create a new release a little later?

haocheng6:
@azat-io No, not at all. I don't think this bug will affect too many users if they just use the config on the documentation site. Also, alphabetical sorting and natural sorting produce the same result in most real-world use cases. So it is ok to release the fix later.

Also, thanks for taking the time to review the PR.

azat-io:
Released in [v1.5.1](https://github.com/azat-io/eslint-plugin-perfectionist/releases/tag/v1.5.1).

Thank you!

PR #57, opened 2023-08-14
fix: handle more complex expressions in sort-objects

stovmascript:
### Description

3e987ae7e6fa9fc21caa7e23fb8b17d9ece027f0 fixes references to dependencies/earlier params (#9), but not in some more complex cases. Examples are provided in tests.

### Additional context

Do you think this is a good approach? I think we should try to search all of the nested nodes for `Identifier`s, which would be pushed to the `dependencies` array. I'm not sure I covered all the possible cases with this approach.

The tests also probably need some more work on the semantics :)

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/rules/sort-objects.ts
+++ b/rules/sort-objects.ts
@@ -158,7 +158,7 @@ export default createEslintRule<Options, MESSAGE_ID>({
 
               if (prop.value.type === AST_NODE_TYPES.AssignmentPattern) {
                 let addDependencies = (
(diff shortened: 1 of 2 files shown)


stovmascript:
@azat-io Wow, that was fast :) I'll try to follow up with a PR adding the missing test cases and possibly doing something about making this auto fixable.

PR #69, opened 2023-09-08
fix: prefix-only builtin modules are defined as external modules in sort-imports rule

### LINKED ISSUE #65: Bug: (sort-imports) prefix-only builtin modules are defined as external modules
### Describe the bug

When using the `node:` prefix for an `builtin` module, the `sort-imports` rule defines it as an `external` module.

### Code example

**IN CONFIG:**
```json
"perfectionist/sort-imports": [
  "error",
  {
    "type": "natural",
    "groups": [
      "builtin",
      "external",
      "internal",
      "parent",
      "sibling",
      "index",
      "side-effect"
    ]
  }
]
```

**IN CODE:**
```js
import { writeFile } from "node:fs/promises"; // ⤸ There is no space between groups.
import { createRoot } from "react-dom/client";
```

### ESLint version

v8.48.0

v2.0.0

### Additional comments

_No response_

### Validations

Wondermarin:
### Description

Fix #65

### Additional context

Because `builtinModules` does not includes modules that start with the prefix `node:` you need to check it manually.

---

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/rules/sort-imports.ts
+++ b/rules/sort-imports.ts
@@ -201,7 +201,9 @@ export default createEslintRule<Options<string[]>, MESSAGE_ID>({
         tsPaths.some(pattern => minimatch(nodeElement.source.value, pattern))
 
       let isCoreModule = (value: string) =>
-        builtinModules.includes(value.startsWith('node:') ? value.split('node:')[1] : value)
+        builtinModules.includes(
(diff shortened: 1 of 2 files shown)


azat-io:
LGTM.
Can you fix formatter issues?

Wondermarin:
It should work now

Issue #83, opened 2023-10-18
Feature: (option for line-length to sort imports by their last line, instead of their total characters)


tthornton3-chwy:
### What rule do you want to change?

perfectionist/sort-imports

### Describe the problem

Good morning! Pardon me if this is already addressed, I took a peek and couldn't find an option / setting that let me do this.

It would be nice if typescript named imports that have a _really_ long type list, and wind up each being on their own line, would actually be at the beginning / factoring in that it's last line is actually short. Otherwise the nice 'wave' of imports looks a bit off, with what visually looks to be a short line at the end. 

Thank you so much for developing this! It works really great, and I can tell a lot of care went into it.

I used to use https://github.com/renke/import-sort with my own custom ordering via plugin
<details>
  <summary>That ordering for reference</summary>
  
  ```js
import {
  ISorterFunction,
  IStyleAPI,
  IStyleItem,
} from "import-sort-style";

import { IImport } from "import-sort-parser";

const length: ISorterFunction = (i1, i2) => {
  let i1length = i1.end - i1.start;
  if (i1length > 80 && i1.namedMembers.length > 0) {
    i1length = i1.moduleName.length + 10;
  }
  let i2length = i2.end - i2.start;
  if (i2length > 80 && i2.namedMembers.length > 0) {
    i2length = i2.moduleName.length + 10;
  }
  const diff = i1length - i2length;
  return diff / Math.abs(diff);
};

function isScopedModule(imported: IImport) {
  return imported.moduleName[0] === "@";
}

function isReact(imported: IImport) {
  return imported.moduleName === "react";
}

export default function (styleApi: IStyleAPI, file: string): IStyleItem[] {
  const {
    alias,
    and,
    not,
    hasNoMember,
    isAbsoluteModule,
    isInstalledModule,
    isNodeModule,
    isRelativeModule,
    unicode,
  } = styleApi;

  return [
    // import "foo"
    {
      match: and(hasNoMember, isAbsoluteModule),
      sort: length,
    },
    { separator: true },
    // import "./foo"
    {
      match: and(hasNoMember, isRelativeModule),
      sort: length
    },
    { separator: true },
    // import … from "fs";
    {
      match: isNodeModule,
      sort: [length],
      sortNamedMembers: alias(unicode),
    },
    { separator: true },
    // import … from "react";
    {
      match: isReact,
      sort: [length],
      sortNamedMembers: alias(unicode),
    },
    // import … from "foo";
    {
[...]

PR #85, opened 2023-10-27
feat(sort-imports): max line length

tthornton3-chwy:
### Description
Closes https://github.com/azat-io/eslint-plugin-perfectionist/issues/83 ! 

Gives support for sorting imports when they are split over multiple lines, thanks to prettier, for example.

### Additional context

If you have any questions or concerns, let me know! I made sure coverage remained at 100%! I wish testing JSON Schemas were easier / part of the plugin, but not yet :(

---

- [ ] Bug fix
- [X] New Feature
- [ ] Documentation update
- [ ] Other


.gitignore
```diff
@@ -11,6 +12,7 @@ dist/
 # VitePress
 docs/.vitepress/cache/
 docs/.vitepress/temp/
+.vitepress/
```
azat-io, line 15: Why?

tthornton3-chwy: I think my IDE or some process put this here-- there was a .vitepress/ folder at the root, with only a `caches` or something like that folder inside. To minimize changes here i just deleted that folder, removed this line so should be all set!

test/utils/areOptionsValid.ts
```diff
@@ -0,0 +1,60 @@
+import type { JSONSchema4 } from '@typescript-eslint/utils/json-schema'
```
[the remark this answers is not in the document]
tthornton3-chwy: Deal! Testing json schema will _eventually_ be a part of the eslint plugin, but for now I just "borrowed" their example method from how their internal team does it itself for now, with the appropriate changes to work for us, which is why the name. But I have now renamed to match this repos style, so should be set!

tthornton3-chwy:
Feedback should be addressed!

azat-io:
Sorry for long response.

Thank you! I'll create a new release soon

azat-io:
Released in [v2.3.0](https://github.com/azat-io/eslint-plugin-perfectionist/releases/tag/v2.3.0).

tthornton3-chwy:
Awesome, thanks for looking at it and getting that in and out @azat-io! Have a good week, and thanks for providing this for free and for everyone.

PR #117, opened 2024-03-18
feat(sort-interfaces): add `optionalityOrder` option

chirokas:
### Description

This PR adds the `optionalityOrder` option to the `sort-interfaces` rule.

This option has three possible values: `ignore`, `optional-first`, `required-first`.

If the `required-first` option is selected, then before any sorting/grouping configurations, we ensure that the members array has all its required elements first and all its optional elements second. We then perform the remaining sorting/grouping checks on the required subset and the optional subset.

### Additional context

```json
// .eslintrc.json
{
  "custom-groups": {
    "callback": "on*"
  },
  "groups": ["unknown", "callback"],
  "optionalityOrder": "required-first"
}
```

❌ Incorrect

```ts
interface ButtonProps {
  backgroundColor?: string
  label: string
  primary?: boolean
  size?: 'large' | 'medium' | 'small'
  onClick?(): void
}
```

✅ Correct

```ts
interface ButtonProps {
  label: string
  backgroundColor?: string
  primary?: boolean
  size?: 'large' | 'medium' | 'small'
  onClick?(): void
}
```

---

- [ ] Bug fix
- [X] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/docs/rules/sort-interfaces.md
+++ b/docs/rules/sort-interfaces.md
@@ -100,6 +100,7 @@ type Group = 'multiline' | CustomGroup
 
 interface Options {
   type?: 'alphabetical' | 'natural' | 'line-length'
+  optionalityOrder?: 'ignore' | 'optional-first' | 'required-first'
   order?: 'asc' | 'desc'
   'ignore-case'?: boolean
   groups?: (Group | Group[])[]
@@ -116,6 +117,13 @@ interface Options {
 - `alphabetical` - sort alphabetically.
 - `natural` - sort in natural order.
 - `line-length` - sort by code line length.
+
+### optionalityOrder
+
+<sub>(default: `'ignore'`)</sub>
+
+- `optional-first` - put all optional members first.
+- `required-first` - put all required members first.
 
 ### order
 
--- a/rules/sort-interfaces.ts
+++ b/rules/sort-interfaces.ts
@@ -2,13 +2,14 @@ import { minimatch } from 'minimatch'
 
 import type { SortingNode } from '../typings'
 
+import { OptionalityOrder, SortOrder, SortType } from '../typings'
 import { createEslintRule } from '../utils/create-eslint-rule'
+import { isMemberOptional } from '../utils/is-member-optional'
 import { getLinesBetween } from '../utils/get-lines-between'
 import { getGroupNumber } from '../utils/get-group-number'
 import { toSingleLine } from '../utils/to-single-line'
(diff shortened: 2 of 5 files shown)


azat-io:
Thanks for the PR. It's a really good and needed feature.
What do you think about using `groups`? Or do you think it would be better to implement a new `optionalityOrder`?

chirokas:
> Thanks for the PR. It's a really good and needed feature. What do you think about using `groups`? Or do you think it would be better to implement a new `optionalityOrder`?

If you use `groups`, subsets of optional/required groups will overlap with subsets of other groups. So I think it would be better to implement a new `optionalityOrder`.

chirokas:
> Thanks for the PR. It's a really good and needed feature. What do you think about using `groups`? Or do you think it would be better to implement a new `optionalityOrder`?

I think it would be better to implement a new `optionalityOrder`.

**Example :**

```json
// .eslintrc.json
{
  "custom-groups": {
    "callback": "on*"
  },
  "groups": ["optional", "callback"]
}
```

```ts
interface ButtonProps {
  backgroundColor?: string;
  label: string;
  primary?: boolean;
  size?: "large" | "medium" | "small";
  onClick?(): void;
}
```

You are given two groups `optional` and `callback`.
Imagine that when you use `defineGroup`, some nodes of the two groups are overlapped. If two nodes overlap, no comparison is possible.

azat-io:
Sorry for the long reply.

Okay, I agree with you.
And what can you say about group priority and optionalityOrder?

For example, these are my settings:

```
{
  "rules": {
    "perfectionist/sort-interfaces": [
      "error",
      {
        "type": "natural",
        "order": "asc",
        "custom-groups": {
          "name": "name"
        },
        "groups": ["name", "unknown"],
        "optionalityOrder": "required-first"
      }
    ]
  }
}
```

How would that work here?

```
interface Props {
  name?: string
  age: number
}
```

azat-io:
Another issue is the naming convention of options.

You're using camelCase in `optionalityOrder `. But we use kebab-case in `'custom-groups'`.

chirokas:
In my plan, it performs the following steps when called:

1. Perform _sorting/grouping_ on a required subset and an optional subset.
2. Put all required members first (or all optional members first).

Correct:

```ts
interface Props {
  age: number
  name?: string
}
```

azat-io:
Okay, thank you!

PR #123, opened 2024-04-07
feat: add sort-intersection-types rule

### LINKED ISSUE #84: Feature: Add a sort-intersection-types rule?
### Describe the rule

https://eslint-plugin-perfectionist.azat.io/rules/sort-union-types exists for _union_ types, but what about _intersection_ types?

### Code example

```ts
// ❌ Incorrect
type Both =
  & { b: 1 }
  & { a: 1 }

// ✅ Correct
type Both =
  & { a: 1 }
  & { b: 1 }
```

### Additional comments

We intentionally named https://typescript-eslint.io/rules/sort-type-constituents as _"type constituents"_ because that refers to the constituents in both intersection and union types. Fun fact!

### Validations

JoshuaKGoldberg:
### Description

Fixes #84.

I copy & pasted the existing `sort-union-types` rule, find-and-replaced _"union"_ to _"intersection"_, then updated tests to use `&` intersection types.

### Additional context

I didn't add the `nullable-last` option because I figured that's not a common scenario with intersection types. Maybe I'm wrong there?

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/docs/rules/sort-intersection-types.md
+++ b/docs/rules/sort-intersection-types.md
@@ -1,0 +1,148 @@
+---
+title: sort-intersection-types
+description: ESLint Plugin Perfectionist rule which enforce sorted intersection types in TypeScript
+---
+
+# sort-intersection-types
+
+💼 This rule is enabled in the following [configs](/configs/): `recommended-alphabetical`, `recommended-line-length`, `recommended-natural`.
+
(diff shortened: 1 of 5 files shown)


azat-io:
Looks awesome!
Thank you! ❤️

PR #124, opened 2024-04-08
fix(sort-exports): work with star exports

### LINKED ISSUE #119: Bug: (sort-exports) does not work for 'export *'
### Describe the bug

sort-exports rule does not work for 'export *'

```js
// .eslintrc.cjs
'perfectionist/sort-exports': [
  'error',
  {
    'type': 'alphabetical',
    'order': 'asc',
  },
],
```

### Code example

```js
export * from './b'
export * from './a' // <- There must be a error here
```

### ESLint version

8.57.0

2.6.0

### Additional comments

_No response_

### Validations

chirokas:
### Description

Fixes #119 .

### Additional context

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/rules/sort-exports.ts
+++ b/rules/sort-exports.ts
@@ -82,15 +82,11 @@ export default createEslintRule<Options, MESSAGE_ID>({
         | TSESTree.ExportNamedDeclarationWithSource
         | TSESTree.ExportAllDeclaration,
     ) => {
-      if (node.type === 'ExportAllDeclaration' && node.exported === null) {
-        parts.push([])
-      } else {
-        parts.at(-1)!.push({
-          size: rangeToDiff(node.range),
-          name: node.source.value,
-          node,
-        })
-      }
+      parts.at(-1)!.push({
+        size: rangeToDiff(node.range),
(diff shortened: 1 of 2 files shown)


azat-io:
I think we should add a check that these are TS or TSX files.
For javascript files, this behavior looks unsafe.

chirokas:
This has nothing to do with `sort-exports`.

If there are two wildcard exports statements that implicitly re-export the same name, neither one is re-exported.

```js
// -- mod1.js --
export const a = 1

// -- mod2.js --
export const a = 3

// -- barrel.js --
export * from './mod1.js'
export * from './mod2.js'

// -- main.js --
import * as ns from './barrel.js'
console.log(ns.a) // undefined
```

Attempting to import the duplicate name directly will throw an error.

```js
import { a } from './barrel.js'
// Uncaught SyntaxError: The requested module './barrel.js' contains conflicting star exports for name 'a'
```

azat-io:
Okay. Thank you!

PR #129, opened 2024-04-15
fix(sort-imports): empty named imports being considered side-effect imports

hampus-stravito:
### Description

I noticed a bug when writing some code in a project where I have set up this plugin (thank you by the way, it helps out a lot!). When I wrote a named import but didn't specify anything to be imported it was being considered a side-effect import and moved to the bottom of the list of imports. Turns out there is no way to distinguish a side-effect import from an empty named import from looking at the AST, so the only thing I could think of what to look for a `} from` string in the nodes code to make it not being considered a side-effect import 🙂

### Additional context

Output from [AST explorer](https://astexplorer.net/#/gist/090f7d39fb3be4d37c64d13fccc31156/210c0b4ea656f0eee3406d52b63937f5fc34754f)

ast explorer output

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/rules/sort-imports.ts
+++ b/rules/sort-imports.ts
@@ -217,7 +217,10 @@ export default createEslintRule<Options<string[]>, MESSAGE_ID>({
     let nodes: SortingNode[] = []
 
     let isSideEffectImport = (node: TSESTree.Node) =>
-      node.type === 'ImportDeclaration' && node.specifiers.length === 0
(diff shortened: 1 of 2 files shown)


azat-io:
Released in [v2.10.0](https://github.com/azat-io/eslint-plugin-perfectionist/releases/tag/v2.10.0)

PR #131, opened 2024-04-21
feat: add custom ignore to enable customizable ignore rule

### LINKED ISSUE #126: Feature: (sort-objects) option for destructuring only

sort-objects

### Describe the problem

We have too many objects that we don't want sorted (e.g. XState machines where properties are grouped logically) to make sort-objects with ignore-pattern and/or ESLint overrides feasible. It would be nice to still have object destructuring available, particularly for React components where currently the props' type get sorted but the prop destructuring does not.

### Code example

Input:
```js
const data = {
  software: 'iOS',
  name: 'iPhone 14 Pro',
};
function Product({ software, name }) { ... }
```
Expected output:
```js
const data = {
  software: 'iOS',
  name: 'iPhone 14 Pro',
};
function Product({ name, software }) { ... }
```

### Additional comments

Not sure if it should be an option to sort-objects or a new/split rule. https://github.com/mskelton/eslint-plugin-sort has separate rules for destructuring-properties and object-properties.

### Validations

KID-joker:
### Description

#126

### Additional context

Because `json-schema` does not support `function` type, it is converted through `object`.

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/docs/content/rules/sort-objects.mdx
+++ b/docs/content/rules/sort-objects.mdx
@@ -216,14 +216,6 @@ Example:
 
 When `false`, this rule will be disabled for the styled-components like libraries.
 
-### ignorePattern
-
-<sub>(default: `[]`)</sub>
-
-If you need to ignore a rule for some interfaces, you can specify their names or a pattern to ignore, for example: `'Component*'` to ignore all interfaces whose names begin with the word Component.
-
-The [minimatch](https://github.com/isaacs/minimatch) library is used for pattern matching.
-
 ### partitionByComment
 
 <sub>(default: `false`)</sub>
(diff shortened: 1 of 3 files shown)


azat-io:
The idea for such an option is a good one. But don't you think the customization is overly complicated? Perhaps it would be better to use just variable names instead of functions?

KID-joker:
I don't quite understand what you mean. How do variable names solve the problem?

rtatarinov:
@KID-joker does it allow to exclude the files with specific name? can u prodive an example for it?

PR #160, opened 2024-06-30
fix(sort-array-includes.md): syntax json `spread-last` => `"spread-la…

INeedJobToStartWork:
…st"` etc

### Description

 Fix Json Syntax error (no quotes in keys which include "-" char)
 
 It wasn't ready to ctrl + c & ctrl + v

### Additional context

-

---

- [ ] Bug fix
- [ ] New Feature
- [X] Documentation update
- [ ] Other

### DIFF
--- a/docs/rules/sort-array-includes.md
+++ b/docs/rules/sort-array-includes.md
@@ -148,7 +148,7 @@ export default [
         {
           type: 'natural',
           order: 'asc',
-          spread-last: true,
+          'spread-last': true,
         },
       ],
     },


docs/rules/sort-array-includes.md
```diff
@@ -146,9 +146,9 @@ export default [
       'perfectionist/sort-array-includes': [
         'error',
         {
-          type: 'natural',
-          order: 'asc',
-          spread-last: true,
+          "type": 'natural',
+          "order": 'asc',
+          "spread-last": true,
```
azat-io, line 151: Could you fix formatter issue?
In the case of `type` and `order` we don't need quotes, in the case of `spread-last` the quotes should be single quotes.

azat-io (changes requested):

docs/rules/sort-array-includes.md
INeedJobToStartWork: Now should be fine.

Issue #193, opened 2024-08-04
Feature: Add `readonly` properties to perfectionist/sort-classes


christopher-buss:
### Describe the rule

I would like the option to group by readonly properties.

I used to use the following extension to organize variables, but I found this to be overall an issue as I couldn't run it through ESLint, therefore it couldn't be run in CI, or other team members wouldn't always know to format.
https://github.com/aljazsim/vs-code-typescript-class-organizer/blob/master/src/member-type.ts

I'd basically like to try get as close to this ordering as possible through ESLint, and readonly properties are one of the main constraints here.

### Code example

Currently the ordering is like this:
```ts
private readonly runway = 5;
private sParticipants = new Array<Player>();
private readonly voted = new Set<Player>();
```

But I'd like them to be able to be grouped as such:
```ts
private readonly runway = 5;
private readonly voted = new Set<Player>();

private sParticipants = new Array<PlayerEntity>();
```

### Additional comments

I realise I may be able to achieve this with custom groups already, but I believe readonly would make sense as a builtin. The original ESLint [rule](https://typescript-eslint.io/rules/member-ordering/#member-types-granular-form) (which had no autofix) supported readonly properties.

### Validations

- [X] Read the [docs](https://perfectionist.dev).
- [X] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

azat-io:
Added in baa701d494baa8d01f977a650080fb52c61dbc46.
Released in [v3.2.0](https://github.com/azat-io/eslint-plugin-perfectionist/releases/tag/v3.2.0).

Issue #190, opened 2024-08-03
Feature: Add `abstract` attributes and methods to `perfectionist/sort-classes` 


hugoAlt:
### What rule do you want to change?

`sort-classes`

### Describe the problem

I personally enjoy having abstract attributes and methods at the top of my classes. I believe that `unknown` groups matches them, but think it would be interesting to add them separately. Same for methods.

More generally speaking, would it not be better to have a system similar to `eslint/naming-convention` where users can define `selectors` and `modifiers` for each, rather than enumerating the list of possible selectors? Complex elements like `private static properties` can not be individually matched currently. 

### Code example

```typescript
abstract class TestClass {

  // ...
  
  
  // Sort me please
  abstract attribute: string;

  // Me too
  abstract method(): void;
}
```

### Validations

- [X] Read the [docs](https://perfectionist.dev).
- [X] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

hugoAlt:
@azat-io I can start working on a PR for this if this is fine for you.

azat-io:
@hugoAlt Sure. That would be great 👍

azat-io:
Added in baa701d494baa8d01f977a650080fb52c61dbc46.
Released in [v3.2.0](https://github.com/azat-io/eslint-plugin-perfectionist/releases/tag/v3.2.0).

PR #194, opened 2024-08-04
feat: adding abstract, readonly and additional public/protected/private groups

hugop95:
Added the additional groups:

`'public-abstract-property'`
`'protected-abstract-property`'
`'abstract-property'`
`'public-abstract-method'`
`'protected-abstract-method'`
`'abstract-method'`
`'static-public-property'`
`'static-private-property'`
`'static-public-method'`
`'public-constructor'`
`'protected-constructor'`
`'private-constructor'`
`'public-property'`
`'public-method'`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 # Node.js
 node_modules/
 .nvmrc
-package-lock.json
 
 # Build
 dist/
@@ -9,6 +8,7 @@ dist/
 # Editor
 .vscode/
 .vim/
+.idea
 
 # Astro
 .astro/
@@ -18,6 +18,3 @@ coverage/
 
 # Misc
 .DS_Store
-
-# Jetbrains IDE
-.idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Node.js
 node_modules/
 .nvmrc
+package-lock.json
 
 # Build
 dist/
@@ -17,3 +18,6 @@ coverage/
 
 # Misc
 .DS_Store
+
+# Jetbrains IDE
+.idea
--- a/rules/sort-classes.ts
+++ b/rules/sort-classes.ts
@@ -1,4 +1,3 @@
-import type { ClassElement } from 'node_modules/@typescript-eslint/types/dist/generated/ast-spec'
 import type { TSESTree } from '@typescript-eslint/types'
 import type { TSESLint } from '@typescript-eslint/utils'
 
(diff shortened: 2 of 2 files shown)


hugop95:
@azat-io As mentioned in https://github.com/azat-io/eslint-plugin-perfectionist/issues/190,
I am adding `abstract` groups as well as some other missing groups that I think would be useful. We could also add `readonly` as per https://github.com/azat-io/eslint-plugin-perfectionist/issues/193.

Do you think the number of official groups would be too much?

.gitignore
```diff
@@ -17,3 +18,6 @@ coverage/
 
 # Misc
 .DS_Store
+
+# Jetbrains IDE
+.idea
```
azat-io, line 23: Could you add `.idea` to `# Editor` block?

.gitignore
```diff
@@ -1,6 +1,7 @@
 # Node.js
 node_modules/
 .nvmrc
+package-lock.json
```
azat-io, line 4: We use pnpm and we don't need to ignore the lock file for npm

azat-io (commented):
Thank you for your pull request! It looks good.

Could you please also fix the following issues:

1. Add new groups to docs (`docs/content/rules/sort-classes.mdx`)
2. Fix ESLint issues
3. Fix Unit tests
4. Add tests for new groups. We want to have 100% code coverage with tests. To run tests, use the `pnpm run test:unit` command.

hugop95:
@azat-io The number of conditional branches is growing exponentially as we allow new modifiers to be added (in this case, `abstract` and `readonly`. We increase that number even more if we choose to allow for other keywords like `override` or `async` in the future.

What do you think about instead dynamically building a list of `modifiers`, and then constructing the list of groups associated?

Example:

`private abstract get a() {...`
Should generate the modifiers `['private', 'abstract']`

Associated to the selectors `['get-method', 'method']`

From which we can generate generate:
`['private-abstract-get-method', 'private-abstract-method', 'abstract-get-method', 'abstract-method', 'private-get-method', 'private-method', 'get-method', 'method']

azat-io:
Yes, the growing number of groups looks like a problem.

What should the API for modifiers look like? I don't really understand the idea yet.

hugop95:
I don't think we have to change the API yet. However, in the code, instead of doing 

```
            let memberGroups = []

            ...

            if (isPrivateProperty) {
                    if (isReadonly) {
                      if (decorated) {
                        memberGroups.push('private-readonly-decorated-property')
                      }
                      memberGroups.push('private-readonly-property')
                    }
                }
            }
           
            // Other complex branches

            for (let memberGroup of memberGroups) {
              defineGroup(memberGroup)
            }
```
We could do:
```
    let modifiers = [];
    let selectors = [];
    if (isStaticMethod) {
      improvedMemberGroups.push('static');
    }
    if (isPrivateMethod) {
      modifiers.push('private');
    } else if (isProtectedMethod) {
      modifiers.push('protected');
    } else {
      modifiers.push('public');
    }
    if (isAbstractMethod) {
      modifiers.push('abstract');
    }
    if (decorated) {
      modifiers.push('decorated');
    }
    if (isGetMethod) {
      selectors.push('get-method');
    } else if (isSetMethod) {
      selectors.push('set-method');
    } 
    selectors.push('method');
    }

   // Generate memberGroups from modifiers + selectors
   let memberGroups = generateMemberGroups(modifiers, selectors)
    for (let memberGroup of memberGroups) {
          defineGroup(memberGroup)
     }
```

hugop95:
My idea of the `generateMemberGroups` function is to iterate over all modifiers union and add selectors at the end, like the following example:

element: `private abstract get a();`
modifiers: `['private', 'abstract']`
selectors: ['get-method', 'method']`

=> `['private-abstract-get-method', 'private-abstract-method', 'abstract-get-method', 'abstract-method', 'private-get-method', 'private-method', 'get-method', 'method']`

If we want to be lenient and allow users to enter `abstract-private-get-method` instead of `private-abstract-get-method`, we can make it so the `generateMemberGroups(modifiers, selectors)` generate all the possible combinations possible regardless of order.

hugop95:
I consider https://github.com/azat-io/eslint-plugin-perfectionist/pull/196 to be a superior implementation compared to this.

azat-io:
Closed in favor of #196.

PR #196, opened 2024-08-05
feat: `sort-classes`: Using dynamic-group-generation to add `readonly`, `override` and `abstract` modifiers, on top of other modifier combinations.

hugop95:
@azat-io 
In order to reduce the number of conditional branches when adding new modifiers like `abstract` or `readonly`, a system automatically generating group names can be used.

Complex cases like `protected abstract override readonly decorated properties` are automatically handled, and users can write
`protected, abstract, override, readonly, and decorated` in any order, followed by `property` to match this.

Each element being parsed is separated into one `selector` and potentially multiple `modifiers` (inspired by https://typescript-eslint.io/rules/naming-convention/#selector-options).

When an element gets parsed and matches multiple groups configured by the user, the following order of importance applies:

1. Selector priority: `constructor`, `get-method` and `set-method` groups will take precedence over `method` groups.
2. If the selector is the same: modifiers quantity priority: the group with the most modifiers matching will be selected.
3. If modifiers quantity is the same, order will be chosen based on modifier priority explained below.

_Example:_ 

```typescript
class Class {

    protected abstract a;

}
```
with the group configuration
````
['protected-property', 'abstract-property']
````

`abstract-property` will be matched because both `protected-property` and `abstract-property` have the same selector (`property`), the same number of modifiers (1), but `abstract` has a higher priority than `protected`.

Here are the following modifiers, associated with their selectors. Selectors and modifiers are sorted by importance, from most to least important.

The `static` modifier has been given the highest priority for all selectors handling it.

### Methods

- Selectors: `constructor`, `get-method`, `set-method`, `method`.
- Modifiers: `static`, `abstract`, `decorated`, `override`, `protected`, `private`, `public`.

### Accessors

- Selector: `accessor-property`.
- Modifiers: `static`, `abstract`, `decorated`, `override`, `protected`, `private`, `public`.

### Index-signature

- Selector: `index-signature`.
- Modifiers: `static`, `readonly`.

### Static-blocks

- Selector: `static-block`.
- Modifiers: No modifiers available.

### Properties

- Selector: `property`.
- Modifiers: `static`, `declare`, `abstract`, `decorated`, `override`, `readonly`, `protected`, `private`, `public`.

### CODE REVIEW

**Comment by hugop95** [comment]
Interesting point to note:
The config
```
 "groups": [
               "get-method",
              "public-method"
           ]
```
on code

```typescript
export class Test {

  get fields(): void {

  }

  public constructor() {
  }
}

```

will consider this code to be invalid: the `constructor` should go before `fields` because they are both considered `public-methods`.

This is because `get fields` generate the following ordered groups: 
```
[
'public-get-method', 
'public-method' (matches group n°2 in config), 
'get-method', 
'method'
]
```
while `constructor` generates: 
```
[
'public-constructor', 
'public-method' (matches group n°2 as well in config), 
'constructor', 
'method'
]
```
In order to fix it, the config must be more precise and replace `get-constructor` with `public-get-constructor`.

hugop95:
@azat-io I think that I am done with this PR. If everything is OK for you, do you want me to update the documentation?

.gitignore
```diff
@@ -8,6 +8,7 @@ dist/
 # Editor
 .vscode/
 .vim/
+.idea
```
OlivierZal, line 11: To be removed, I guess 😉

rules/sort-classes/sort-classes-utils.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+  let allModifiersCombinations: string[][] = []
+  for (let i = modifiers.length; i > 0; i--) {
+    allModifiersCombinations = [
+      ...allModifiersCombinations,
+      ...getCombinations(modifiers, i),
+    ]
+  }
+  let allModifiersCombinationPermutations = allModifiersCombinations
+    .map(result => getPermutations(result))
```
OlivierZal, line 33: Maybe use `flatMap()` for performances?

OlivierZal:
This work is awesome @hugop95!

Don’t forget to update the documentation.

rules/sort-classes/sort-classes-utils.ts
hugop95: Fixed

hugop95:
I have fixed the mentioned issues and updated the documentation as well. The Example n°2 was sorted based on the default sort.

docs/content/rules/sort-classes.mdx
```diff
@@ -215,146 +215,194 @@ Allows you to use comments to separate the class members into logical groups. Th
...
+  [key: string]: any;
 
-  // 'decorated-property'
+  // 'public-static-property'
+  static instance: Example;
+
+  // 'protected-abstract-override-readonly-decorated-property'
   @SomeDecorator
-  public value: number;
+  protected abstract override readonly _value: number;
```
OlivierZal, line 295: This illustrates the problem I have with `_` in JS: if I'm not mistaken, the rule will consider it as a private name, but because of the `protected` it will be flagged it as `protected`.

As a comparison, `protected #value` would have raised an error.

We shouldn't mix official JS syntax and developer conventions.

docs/content/rules/sort-classes.mdx
```diff
@@ -215,146 +215,194 @@ Allows you to use comments to separate the class members into logical groups. Th
...
+- Modifiers: `abstract`, `decorated`, `override`, `readonly`, `protected`, `private`, `public`, `static`.
+- Example: `readonly-decorated-property`.
+
+#### Important notes
+
+##### Scope of the `private` modifier
+The `private` modifier will currently match any of the following:
+- Elements with the `private` keyword.
+- Elements with their name starting with `#`.
+- Elements with their name starting with `_`.
```
OlivierZal, line 246: I would get rid...

docs/content/rules/sort-classes.mdx
```diff
@@ -215,146 +215,194 @@ Allows you to use comments to separate the class members into logical groups. Th
...
-- `'unknown'` — Members that don’t fit into any other group.
+Predefined groups are characterized by a single selector and potentially multiple modifiers. You may enter modifiers in any order, but the selector must always come at the end.
 
-Example:
+#### Methods
+- Selectors: `constructor`, `get-method`, `set-method`, `method`.
+- Modifiers: `abstract`, `decorated`, `override`, `protected`, `private`, `public`, `static`.
+- Example: `protected-abstract-override-method` or `static-get-method`.
+
+#### Accessors
```
OlivierZal, line 225: Maybe merge with the previous section since it's the same usage:
```md
#### Methods and accessors
- Selectors: `get-method`, `set-method`, `method`, `accessor-property`.
 - Modifiers: `abstract`, `decorated`, `override`, `protected`, `private`, `public`, `static`.
 - Example: `private-static-accessor-property`.
 ```

File: docs/content/rules/sort-classes.mdx [hunk 6]
```diff
@@ -215,146 +215,194 @@ Allows you to use comments to separate the class members into logical groups. Th
...
+- Selectors: `accessor-property`.
+- Modifiers: `abstract`, `decorated`, `override`, `protected`, `private`, `public`, `static`.
+- Example: `private-static-accessor-property`.
+
+#### Index-signature
+- Selectors: `index-signature`.
+- Modifiers: No modifier available.
```
OlivierZal, line 232: An further enhancement could add `readonly` here.

We also could add the `call` signatures:
```
(param: string): void {
  console.log(param);
}

// With array function
public callSignature: (param: string) => void = (param: string): void => {
  console.log(param);
};
```

docs/content/rules/sort-classes.mdx
OlivierZal, line 246: See https://github.com/azat-io/eslint-plugin-perfectionist/pull/196/files#r1706653975

docs/content/rules/sort-classes.mdx
```diff
@@ -215,146 +215,194 @@ Allows you to use comments to separate the class members into logical groups. Th
...
-- `'get-method'` — Getter methods.
-- `'set-method'` — Setter methods.
-- `'method'` — Regular methods.
-- `'unknown'` — Members that don’t fit into any other group.
+Predefined groups are characterized by a single selector and potentially multiple modifiers. You may enter modifiers in any order, but the selector must always come at the end.
 
-Example:
+#### Methods
+- Selectors: `constructor`, `get-method`, `set-method`, `method`.
+- Modifiers: `abstract`, `decorated`, `override`, `protected`, `private`, `public`, `static`.
```
OlivierZal, line 222: Maybe we could remind somewhere that for every selector, `abstract` is incompatible with:
- `static`
- `private`
- `decorated`

rules/sort-classes/sort-classes.ts
```diff
@@ -278,105 +288,127 @@ export default createEslintRule<Options, MESSAGE_ID>({
               }
             }
 
-            let isPrivate = name.startsWith('_') || name.startsWith('#')
+            let isPrivateName = name.startsWith('_') || name.startsWith('#')
```
OlivierZal, line 291: See https://github.com/azat-io/eslint-plugin-perfectionist/pull/196/files#r1706653975

OlivierZal (approved):

hugop95:
@OlivierZal What do you think about increasing the priority of the `static` modifier over accessibility modifiers? Currently,
`public-property` and `public-method` take precedence over `static-property` and `static-method`, which can be quite confusing:

```ts
class Test {
  public b;
  static a;
}
```

with the configuration 
```json
 "groups": [
              "public-property",
              "static-property"
            ]
```
will consider the code invalid because `a` is a `public-property` before a `static-property`.

OlivierZal:
I agree! Indeed `static` should be on top.

That being said, I'm not the project owner. I share my humble advice as a user because I believe @azat-io's project is awesome ❤️ and deserves to be enriched and shared with other developers.

Ultimately, @azat-io should have the final say on the features. 😊

hugop95:
Added the `declare` modifier associated to the `property` selector.

docs/content/rules/sort-classes.mdx
OlivierZal: `_` to be removed

hugop95:
Added `readonly` and `static` modifiers to `index-signature`.

docs/content/rules/sort-classes.mdx
```diff
@@ -241,8 +241,8 @@ The `declare` modifier is incompatible with the `override` and `decorated` modif
 
 #### Index-signature
```
OlivierZal, line 242: Plural if you want to be consistent with other section titles

azat-io:
Thanks for your work! LGTM.

Are you sure you should put the sort-classes rule code in a separate folder?

hugop95:
> Thanks for your work! LGTM.
> 
> Are you sure you should put the sort-classes rule code in a separate folder?

Because of the `sort-class-utils`, I thought it would be better to regroup the files together in the same folder, but I can move them out if needed.

azat-io:
Let's go back to the way it was before, shall we?

hugop95:
> What do you think about increasing the priority of the `static` modifier over accessibility modifiers? Currently,
> `public-property` and `public-method` take precedence over `static-property` and `static-method`, which can be quite confusing:
> 
> ```ts
> class Test {
>   public b;
>   static a;
> }
> ```
> 
> with the configuration 
> ```json
>  "groups": [
>               "public-property",
>               "static-property"
>             ]
> ```
> will consider the code invalid because `a` is a `public-property` before a `static-property`.
> 
> I would consider it to be the highest modifier priority.

@azat-io What about this?

hugop95:
> Let's go back to the way it was before, shall we?

Fixed.

christopher-buss:
> > What do you think about increasing the priority of the `static` modifier over accessibility modifiers? Currently,
> > `public-property` and `public-method` take precedence over `static-property` and `static-method`, which can be quite confusing:

It would make sense to me as an outsider, as per the eslint [member-ordering](https://typescript-eslint.io/rules/member-ordering/#member-types-granular-form) rule, static properties come first. I've posed this in another issue, but the order I'm personally trying to mimic is [here](https://github.com/aljazsim/vs-code-typescript-class-organizer/blob/master/src/member-type.ts), which also has all the static properties hoisted to the top.

hugop95:
- Added `static-block`
- Increased priority of the `static` modifier to be above everything else. 3 already existing unit tests needed to be adjusted as `private` had a greater priority over `static` in those.

OlivierZal:
Great @hugop95

azat-io:
@hugop95 I looks good. Great work. I think it's done, don't you? It seems we have one small detail left to correct in ESLint errors.

@OlivierZal Thank you so much for the review! Do you have any more questions?

OlivierZal:
You're welcome @azat-io, I like so much your project: it really deserves to be promoted among developers.

No more question for me.

hugop95:
> @hugop95 I looks good. Great work. I think it's done, don't you? It seems we have one small detail left to correct in ESLint errors.
> 
> @OlivierZal Thank you so much for the review! Do you have any more questions?

Error fixed, with the addition of the `else if (member.type === 'StaticBlock')`, it seems that the
```
else if (
              member.type === 'PropertyDefinition' ||
              member.type === 'TSAbstractPropertyDefinition'
            )
```
was not deemed necessary anymore for ESLint because `member` is necessarily a property at this stage. I have removed the condition, but if you prefer to keep it and `eslint-ignore` it, I can revert that.

PR #199, opened 2024-08-07
fix: switch `export default` to `export =` in `dist/index.d.ts`

### LINKED ISSUE #177: Typescript error 'Property 'configs' does not exist on type....
### Describe the bug

```
Property 'configs' does not exist on type 'typeof import("/home/kyle/dev/eslint-config/node_modules/eslint-plugin-perfectionist/dist/index")'.ts(2339)
```

### Code example

```ts
import perfectionist from "eslint-plugin-perfectionist";
import tseslint from "typescript-eslint";

export default tseslint.config(
  perfectionist.configs["recommended-natural"],
);
```

### ESLint version

8.57

3

### Additional comments

I'm wondering if the error is because of `tseslint.config()` helper function. `// @ts-expect-error` ignores it, and it does seem to work.

image

### Validations

JoshuaKGoldberg:
### Description

Manually changes the export from an ESM-style `export default` to a CJS-style `export =`. This better describes `dist/index.js`'s `module.exports = index`.

Fixes #177.

### Additional context

You can test this out locally by either:

* Replacing `node_modules/eslint-plugin-perfectionist/dist/index.d.ts` in https://github.com/JoshuaKGoldberg/repros/tree/repro-eslint-plugin-perfectionist-default-export-types
* Running `npx @arethetypeswrong/cli --pack` on the repo

Before this change, attw's `"eslint-plugin-perfectionist"` report was:

```plaintext
node10: ❗️ Incorrect default export
node16 (from CJS): ❗️ Incorrect default export
node16 (from ESM): 🐛 Used fallback condition
🎭 Masquerading as CJS
bundler: 🐛 Used fallback condition
```

Now:

```plaintext
node10: 🟢 
node16 (from CJS): 🟢 (CJS)
node16 (from ESM): 🐛 Used fallback condition
🎭 Masquerading as CJS
bundler: 🐛 Used fallback condition
```

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

- ~[ ] Ideally, include relevant tests that fail without this PR but pass with it.~ _(not sure how to here?)_

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -96,8 +96,7 @@ export default defineConfig({
           'dist/index.d.ts',
           (await fs.readFile('dist/index.d.ts'))
             .toString()
-            .replace(/\nexport .+/, '')
-             + "export = _default"
+            .replace(/\nexport .+/, '') + 'export = _default',
         )
       },
       include: [
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -91,6 +91,15 @@ export default defineConfig({
   },
   plugins: [
     dts({
+      afterBuild: async () => {
+        await fs.writeFile(
+          'dist/index.d.ts',
+          (await fs.readFile('dist/index.d.ts'))
+            .toString()
+            .replace(/\nexport .+/, '')
+             + "export = _default"
+        )
+      },
       include: [
         path.join(__dirname, 'index.ts'),
         path.join(__dirname, 'typings'),


vite.config.ts
```diff
@@ -91,6 +91,15 @@ export default defineConfig({
...
     dts({
+      afterBuild: async () => {
+        await fs.writeFile(
+          'dist/index.d.ts',
+          (await fs.readFile('dist/index.d.ts'))
+            .toString()
+            .replace(/\nexport .+/, '')
+             + "export = _default"
+        )
+      },
```
JoshuaKGoldberg, line 102: This is admittedly kind of hacky 😞 but I couldn't figure out a way to get this through the existing tooling.

* Switching the `tsconfig.json` `module` to `"nodenext"` doesn't change anything
* Neither does `"commonjs"` with a removal of `verbatimModuleSyntax`
* `vite-plugin-dts` doesn't have options for dual-format-emit that I could find

azat-io:
Thank you for your work!
Released in [v3.1.3](https://github.com/azat-io/eslint-plugin-perfectionist/releases/tag/v3.1.3)

JoshuaKGoldberg:
Thanks for the quick response + review! ⚡

Issue #200, opened 2024-08-08
Feature: `sort-enums` support sort by enum value instead of enum name


fkworld:
### What rule do you want to change?

sort-enums

### Describe the problem

For example, a numeric enums, I wish to sort by enum value, instead of enum name.

### Sort by enum value:
```ts
enum Direction {
  Up = 1,
  Down = 2,
  Left = 3,
  Right = 4,
}
```

### Sort by enum name:
```ts
enum Direction {
  Down = 2,
  Left = 3,
  Right = 4,
  Up = 1,
}
```

### Validations

- [X] Read the [docs](https://perfectionist.dev).
- [X] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

azat-io:
Added in 285a4516e6ee3893d888e1f26fb75424af003efc.
Released in [v3.2.0](https://github.com/azat-io/eslint-plugin-perfectionist/releases/tag/v3.2.0).

PR #203, opened 2024-08-13
feat: - Implements #200 (`sort-enums` support sort by enum value instead of enum name)

hugop95:
### Description

Implementation of https://github.com/azat-io/eslint-plugin-perfectionist/issues/200: `sort-enums` support sort by enum value instead of enum name 

- Adds a boolean `sortByValue` option (by default `false`) to `sort-enums`, comparing enum values rather than names. Numeric enums will always be sorted by their value numerically when this flag is true. The sort `order` is maintained.
- Adds a boolean `forceNumericSort` option (by default `false`). If `true`, numeric enums will always be sorted by their value numerically, regardless of the sort `type` entered and even if `sortByValue` is `false`. The sort `order` is maintained.
- Improved typings of compare and sort-nodes function.

This allows users to keep sorting their string enums alphabetically while sorting numeric enums by their values.

- [x] New Feature
- [x] Documentation update

### DIFF
--- a/docs/content/rules/sort-enums.mdx
+++ b/docs/content/rules/sort-enums.mdx
@@ -121,11 +121,16 @@ Controls whether sorting should be done using the enum's values or names.
 - `true` — Use enum values.
 - `false` — Use enum names.
 
+When this setting is `true`, numeric enums will have their values sorted numerically regardless of the `type` setting.
+
 ### forceNumericSort
 
 <sub>default: `false`</sub>
 
-TODO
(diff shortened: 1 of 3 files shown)


hugop95:
Possible improvement: Automatically sorting numeric enums by natural order if `compareValues` is `true`.

hugop95:
Tests and documentations to be added.

azat-io:
It looks good. Thank you for this PR!

Can I merge it?

hugop95:
> It looks good. Thank you for this PR!
> 
> Can I merge it?

Yes!

azat-io:
Thank you!

PR #204, opened 2024-08-14
`sort-classes` Add the element's matched group in ESlint error messages

hugop95:
### Description

I believe that the `modifiers` and `selectors` priority order for `sort-classes` is intuitive enough to do what users expect the rule to do in cases where an element matches multiple user-entered groups. However, giving the matched group's name in the ESLint error message would still be a good idea. This can help users in debugging their configuration, especially complex ones involving custom groups.

New error added: `unexpectedClassesGroupOrder`. This error will show the element's respective group they matched.

_Note_:  Elements that do not belong in the same group but end up being regrouped together by the user's configuration will not throw an `unexpectedClassesGroupOrder` error, but a usual `unexpectedClassOrder` instead.

### Code example 1

Configuration: 

```json
   "perfectionist/sort-classes": ["error", {
          "groups": [
            "get-method",
            "constructor",
            "protected-abstract-override-method"
          ]
        }]
```

Code: 
```ts
abstract class Class extends AbstractClass {
  
  constructor() {

  }

  protected abstract override get getMethod() {
    // This is a get-method before being a method, so it should go above the constructor
  }

}
```

Message:
image

### Code example 2

Configuration:
```
 "groups": [
              ["static-property", "private-property", "property"],
              "constructor",
              ["static-method", "private-method", "method"],
              "unknown"
            ]
```

Code:
```ts
class Class {

  method() {}

  static b = 'b'

  private a
}
```

Message:
image
 
- [x] New Feature

### DIFF
--- a/rules/sort-classes.ts
+++ b/rules/sort-classes.ts
@@ -514,9 +514,10 @@ export default createEslintRule<Options, MESSAGE_ID>({
                   isPositive(compare(left, right, options))))
             ) {
               context.report({
-                messageId: leftNum !== rightNum
-                  ? 'unexpectedClassesGroupOrder'
-                  : 'unexpectedClassesOrder',
+                messageId:
+                  leftNum !== rightNum
+                    ? 'unexpectedClassesGroupOrder'
+                    : 'unexpectedClassesOrder',
                 data: {
                   left: toSingleLine(left.name),
                   leftGroup: left.group,
--- a/rules/sort-classes.ts
+++ b/rules/sort-classes.ts
@@ -207,7 +207,8 @@ export default createEslintRule<Options, MESSAGE_ID>({
       },
     ],
     messages: {
-      unexpectedClassesOrder: 'Expected "{{right}}" to come before "{{left}}".',
+      unexpectedClassesOrder:
+        'Expected "{{right}}" ({{rightGroup}}) to come before "{{left}}" ({{leftGroup}}).',
     },
   },
   defaultOptions: [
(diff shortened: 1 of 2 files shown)


OlivierZal:
Nice @hugop95

azat-io:
Hi, thanks for the pool request!

Can you tell me what happens if a user doesn't have groups set?
Will there be an error like “Expected ‘y’ unknown to come before ‘x’ unknown”?

That seems like it would be redundant information?

azat-io:
We currently have an `unexpectedClassesOrder` error in use. Perhaps it would be better to create a new error `unexpectedClassesGroupOrder`. And show it only when class elements have different groups?

hugop95:
> Hi, thanks for the pool request!
> 
> Can you tell me what happens if a user doesn't have groups set? Will there be an error like “Expected ‘y’ unknown to come before ‘x’ unknown”?
> 
> That seems like it would be redundant information?

If an element matches no group set by the user, it will indeed display `unknown` 
Example:
image

I think this is interesting information to have: this means that the user's configuration (or default configuration if they don't use any) does not handle that element.

hugop95:
> We currently have an `unexpectedClassesOrder` error in use. Perhaps it would be better to create a new error `unexpectedClassesGroupOrder`. And show it only when class elements have different groups?

That sounds like a good idea, I'll add it.

hugop95:
Added.

azat-io:
Thank you!

Issue #189, opened 2024-08-03
Feature: Allow `perfectionist/sort-classes` to match `property-methods`: properties whose values are methods.


hugoAlt:
### What rule do you want to change?

`sort-classes`

### Describe the problem

`eslint/naming-convention` defines `classMethod` as "matches any class method. Also matches properties that have direct function expression or arrow function expression values" (https://typescript-eslint.io/rules/naming-convention/#selector-options).

`perfectionnists/sort-classes` does not recognize arrow methods as class methods.

### Code example

Configuration:

```JSON
"perfectionist/sort-classes": ["error", {
          "groups": [
            "index-signature",
            "static-property",
            "property",
            "private-property",
            [
              "get-method",
              "set-method"
            ],
            "constructor",
            "static-method",
            "private-method",
            "static-private-method",
            "method"
          ]
        }]
```

```typescript
class TestClass {

  attribute: string;

  constructor() {
    // Constructor stuff
  }

  arrowMethod = () => {
    // Arrow method stuff
  }

  method() {
    // Method stuff
  }
}

```

gets reformated into

```typescript
class TestClass {

  arrowMethod = () => {
    // Arrow method stuff
  }

  attribute: string;

  constructor() {
    // Constructor stuff
  }

  method() {
    // Method stuff
  }
}
```

Is there a way to make sure that `arrowMethod` gets grouped with `method`?

### Validations

- [X] Read the [docs](https://perfectionist.dev).
- [X] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

hugop95:
Feature ready in forked repo (https://github.com/hugop95/eslint-plugin-perfectionist/tree/feat/arrow-method), waiting for merge of https://github.com/azat-io/eslint-plugin-perfectionist/pull/196 before creating PR.

OlivierZal:
Hi @hugop95, I'm not 100% sure "method properties could be recognized as methods and not properties" straight forward.

Indeed arrow-functions misses some properties that class methods have, e.g. they cannot be binded.

Also some IDEs like VSCode don't recognize them as methods:

- either when declared on top of a class (case 1):

<img width="707" alt="Capture d’écran 2024-08-08 à 23 35 57" src="https://github.com/user-attachments/assets/4cc17396-f906-4a57-a446-3f7b992b28b1">

including when this declared property / arrow-function is assigned / used in the class

[...]

PR #205, opened 2024-08-14
Allow `sort-classes` to match `property-function`: properties whose values are directly functions or arrow-functions.

hugop95:
### Description

Solves https://github.com/azat-io/eslint-plugin-perfectionist/issues/189.

- Adds a new `selector`: `property-function`, with a higher priority than `property`. This selector will only match properties whose **values** are **directly** functions or arrow-functions (directly meaning that assigning a property a literal referring to a function will not detect it as a `property-function`). Properties **typed** as a function but without values will not be recognized as `property-function`.

### Additional context

I have been wondering if this feature should be implemented as a custom-group feature instead. (c.f: https://github.com/azat-io/eslint-plugin-perfectionist/issues/189#issuecomment-2276853937)

I agree with @OlivierZal that this is probably worth adding as an official group.

- [x] New Feature

### DIFF
--- a/docs/content/rules/sort-classes.mdx
+++ b/docs/content/rules/sort-classes.mdx
@@ -232,12 +232,16 @@ The `abstract` modifier is incompatible with the `static`, `private` and `decora
 `constructor`, `get-method` and `set-method` elements will also be matched as `method`.
 
 #### Properties
-- Selector: `property`.
+- Selectors: `function-property`, `property`.
 - Modifiers: `static`, `declare`, `abstract`, `decorated`, `override`, `readonly`, `protected`, `private`, `public`.
 - Example: `readonly-decorated-property`.
 
 The `abstract` modifier is incompatible with the `static`, `private` and `decorated` modifiers.
+
 The `declare` modifier is incompatible with the `override` and `decorated` modifiers.
+
+The `function-property` selector will match properties whose values are defined functions or arrow-functions.
+As such, the `declare` and `abstract` modifiers are incompatible with this selector.
 
 #### Index-signatures
 - Selector: `index-signature`.
@@ -316,6 +320,12 @@ abstract class Example extends BaseExample {
   // 'private-decorated-property'
   @SomeDecorator
   private _value: number;
+
+  // private-function-property
+  private arrowProperty = () => {};
+
+  // private-function-property
+  private functionProperty = function() {};
 
   // 'private-property'
   private name: string;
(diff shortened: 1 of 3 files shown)


OlivierZal:
@hugop95, I would just add the `call-signature` (`TSCallSignatureDeclaration`) use case in the scope of this PR, since this type is kind of an unassigned arrow-function signature.

hugop95:
> @hugop95, I would just add the `call-signature` (`TSCallSignatureDeclaration`) use case in the scope of this PR, since this type is kind of an unassigned arrow-function (signature).

Ok, I'll add it!

hugop95:
@OlivierZal Some notes and questions regarding `call-signature`:

- `a: () => {}` can be detected with `TSFunctionType`. Classes can't have call signatures if I recall, so I don't think `TSCallSignatureDeclaration` will be used. We could rename `call-signature` to `function-type-property` ?
- Take the following case:

```ts
type MyFunctionType = () => void; // A very complex function type

class Class {
   a: FunctionType
}
```

Should `a` be marked as `function-type-property`?
For information,
```ts
class Class {
    a = () => {}
    b = a
}
```
will **not** mark `b` as `function-property` currently.

- I think that all elements marked as `function-property` should also be marked as `function-type-property`. What do you think? In that case, the `selector` priority order related to properties would be: `function-property` > `function-type-property` > `property`.

OlivierZal:
@hugop95, all these are good questions and need some research from my end, because I think that call signatures do belong to classes.

There's also another case to handle I've just thought about: `TSConstructSignatureDeclaration`, e.g.
```
const a = new () => any
```

OlivierZal:
@hugop95, maybe this could help: https://babeljs.io/docs/v8-migration-api

hugop95:
> There's also another case to handle I've just thought about: `TSConstructSignatureDeclaration`, e.g.
> 
> ```
> const a = new () => any
> ```

These types of cases are why I was thinking of adding this feature in custom groups. In the proof of concept branch that I made:
https://github.com/hugop95/eslint-plugin-perfectionist/blob/c67783c9222425dc7e5e7de7e0d5c9cbb709a159/rules/sort-classes-utils.ts#L177

There is the possibility to explicitly enter patterns to match things types like the example you showed.

OlivierZal (approved):

hugop95:
Alright then!  

@azat-io PR is ready.

azat-io:
Thank you!

Issue #201, opened 2024-08-09
Feature: `sort-classes` Improved custom group system


hugop95:
### What rule do you want to change?

`sort-classes`

### Describe the problem

Improve the customGroups API by allowing an array of `CustomGroup` elements to be entered rather than an object. More information in PR: https://github.com/azat-io/eslint-plugin-perfectionist/pull/208

### Validations

- [X] Read the [docs](https://perfectionist.dev).
- [X] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

azat-io:
Added in baa701d494baa8d01f977a650080fb52c61dbc46.
Released in [v3.2.0](https://github.com/azat-io/eslint-plugin-perfectionist/releases/tag/v3.2.0).

hugop95:
@azat-io Not yet implemented! 😄 Could you reopen please?

azat-io:
Reopened.

PR #208, opened 2024-08-14
feat: `sort-classes`: #201: Improved and extensible custom groups

### LINKED ISSUE #165: Feature: Support grouping decorated entities by decorator
### Describe the rule

We have the option of grouping decorated entities together with `decorated-property`, `decorated-method`, etc. I would like the option of grouping the decorated entities by the decorator used. I don't like having to use a naming convention for every decorator.
Maybe supporting an extra configuration on the custom groups would be an option:
```json
"custom-groups": {
  "inputs": "{ 'groupByDecorator': 'Input' }",
  "outputs": "{ 'groupByDecorator': 'Output' }"
  ...
}
```

### Code example

This is the type of sorting I would like to end up with:

```ts
export class ExampleComponent {
  @Output()
  public changed = new EventEmitter();
  @Output()
  public clicked = new EventEmitter();
  @Output()
  public loaded = new EventEmitter();

  @Input()
  public baseColor: string;
  @Input()
  public item: string;
  @Input()
  public secondaryColor: string;
  
  @ViewChild('form')
  public form: TemplateRef<unknown>;
  @ViewChild('table')
  public table: TemplateRef<unknown>;
  
  @ViewChildren('elements')
  public elements: QueryList<ElementRef<any>>;
}
```

### Additional comments

_No response_

### Validations

hugop95:
### Description

Fixes https://github.com/azat-io/eslint-plugin-perfectionist/issues/201

Typings for this rule have been moved in their own file `sort-classes.types.ts`.

Improves the `customGroups` API by allowing an array of `CustomGroup` to be entered rather than an object. Entering an object will still use the currently existing `customGroups` API.

A `CustomGroup` may be one of the following:

```ts
interface SingleCustomGroup {
  groupName: string
  type?: 'alphabetical' | 'natural' | 'line-length' | 'unsorted'
  order?: 'asc' | 'desc'
  selector?: string
  modifiers?: string[]
  elementNamePattern?: string
  decoratorNamePattern?: string
}
```

or 

```ts
interface CustomGroupBlock {
  groupName: string
  type?: 'alphabetical' | 'natural' | 'line-length' | 'unsorted'
  order?: 'asc' | 'desc'
  anyOf: Array<{
      selector?: string
      modifiers?: string[]
      elementNamePattern?: string
      decoratorNamePattern?: string
  }>
}
```
The objectives are:
- To give the possibility to group elements in a precise way by offering various filters.
- to give the possibility to users to choose the sorting type of each group should they want, in particular: `unsorted`. This allows users to have complex configurations where they would want `methods` to be grouped together below the `constructor`, but remain unsorted. If the custom group's `type` or `order` are not given, the top-level `type` and `order` options are taken.

The `customGroups` list is ordered: the first `CustomGroup` definition that matches an element will be used.
Custom groups have a higher priority than any official group. This is important:

The custom group
```
{
    groupName: 'myCustomGroup',
    selector: 'method',
   ....
}
```

will match `constructor() {}` before the official group `constructor`, because constructors match the `method` selector (this is possibly something that should to be addressed in the future).

As such, users can do
```
customGroups: [
    {
        groupName: 'constructor',
        selector: 'constructor',
    },
    {
        groupName: 'myCustomGroup',
        selector: 'method',
       ....
    }
]
```

to ensure that `constructor() {}` gets matched with the custom group `constructor` (which replicates what the official `constructor` group does) before `myCustomGroup`.

This interface is easily extensible if additional features need to be added later on.

- `groupName`: The group's name. Today in `customGroups`, this is one key of the object.
- `selector`: Filter on the `selector` of the element.
- `modifiers`: Filter on the `modifiers` of the element. (All the modifiers of the element must be present in that list)
- `elementNamePattern`: If entered, will check that the name of the element matches the pattern entered. Today in `customGroups`, this is the value associated to a key of the object.
- `decoratorNamePattern`: If entered, will check that at least one `decorator` matches the pattern entered. (fixes #165)
- `type`: Overrides the sort type for that custom group. `unsorted` will not sort the group.
- `order`: Overrides the sort order for that custom group

### Code example

Here is the simplified configuration that fixes #165:
```json
  {
            "groups": [
              "input",
              "output"
            ],
            "customGroups": [
              {
                "groupName": "input",
                "decoratorNamePattern": "Input"
              },
              {
                "groupName": "output",
                "decoratorNamePattern": "Output"
              }
            ]
          }
```

Here is an example of configuration that groups methods and properties together without sorting them

```json
{
            "groups": [
              "constructor",
              "unsortedMethodsAndProperties"
            ],
            "customGroups": [
              {
                // Prioritize matching constructor before anything else:
                // `constructor` are methods but should not be treated as such and grouped with properties here.
                "groupName": "constructor",
                "selector": "constructor"
              },
              {
                "groupName": "unsortedMethodsAndProperties",
                "type": "unsorted",
                "anyOf": [
                  {
                    "selector": "property"
                  },
                  {
                    "selector": "method"
                  }
                ]
              }
            ]
          }
```

- [x] Feature implementation
- [x] Tests
- [x] Documentation 

- [x] New Feature
- [x] Documentation update

### DIFF
--- a/docs/content/rules/sort-classes.mdx
+++ b/docs/content/rules/sort-classes.mdx
@@ -450,9 +450,60 @@ abstract class Example extends BaseExample {
 
 ### customGroups
 
-<sub>default: `{}`</sub>
+<sub>default: `[]`</sub>
 
-You can define your own groups for class members using custom glob patterns for matching.
+You can define your own groups to match very specific members.
+
+A custom group definition may follow one of the two following interfaces:
+
+```ts
+interface CustomGroupDefinition {
+  groupName: string
+  type?: 'alphabetical' | 'natural' | 'line-length' | 'unsorted'
+  order?: 'asc' | 'desc'
+  selector?: string
+  modifiers?: string[]
+  elementNamePattern?: string
+  decoratorNamePattern?: string
+}
+```
+A class member will match a `CustomGroupDefinition` group if it matches all the filters of the custom group's definition.
+
+or:
+
+```ts
+interface CustomGroupBlockDefinition {
+  groupName: string
+  type?: 'alphabetical' | 'natural' | 'line-length' | 'unsorted'
+  order?: 'asc' | 'desc'
+  anyOf: Array<{
+      selector?: string
+      modifiers?: string[]
+      elementNamePattern?: string
+      decoratorNamePattern?: string
+  }>
+}
+```
+
+A class member will match a `CustomGroupBlockDefinition` group if it matches all the filters of at least one of the `anyOf` items.
+
+#### Attributes
+
+- `groupName`: The group's name, which needs to be put in the `groups` option.
+- `selector`: Filter on the `selector` of the element.
+- `modifiers`: Filter on the `modifiers` of the element. (All the modifiers of the element must be present in that list)
+- `elementNamePattern`: If entered, will check that the name of the element matches the pattern entered.
+- `decoratorNamePattern`: If entered, will check that at least one `decorator` matches the pattern entered.
+- `type`: Overrides the sort type for that custom group. `unsorted` will not sort the group.
+- `order`: Overrides the sort order for that custom group
+
+#### Match importance
+
+The `customGroups` list is ordered:
+The first custom group definition that matches an element will be used.
+
+Custom groups have a higher priority than any predefined group. If you want a predefined group to take precedence over a custom group,
+you must write a custom group definition that does the same as what the predefined group does (using `selector` and `modifiers` filters), and put it first in the list.
 
 Example:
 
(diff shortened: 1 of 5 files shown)


hugop95:
@OlivierZal Do you have some use cases related to sorting properties with specific values (other than function properties) or types? I don't know yet how `valueTypePattern` or `typeDeclarationPattern` could be used.

OlivierZal:
@hugop95, not really. My only real-life use case with custom groups is "eventHandlers "(`onInit()`, `onUninit()`, `on('event', callback)` etc.).

OlivierZal:
@hugop95,

Indeed I don't see concrete use for type filters within options. Maybe could we be agile and implement when it is requested by users? (so we don't unnecessarily complexify the code). 

On the other hand `decoratorElementPattern` is an excellent idea, e.g. to gather all methods decorated with `@Infer`.

OlivierZal:
Another idea (but not in the scope of this PR) could be adding an option to `sortDecoratedElementsByDecoratorName` (I don't know how/where yet).

hugop95:
> @hugop95,
> 
> Indeed I don't see concrete use case for type filters within options. Maybe could we be agile and not implement until it is requested by users? (so we don't unnecessarily complexify the code).

It's probably what I will do, the interface is extensible so it can be added later without much work.

hugop95:
@azat-io If this PR is fine for you, I can start updating the documentation.

hugop95:
Added the following features:
A custom group can be one of the following:

```ts
interface SingleCustomGroup {
  groupName: string
  type: 'alphabetical' | 'natural' | 'line-length' | 'unsorted'
  order: 'asc' | 'desc
  selector?: string
  modifiers?: string[]
  elementNamePattern?: string
  decoratorNamePattern?: string'
}
```

or 

```ts
interface CustomGroupBlock {
  groupName: string
  type: 'alphabetical' | 'natural' | 'line-length' | 'unsorted'
  order: 'asc' | 'desc'
  subgroups: Array<{
      selector?: string
      modifiers?: string[]
      elementNamePattern?: string
      decoratorNamePattern?: string
  }>
}
```

The objective is to give the possibility to users to choose the sorting type of each group should they want, in particular: `unsorted`. This allows users to have complex configurations where they would want `methods` to be grouped together below the `constructor`, but remain unsorted.

`CustomGroupBlock` allows users to regroup multiple subgroups together.

Example

```
{
            "groups": [
              "index-signature",
              "unsortedMethodsAndProperties"
            ],
            "customGroups": [
              {
                "groupName": "unsortedMethodsAndProperties",
                "type": "unsorted",
                "subgroups": [
                  {
                    "selector": "property"
                  },
                  {
                    "selector": "method"
                  }
                ]
              }
            ]
          }
``` 

This would be somehow equivalent to
```
{
            "groups": [
              "index-signature",
              ["method", "property"]
            ],
}
```
with the added benefit of customizing the sorting type of the `["method", "property"]` group among other things.

hugop95:
PR ready for review.

azat-io:
This looks like a breaking change that would require a major version release. Would it be possible to use the old configurations?

Perhaps we can support both objects and arrays in groups?

hugop95:
> This looks like a breaking change that would require a major version release. Would it be possible to use the old configurations?
> 
> 
> 
> Perhaps we can support both objects and arrays in groups?

Both are supported! If an object is entered instead of an array, the old API will be used.

azat-io:
Nice! Thank you!

Issue #209, opened 2024-08-14
Bug: `sort-classes` Attributes with the same name but within different groups may not trigger linting


hugop95:
### Describe the bug

Comparisons between attributes with the same name currently do not trigger linting. 
This means that a class that only has sorting errors between elements with the same name will not trigger linting and will remain unsorted.

### Code example

Configuration

```
 "groups": [
            "static-property",
            "property"
          ]
```

Valid code:
```ts
class Class {
  a
  static a;
}
```

Invalid code:
```ts
class Class {
  a
  static b;
}
```

However, the code below will be sorted correctly because `b` and `a` will trigger linting.
```ts
class Class {
  b
  a
  static a;
}
```

### ESLint version

8.57.0

### ESLint Plugin Perfectionist version

3.2.0

### Validations

- [X] Read the [docs](https://perfectionist.dev).
- [X] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

PR #210, opened 2024-08-14
fix: #209 and #212 - `sort-classes` Attributes with the same name but within different groups may not trigger linting + line-length can break overload-signatures

hugop95:
Fixes https://github.com/azat-io/eslint-plugin-perfectionist/issues/209 and fixes https://github.com/azat-io/eslint-plugin-perfectionist/issues/212

Does **not** correct https://github.com/azat-io/eslint-plugin-perfectionist/issues/211.

This comparison https://github.com/azat-io/eslint-plugin-perfectionist/blob/4f3da85bf2efd779232950ee621904a665d6df17/rules/sort-classes.ts#L526 

seems responsible for the two issues linked.

- #209: Explicit
- #212: I assume that this check was put to prevent overload-signatures from trying to sort themselves. 

This is not a problem with the `natural` and `alphabetical` sort type because overload signatures have the same name in this case and as such do not impact sorting. It is with `line-length` because overload signatures do not necessarily have the same length.

The `left.name !== right.name` check does prevent overload-signature comparisons from activating linting but does not impact the sorting result itself, meaning that if any other class member activates linting because it is unordered, overload-signatures order may still change, and as such break code.

This PR does two things:
- Remove the `left.name !== right.name` check
- Make sure that the line-length for a group of overload signatures is the same between all its members (the size of the method implementation has been chosen).

- [x] Bug fix

### DIFF
--- a/rules/sort-classes-utils.ts
+++ b/rules/sort-classes-utils.ts
@@ -1,3 +1,5 @@
+import type { TSESTree } from '@typescript-eslint/utils'
+
 import type { Modifier, Selector } from './sort-classes'
 
 /**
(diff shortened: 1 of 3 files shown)


azat-io:
Thank you! I'll merge it today.
Is this a critical bug? Or can I wait for #208 and create a release later?

hugop95:
Not critical, #209 is a edge case that has no reason to happen in real projects. #212 is more serious but I don't think `line-length` is a poplar sort choice for this rule.

OlivierZal:
@hugop95, I also think we could avoid theoretical errors by replacing `isPrivateName` logic by something more reliable like https://github.com/typescript-eslint/typescript-eslint/blob/a8f91e3b84e0afcb4c08b23fde2596517eb9810e/packages/eslint-plugin/src/rules/member-ordering.ts#L533

hugop95:
> @hugop95, I think we also could avoid theoretical errors by replacing the current `isPrivateName` logic (based on name start) by something more reliable like https://github.com/typescript-eslint/typescript-eslint/blob/a8f91e3b84e0afcb4c08b23fde2596517eb9810e/packages/eslint-plugin/src/rules/member-ordering.ts#L533

Indeed, the current check is a bit clunky. I think we can add that change to a new PR.

OlivierZal:
@hugop95, I’ve also an error with `decorated-private-accessor` which is not properly sorted. 

I’m not at home so I cannot provide details, will try tonight.

hugop95:
> @hugop95, I’ve also an error with `decorated-private-accessor` which is not properly sorted.
> 
> I’m not at home so I cannot provide details, will try tonight.

Do you mean `decorated-private-accessor-property`?

OlivierZal:
Yes, but I confess I did it quickly before to leave home this morning. Need to test and improve my config.

OlivierZal:
@hugop95, I confirm it's a false positive: sometimes (rarely) the rule fails to be applied when saving in VSCode because I have many rules which are evaluated at the same time.

azat-io:
LGTM.

@OlivierZal Do you have any issues about this PR?

OlivierZal:
@azat-io it's also good for me.

PR #215, opened 2024-08-15
feat: `sort-classes`: Add `static-block` and `accessor-property` to default `groups` options

hugop95:
Default `groups` options currently do not handle `static-block` and `accessor-property`. Those elements get marked as `unknown` and end up at the bottom of the class.

This PR changes the default `groups` options by adding `static-block` at the very top, and grouping properties and accessor-properties together.

- [x] New Feature


hugop95:
> Should we add tests for this?

I will add some

hugop95:
Test added and documentation updated.

PR #216, opened 2024-08-15
feat: `sort-classes`: #197: Adds `group-kind`  option

### LINKED ISSUE #197: Feature: Add `groupKind` option to `perfectionist/sort-classes`

`perfectionist/sort-classes`

### Describe the problem

Like for `perfectionist/sort-interfaces`, a `groupKind: 'optional-first' | 'required-first' | 'mixed' = 'mixed'` option would be useful and allow what the `optionalityOrder?: 'optional-first' | 'required-first'` option does for the frozen [member-ordering](https://typescript-eslint.io/rules/member-ordering/) rule.

### Code example

The proposed option exists [here](https://github.com/azat-io/eslint-plugin-perfectionist/blob/main/rules/sort-interfaces.ts#L26) for another rule (`perfectionist/sort-interfaces`).

### Additional comments

_No response_

### Validations

hugop95:
### Description

Alternative PR: https://github.com/azat-io/eslint-plugin-perfectionist/pull/219. Fixes #197.

This PR adds a `group-kind` option similar to what exists for the `sort-object-types` rule, with 3 possible values:
`mixed` (default), `optional-first` and `required-first`.

Optionality applies to each group, not as a whole: if `optional-first` is selected, each group will have its optional members at its top.

The algorithm used to detect this error is similar to the one used in `sort-object-types`, and is slightly more compact than what existed, so I also updated it in the `sort-object-types.ts` as well.

Also updated a missing documented attribute in `sort-enums`.

Only properties and accessor properties are affected by this flag, so it would be understandable to instead only add an `optional` modifier to properies and accessor properties rather than a top-level flag.

- [x] New Feature

### DIFF
--- a/docs/content/rules/sort-classes.mdx
+++ b/docs/content/rules/sort-classes.mdx
@@ -191,6 +191,16 @@ Allows you to use comments to separate the class members into logical groups. Th
 -	`false` — Comments will not be used as delimiters.
 - string — A glob pattern to specify which comments should act as delimiters.
 
+### groupKind
+
+<sub>default: `'mixed'`</sub>
+
+Allows you to group properties by their kind, determining whether required properties should come before or after optional properties.
+
+- `mixed` — Do not group properties by their kind; required properties are sorted together with optional properties.
+- `required-first` — Group all required properties before optional properties.
+- `optional-first` — Group all optional properties before required properties.
+
 ### groups
 
 <sub>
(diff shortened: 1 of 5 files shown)


hugop95:
Putting as draft because I just noticed an small issue algorithm mentioned above.

hugop95:
Closing due to alternative PR merged.

PR #219, opened 2024-08-16
feat: `sort-classes`: #197: Adds `optional` modifier to `property` and `method`

### LINKED ISSUE #197: Feature: Add `groupKind` option to `perfectionist/sort-classes`

hugop95:
### Description

Alternative PR to https://github.com/azat-io/eslint-plugin-perfectionist/pull/216, fixes #197 in a different way than the original idea.

The `optional` modifier has a lower priority than accessibility modifiers `public`, `protected` and `private`, meaning that

`a?` gets recognized as a `public-property` before `optional-property`.

Note that these PRs do not solve the exact same cases:

- This PR allows users to place optional members in a different group than the required ones while #216 only sorts optional and required members from the same group.
- This PR does not allow users to put `property` and `method` in the same group while putting optional members from both groups at the top (or at the bottom). #216 allows it.

- [x] New Feature

### DIFF
--- a/docs/content/rules/sort-classes.mdx
+++ b/docs/content/rules/sort-classes.mdx
@@ -223,15 +223,23 @@ Predefined groups are characterized by a single selector and potentially multipl
 - Modifiers: `protected`, `private`, `public`.
 - Example: `protected-constructor`, `private-constructor`, `public-constructor` or `constructor`.
 
-#### Methods and accessors
-- Methods selectors: `get-method`, `set-method`, `method`.
-- Accessors selector: `accessor-property`.
-- Modifiers: `static`, `abstract`, `decorated`, `override`, `protected`, `private`, `public`.
+#### Methods
+- Selectors: `get-method`, `set-method`, `method`.
+- Modifiers: `static`, `abstract`, `decorated`, `override`, `protected`, `private`, `public`, `optional`.
 - Example: `private-static-accessor-property`, `protected-abstract-override-method` or `static-get-method`.
+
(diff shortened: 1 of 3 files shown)


OlivierZal:
@hugop95 

An accessor cannot be optional:

Capture d’écran 2024-08-16 à 18 00 22

hugop95:
> @hugop95 
> 
> An accessor cannot be optional:
> 
> Capture d’écran 2024-08-16 à 18 00 22
> 

Thanks, I'll fix it

hugop95:
Added optional methods

OlivierZal:
@hugop95, I confess I didn’t know that class method could be optional 😅

However I think abstract methods can’t be.

OlivierZal:
I confess I didn’t know about optional class methods, and that even abstract method could be optional 😅

hugop95:
Not something that I use either!

PR #220, opened 2024-08-16
feat: `sort-classes`: #207: Do not sort `unknown` elements if `unknown` is not referenced in `groups`

### LINKED ISSUE #207: Feature: sort-classes ignore methods

sort-classes

### Describe the problem

One of my biggest restrictions is that I think it can be too restrictive to sort classes, as sometimes it makes sense to keep blocks of code that are meaningful to each other closeby. That being said, I do want the option to sort the variablesa at the top. I was hoping that just not adding method blocks would have been enough, but I realise it still sorts everything else in the class by alphabetical order.

### Code example

-

### Additional comments

_No response_

### Validations

hugop95:
Fixes #207.

This PR adds the following:
- A new behavior for the `unknown` group: If the `unknown` group is not referenced in the user's configuration, elements from the `unknown` group will be ignored and not sorted. This prevents `unknown` elements from being put and sorted at the bottom of the class.

I have chosen not to add the `ignoreGroups` option as I believe this can be added in https://github.com/azat-io/eslint-plugin-perfectionist/pull/208 instead. This feature can be added later if needed anyway.

Example configuration that prevents `method` from being sorted:

```
"perfectionist/sort-classes": ["error", {
          "groups": [
            // Groups that do not include `method` nor `unknown`
          ]
        }
]
```

- [x] Implement feature
- [x] Add tests
- [x] Update documentation 

- [x] New Feature

### DIFF
--- a/docs/content/rules/sort-classes.mdx
+++ b/docs/content/rules/sort-classes.mdx
@@ -266,7 +266,8 @@ Elements that are not `protected` nor `private` will be matched with the `public
 
 
 ##### The `unknown` group
-Members that don’t fit into any group entered by the user will be placed in the `unknown` group.
+Members that don’t fit into any group entered by the user will be placed in the `unknown` group. If the `unknown` group is not specified in `groups`,
+the members will remain in their original order.
 
 ##### Behavior when multiple groups match an element
 
--- a/rules/sort-classes.ts
+++ b/rules/sort-classes.ts
@@ -126,16 +126,10 @@ type Options = [
     type: 'alphabetical' | 'line-length' | 'natural'
     partitionByComment: string[] | boolean | string
     groups: (Group[] | Group)[]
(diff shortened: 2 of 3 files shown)


hugop95:
Elements that do not match with the user's configuration are put in the `unknown` group, and the `unknown` group will be placed and sorted at the very bottom of the class if it is not explicitly defined in the user's `group` configuration, which I think is not what the user expects to be happening.

Maybe we can improve that behavior by leaving `unknown` elements at their original position if the `unknown` group is not in the user's configuration.

hugop95:
@azat-io If this PR is fine for you, I can start updating the documentation.

azat-io:
Released in [v3.3.0](https://github.com/azat-io/eslint-plugin-perfectionist/releases/tag/v3.3.0).

PR #228, opened 2024-08-21
feat: `sort-union-types`: #223: Updated documentation

### LINKED ISSUE #223: Docs: Explain & show examples of 'groups'
### Describe the rule

https://perfectionist.dev/rules/sort-union-types#groups exists that mentions what `groups` are... but I can't find any examples of how to use it on the docs site. Would you be up for adding some please?

Specifically, I'm coming over from https://github.com/typescript-eslint/typescript-eslint/pull/9698: we'd like to configure `null` and `undefined` to come last in our union types. This was previously available as a `nullable-last` option (https://github.com/azat-io/eslint-plugin-perfectionist/issues/61) but was removed in 3.0.0.

### Code example

```js
'perfectionist/sort-union-types': [
  'error',
  {
    groups: [ /* ??? */ ]
  }
]
```

### Additional comments

Sorry for using the wrong template 😅 there wasn't one I could find for docs / website stuff.

### Validations

hugop95:
Fixes #223 .

### Description

Improved documentation by adding group examples in the `Usage` section, inspired by the example above.

- [x] Documentation update


docs/content/rules/sort-union-types.mdx
```diff
@@ -201,7 +201,20 @@ type Example =
                   type: 'alphabetical',
                   order: 'asc',
                   ignoreCase: true,
-                  groups: [],
+                  groups: [
+                    'conditional',
```
azat-io, line 205: We specify default values everywhere in the examples. Maybe it is unnecessary?

hugop95: I reverted the `Usage` section to what it was and updated the example section directly instead.

PR #230, opened 2024-08-21
fix: `sort-switch-case`: #229: case 'default': is confounded with default: clause

### LINKED ISSUE #229: Bug(sort-switch-case): `case 'default':` is confounded with `default:` clause
### Describe the bug

A switch ... case control flow with a `case 'default` is mishandled and report a false positive error. 

### Code example

```js
    switch (variable) {
      case 'add':
        break
      case 'remove':
        break
      case 'default':
        break
      default:
        break
```

### ESLint version

9.9.0

3.2.0

### Additional comments

Using `recommended-natural` configuration. 

### Validations

hugop95:
### Description

Fixes #229.

- [x] Bug fix

### DIFF
--- a/rules/sort-switch-case.ts
+++ b/rules/sort-switch-case.ts
@@ -24,6 +24,10 @@ type Options = [
   }>,
 ]
 
+interface SortSwitchCaseSortingNode extends SortingNode<TSESTree.SwitchCase> {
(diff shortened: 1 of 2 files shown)


azat-io:
Thanks!

Issue #214, opened 2024-08-15
Feature:  Warn the user/Throw an error when a group entered in config does not exist


hugop95:
### Describe the rule

Groups that are not part of predefined groups nor entered as custom groups have no reason to exist in the `groups` attribute.

### Code example

Configuration
```
        "perfectionist/sort-classes": ["error", {
          "groups": [
            "accessor",
            "property",
            "unknown"
          ]
        }
```

Code
```ts
class Class {
  a;
  accessor b;
}
```

This code is valid because `accessor b` is part of the `unknown` group. Here, the user likely meant to write the `accessor-property` predefined group, but has no way to see the typo. Any group entered in the configuration should either:
- Be a predefined group.
- Be a custom group. 

### Additional comments

Affected rules:

- [ ] `sort-classes`
- [x] `sort-astro-attributes`
- [x] `sort-imports`
- [x] `sort-interfaces`
- [x] `sort-intersection-types`
- [x] `sort-jsx-props`
- [x] `sort-object-types`
- [x] `sort-object`
- [x] `sort-svelte-attributes`
- [x] `sort-union-types`
- [x] `sort-vue-attributes`

### Validations

- [X] Read the [docs](https://perfectionist.dev).
- [X] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

PR #232, opened 2024-08-27
feat: all rules: #214: Warn the user/Throw an error when a group entered in config does not exist

hugop95:
### Description

Resolves (partially) #214. The `sort-classes` rule is not part of this PR as checking if a group entered by the user is part of predefined groups requires a more complex strategy which I believe deserves its own PR.

Checks and validates group configuration for rules using `groups`. Configurations referencing groups that are
- not part of predefined groups
- and not defined in `customGroups`

are very likely user mistakes or typos. These groups are considered invalid and will throw an error, stopping ESLint.

image

Affected rules:
- `sort-astro-attributes`
- `sort-imports`
- `sort-interfaces`
- `sort-intersection-types`
- `sort-jsx-props`
- `sort-object-types`
- `sort-object`
- `sort-svelte-attributes`
- `sort-union-types`
- `sort-vue-attributes`

Remaining to do:

- [x] Add tests for each rule affected.

### Additional context

- Some tests have been slightly changed as they were referencing invalid groups. In those cases, invalid groups were removed from the test configuration.

- Ideally, we should be testing that invalid group configurations throw an error for each rule. Unfortunately, it is currently not possible to test rules that throw errors (https://github.com/eslint/eslint/issues/13434 | https://github.com/eslint/eslint/pull/16823). Furthermore, getting 100% code coverage in `validateGroupsConfiguration.ts` is not possible only through ESLint's RuleTester as there is no way to catch
the error thrown. We get 100% coverage temporarily in that file with a simple unit test in `validateGroupsConfiguration.test.ts` until that feature is implemented in ESLint.

- [x] New Feature

### DIFF
--- a/test/sort-objects.test.ts
+++ b/test/sort-objects.test.ts
@@ -276,7 +276,7 @@ describe(ruleName, () => {
             {
               ...options,
               customGroups: { top: ['c', 'b'] },
-              groups: ['top', 'unknown'],
+              groups: ['top'],
             },
           ],
         },
@@ -303,7 +303,7 @@ describe(ruleName, () => {
             {
               ...options,
               customGroups: { top: ['c', 'b'] },
-              groups: ['top', 'unknown'],
+              groups: ['top'],
             },
           ],
           errors: [
@@ -1163,7 +1163,7 @@ describe(ruleName, () => {
             {
               ...options,
               customGroups: { top: ['c', 'b'] },
-              groups: ['top', 'unknown'],
+              groups: ['top'],
             },
           ],
         },
@@ -1190,7 +1190,7 @@ describe(ruleName, () => {
             {
               ...options,
               customGroups: { top: ['c', 'b'] },
-              groups: ['top', 'unknown'],
+              groups: ['top'],
             },
           ],
           errors: [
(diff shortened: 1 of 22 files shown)


test/sort-objects.test.ts
```diff
@@ -276,7 +276,7 @@ describe(ruleName, () => {
             {
               ...options,
               customGroups: { top: ['c', 'b'] },
-              groups: ['top', 'unknown'],
```
azat-io, line 279: Why remove `'unknown'`?

test/validate-groups-configuration.test.ts
```diff
@@ -0,0 +1,21 @@
+import { describe, expect, it } from 'vitest'
+
+import { validateGroupsConfiguration } from '../utils/validate-groups-configuration'
+
+/**
+ * It is currently not possible to test rules that throw errors (https://github.com/eslint/eslint/issues/13434),
```
azat-io, line 6: Let's try to stick to the 80 character line length limit rule as stated in the prettier config?
I suggest breaking the comment down into a few lines.

azat-io (commented):
Nice job! ❤️

test/validate-groups-configuration.test.ts
hugop95: Fixed

test/sort-objects.test.ts
hugop95: Fixed. Mistake on my part.

azat-io:
Great. Thank you!

PR #233, opened 2024-08-28
feat: Make `test:unit` check fail if code coverage is not 100%

hugop95:
### Description

This PR forces code coverage to be at 100% for the `test:unit` check to succeed.

image

### What is the purpose of this pull request

- [x] Other


azat-io:
Nice! Thanks! ❤️

PR #239, opened 2024-08-28
fix: `sort-classes` #234, #237, #238, #240, #241, #242, #243, #244 and #245: Dependency related issues

### LINKED ISSUE #234: Bug(sort-classes): static section incorrectly placed before static properties definition
### Describe the bug

static section incorrectly placed before static properties definition triggering JS runtime error. 

### Code example

Sorting done by `sort-classes` rule:

```ts
class A { 
  static {
    Configuration.cache = new Map<
      string,
      object
    >([
      ['foo', new Foo()],
      ['bar', new Bar()],
    ])
  }

  private static cache: Map<string, object>
}
```

### ESLint version

9.9.1

3.3.0

### Additional comments

_No response_

### Validations

### LINKED ISSUE #237: Bug: `sort-classes`: Static properties referenced using MyClass.x are not detected as dependencies
### Describe the bug

Properties dependent on other properties through the `this.x` keyword are properly detected and sorted. However, static properties dependent on other static properties through the `MyClass.x` keyword are not.

### Code example

The following code is invalid and will try to put `a` before `b` when it should not.

```typescript
class Class {
  static b = 1

  static a = Class.b
}
```

### ESLint version

8.57.0

v3.3.0

### Additional comments

_No response_

### Validations

### LINKED ISSUE #238: Bug: `sort-classes`: Function properties's body dependencies are not properly detected when tests say they should
### Describe the bug

Function properties's body dependencies are not properly detected.

### Code example

```ts
class Class {
  z = 10

  a = () => {
    return this.z;
  }
}
```

This code is considered invalid and requires property `a` to be before property `z`.

While having `a` before `z` will not cause issues at compile time nor at runtime, the current codebase and tests clearly indicate that the expected sorting should be `z` before `a` because `a`'s body depends on `z`.

This issue becomes more problematic with `static` blocks (see https://github.com/azat-io/eslint-plugin-perfectionist/issues/234).
In this case, body dependencies become relevant to ensure that code still compiles and runs correctly after sorting.

The tests associated to this issue do not work as expected. For example https://github.com/azat-io/eslint-plugin-perfectionist/blob/305d88c6a7a901321eaed5313edc13a9989d59f8/test/sort-classes.test.ts#L2671-L2681

This test does nothing, because the sort type is `alphabetical`, so `method` should come after `a` anyway.

Either:
- Function properties should ignore body dependencies because they do not actually require the properties they depend on to be before them.
- Or function properties should always be placed after their dependencies even if this does not impact compilation or runtime.

### ESLint version

8.57.0

v3.3.0

### Additional comments

_No response_

### Validations

### LINKED ISSUE #240: Bug: `sort-classes`: Dependencies ignore the notion of `static`
### Describe the bug

Properties and static properties can coexist in the same class with the same name.

Dependencies extraction ignores the notion of `static` and may consider that `a` depends on `b` even if it's not the case.

### Code example

Group configuration
```
"groups": [
  ["property", "static-property"]
]
```

```ts
export class Class{
  b = 1;
  a = this.b;

  static b = 1;
}
```

This code is considered invalid (it is valid) and requires `static b` to be before `a` because it is considered a dependency of `a`. 

### ESLint version

8.57.0

v3.3.0

### Additional comments

_No response_

### Validations

### LINKED ISSUE #241: Bug: `sort-classes`: Group configuration order takes precedence over dependencies
### Describe the bug

Group configuration order takes precedence over dependencies, resulting in code that does not compile.

### Code example

Group configuration
```
"groups": [
  "private-property",
  "public-property"
]
```

Code
```ts
class Class {
  public a = 1;
  private b = this.a;
}
```

This code is considered invalid and requires `b` to be before `a` due to the groups order specified in the configuration. This leads to code that does not compile because `b` depends on `a`.

### ESLint version

8.57.0

v3.3.0

### Additional comments

This likely affects other rules that use both the notion of groups and dependencies.

### Validations

### LINKED ISSUE #242: Bug: `sort-classes`: Dependencies can cause undefined sorting behavior
### Describe the bug

The sorting algorithm sorts nodes by dependency relationships and other parameters (to sort by name, etc) in the same stage. This can lead to undefined sorting behavior:

### Code example

```ts
class Class {

  a = this.c

  b = 10

  c = 10

}
```

This code is considered valid when it should not, because `a` depends on `c`.
The issue arises from the fact that according to the sorting algorithm:
- `a` and `b` are independent, so `a` < `b` (alphabetical sort)
- `b` and `c` are independent, so `b` < `c` (alphabetical sort)

However, `a` depends on `c` so `c` < `a` (dependency-sort)

We get the following circular chain:
`a` < `b` < `c` < `a`

### ESLint version

8.57.0

v3.3.0

### Additional comments

Theoretically affects other rules that use dependencies as well.

### Validations

### LINKED ISSUE #243: Bug: `sort-classes`: Nested/Optional chained/Non-null asserted/Unary property references are not detected as dependencies
### Describe the bug

See code example.

### Code example

```ts
class Class {
  b = new Subject();
  a = this.b;
}
```
is valid: `a` is correctly detected as a dependency of `b`.

However:
```ts
class Class {
  b = new Subject();
  a = this.b.asObservable();
}
```
is not.

### ESLint version

8.57.0

v3.3.0

### Additional comments

Also includes optional chained non-null asserted and unary properties.

### Validations

### LINKED ISSUE #244: Bug: `sort-classes`: Arrow function property calls are not detected as dependencies
### Describe the bug

See code example.

### Code example

```ts
class Class {
  b = () => {
    return 1
  }
  a = this.b()
}
```

is considered invalid and requires `a` to be before `b` when it shouldn't, because b is a property.

```ts
class Class {
  a = this.b()
  b() {
    return 1
  }
}
```
is considered valid as expected.

### ESLint version

8.57.0

v3.3.0

### Additional comments

_No response_

### Validations

### LINKED ISSUE #245: Bug: `sort-classes`: Properties referenced in objects are not detected as dependencies
### Describe the bug

See code example

### Code example

```ts
class Class {
  b = 1
  a = {
    b: this.b
  }
}
```
This code is considered invalid and requires `a` to be before `b` when it shouldn't.

### ESLint version

8.57.0

v3.3.0

### Additional comments

_No response_

### Validations

hugop95:
### Description

Fixes #234.
Fixes #237.
Fixes #238.
Fixes #240.
Fixes #241.
Fixes #242.
Fixes #243.
Fixes #244.
Fixes #245.

This PR fixes various issues related to dependency detection.

### #234: static section incorrectly placed before static properties definition

`static` block's body dependencies will now be taken into account.

- [x] Tests added

### #237 Static properties referenced using MyClass.x are not detected as dependencies

Static properties referenced with the class's name keyword will now be detected as dependencies.

- [x] Tests added

### #238 Function properties's body dependencies are not properly detected when tests say they should

I have chosen to consider that function properties, just like methods, should not be affected by dependencies, because they don't depend on their order to compile or run correctly.
This means that the `works with body dependencies` tests become irrelevant.

These tests were not working as expected in the first place:

_Example 1_ (code considered valid)
https://github.com/azat-io/eslint-plugin-perfectionist/blob/305d88c6a7a901321eaed5313edc13a9989d59f8/test/sort-classes.test.ts#L2683-L2698

This test for example does nothing as the sort type is `alphabetical` or `natural`, and `groups` options is the default one, meaning that methods are expected to be after properties. 

This test is also confusing because the code https://github.com/azat-io/eslint-plugin-perfectionist/blob/305d88c6a7a901321eaed5313edc13a9989d59f8/rules/sort-classes.ts#L523-L525 shows that only properties will have their dependencies extracted, and not methods (which makes sense: methods compilation/runtime are not affected by their dependencies order). Is the test supposed to show that methods order is unaffected by their body dependencies, or is the implementation wrong and methods should be sorted after their body dependencies?

_Example 2_ (code considered valid)
https://github.com/azat-io/eslint-plugin-perfectionist/blob/305d88c6a7a901321eaed5313edc13a9989d59f8/test/sort-classes.test.ts#L2658-L2668
This test does nothing as well because `a` and `method` are both properties that are already sorted alphabetically/naturally, so `method` depending (or not) on `a` has no impact.

- [x] Dead tests removed
- [x] New tests added

### #240 Dependencies ignore the notion of static

This requires:
- Keeping track of whether `this` refers to the static context or not.
- Making a distinction between the name of a node and its "dependency name" (`a` and `static a` have the same node name, but a different dependency name). Today, dependencies are currently linked by the node's name.

- [x] Tests added

### #241 Group configuration order takes precedence over dependencies

Other rules that also use dependencies and groups have been updated.

- [x] Tests added

### #242 Dependencies can cause undefined sorting behavior

Node sorting has been separated into two stages:

1. Sorting by alphabetical/natural/line-length type
2. Sorting by dependencies

Dependency sorting uses topological sorting and detects circular dependencies.

Other rules that also use dependencies have been updated.

- [x] Tests added

### #243 Nested property references are not detected as dependencies

- [x] Tests added

### #244 Arrow function property calls are not detected as dependencies

Arrow function property references will be detected as dependencies, but not methods.

- [x] Tests added

### #245 Properties referenced in objects are not detected as dependencies

Detects dependencies in keys and values of objects.

- [x] Tests added
---

- [x] Bug fix

### DIFF
--- a/rules/sort-array-includes.ts
+++ b/rules/sort-array-includes.ts
@@ -146,7 +146,7 @@ export default createEslintRule<Options, MESSAGE_ID>({
             if (
               (options.groupKind !== 'mixed' && leftNum > rightNum) ||
               ((options.groupKind === 'mixed' || leftNum === rightNum) &&
-                isPositive(compare(left, right, options, true)))
+                isPositive(compare(left, right, options)))
             ) {
               context.report({
                 messageId: 'unexpectedArrayIncludesOrder',
--- a/rules/sort-array-includes.ts
+++ b/rules/sort-array-includes.ts
@@ -146,7 +146,7 @@ export default createEslintRule<Options, MESSAGE_ID>({
             if (
               (options.groupKind !== 'mixed' && leftNum > rightNum) ||
               ((options.groupKind === 'mixed' || leftNum === rightNum) &&
-                isPositive(compare(left, right, options)))
+                isPositive(compare(left, right, options, true)))
             ) {
               context.report({
                 messageId: 'unexpectedArrayIncludesOrder',
--- a/rules/sort-astro-attributes.ts
+++ b/rules/sort-astro-attributes.ts
@@ -191,7 +191,7 @@ export default createEslintRule<Options<string[]>, MESSAGE_ID>({
               if (
                 leftNum > rightNum ||
                 (leftNum === rightNum &&
-                  isPositive(compare(left, right, options, true)))
+                  isPositive(compare(left, right, options)))
               ) {
                 context.report({
                   messageId: 'unexpectedAstroAttributesOrder',
--- a/rules/sort-astro-attributes.ts
+++ b/rules/sort-astro-attributes.ts
@@ -191,7 +191,7 @@ export default createEslintRule<Options<string[]>, MESSAGE_ID>({
               if (
                 leftNum > rightNum ||
                 (leftNum === rightNum &&
-                  isPositive(compare(left, right, options)))
+                  isPositive(compare(left, right, options, true)))
               ) {
                 context.report({
                   messageId: 'unexpectedAstroAttributesOrder',
--- a/rules/sort-classes.ts
+++ b/rules/sort-classes.ts
@@ -15,13 +15,11 @@ import { getSourceCode } from '../utils/get-source-code'
 import { toSingleLine } from '../utils/to-single-line'
 import { rangeToDiff } from '../utils/range-to-diff'
 import { getSettings } from '../utils/get-settings'
-import { isPositive } from '../utils/is-positive'
 import { useGroups } from '../utils/use-groups'
 import { sortNodes } from '../utils/sort-nodes'
 import { makeFixes } from '../utils/make-fixes'
 import { complete } from '../utils/complete'
 import { pairwise } from '../utils/pairwise'
-import { compare } from '../utils/compare'
 
 type MESSAGE_ID = 'unexpectedClassesGroupOrder' | 'unexpectedClassesOrder'
 
@@ -126,7 +124,7 @@ type Group =
   | 'unknown'
   | string
 
-type Options = [
+export type Options = [
   Partial<{
     customGroups: { [key: string]: string[] | string }
     type: 'alphabetical' | 'line-length' | 'natural'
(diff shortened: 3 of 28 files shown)


azat-io:
Nice job! 👍

hugop95:
@azat-io Reopened and rebased here: https://github.com/azat-io/eslint-plugin-perfectionist/pull/246

PR #247, opened 2024-09-02
feat: `sort-sets`: New rule

hugop95:
### Description

This PR adds a new rule: `sort-sets`. This rule directly uses the `sort-array-includes`'s rule code, so its options are the same.

However, it doesn't verify that the `has` method is directly called afterward. The only side effect I can think of from sorting a set is that because the iteration order of `Set` is the insertion order, code that depends on this may get affected, but the `sort-map` rule has the same side effect and doesn't seem to take this into account either.

- [x] New Feature

### DIFF
--- a/docs/content/rules/sort-sets.mdx
+++ b/docs/content/rules/sort-sets.mdx
@@ -18,8 +18,6 @@ import { dedent } from 'ts-dedent'
 
 Enforce sorted set values.
 
-This rule works exactly like the [sort-array-includes](./sort-array-includes) rule.
-
 By keeping sets sorted, developers can quickly scan and verify the values, making the code more predictable and reducing the likelihood of errors. This practice simplifies debugging and enhances the overall clarity of the codebase.
 
 
--- a/docs/content/rules/sort-sets.mdx
+++ b/docs/content/rules/sort-sets.mdx
@@ -1,0 +1,237 @@
+---
+title: sort-sets
+description: Ensure your sets are sorted. Improve readability and maintain consistent code with this ESLint rule
+shortDescription: Enforce sorted sets
+keywords:
+  - eslint
+  - sort set
+  - eslint rule
+  - coding standards
+  - code quality
+  - javascript linting
+  - set sorting
+---
+
+import CodeExample from '../../components/CodeExample.svelte'
+import CodeTabs from '../../components/CodeTabs.svelte'
+import { dedent } from 'ts-dedent'
+
+Enforce sorted set values.
+
+This rule works exactly like the [sort-array-includes](./sort-array-includes) rule.
+
+By keeping sets sorted, developers can quickly scan and verify the values, making the code more predictable and reducing the likelihood of errors. This practice simplifies debugging and enhances the overall clarity of the codebase.
+
+
+## Try it out
+
+<CodeExample
+  alphabetical={dedent`
+    const getProductCategories = (product) => {
+      let electronics = new Set([
+        'Drone',
+        'Headphones',
+        'Keyboard',
+        'Laptop',
+        'Monitor',
+        'Mouse',
+        'Router',
+        'Smartphone',
+        'Smartwatch',
+        'Tablet',
+      ])
+      let accessories = new Set([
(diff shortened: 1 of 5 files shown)


docs/content/rules/sort-sets.mdx
```diff
@@ -0,0 +1,237 @@
...
+    },
+  ]}
+  type="config-type"
+  client:load
+  lang="ts"
+/>
+
+## Version
+
+This rule was introduced in [v3.4.0](https://github.com/azat-io/eslint-plugin-perfectionist/releases/tag/v3.4.0).
```
hugop95, line 232: Set by default to 3.4.0, to change if needed.

azat-io:
Do you think the order of the elements in Set matters?

I mean, how safe is it?

```js
let s1 = new Set([1,2,3])
let firstS1 = s1.values().next().value // 1

let s2 = new Set([3,2,1])
let firstS1 = s1.values().next().value // 3
```

docs/content/rules/sort-sets.mdx
azat-io: It's okay. 👍

hugop95:
> Do you think the order of the elements in Set matters?
> 
> I mean, how safe is it?

This potential issue already exists in `sort-maps`:

```ts
let map1= new Map([
  ['a', 'a'],
  ['b', 'b'],
]);
let firstMap1 = map1.values().next().value // a

let map2= new Map([
  ['b', 'b'],
  ['a', 'a'],
]);
let firstMap2 = map2.values().next().value // b
```

I assume that most cases using `Set` or `Map` are not affected by insertion sort order. However, if we choose to be as safe as possible, we could
- Either force the user to activate the rule rather than have it as default in `recommended` configs, because it can in very rare cases break code if code depends on insertion order
- Have `sort-sets` work similarly to `sort-array-includes`, and activate only if the `has` method is directly used afterward (or make this an option, allowing users to have it active even if `has` is not used).

In all cases, `sort-maps` is still affected by insertion order as of today.

azat-io:
Okay!

docs/content/rules/sort-sets.mdx
```diff
@@ -0,0 +1,237 @@
...
+  - set sorting
+---
+
+import CodeExample from '../../components/CodeExample.svelte'
+import CodeTabs from '../../components/CodeTabs.svelte'
+import { dedent } from 'ts-dedent'
+
+Enforce sorted set values.
+
+This rule works exactly like the [sort-array-includes](./sort-array-includes) rule.
```
azat-io, line 21: I think it's unnecessary.

rules/sort-sets.ts
```diff
@@ -0,0 +1,48 @@
...
+        (node.arguments[0]?.type === 'ArrayExpression' ||
+          (node.arguments[0]?.type === 'NewExpression' &&
+            'name' in node.arguments[0].callee &&
+            node.arguments[0].callee.name === 'Array'))
+      ) {
+        let elements =
+          node.arguments[0].type === 'ArrayExpression'
+            ? node.arguments[0].elements
+            : node.arguments[0].arguments
+        sortArray<MESSAGE_ID>(context, 'unexpectedSetsOrder', elements)
```
azat-io, line 44: Wouldn't it be better to copypaste the code?
In case there are differences in the rule in the future.

hugop95: I don't believe it is necessary:
The code in common manages the sorting of the array element which is common whether we use
- `new Set(['a', 'b', 'c'])`
- or `['a', 'b', 'c'].includes(...` / `new Array('a', 'b', 'c').includes(...`

The only thing that could change between the two rules are options which only impact when we choose to sort or not.

docs/content/rules/sort-sets.mdx
hugop95: Fixed

azat-io:
Released in [v3.4.0](https://github.com/azat-io/eslint-plugin-perfectionist/releases/tag/v3.4.0).

PR #251, opened 2024-09-02
feat: #97: Keep comments above elements being sorted

### LINKED ISSUE #97: If something should be sorted, move all above comments with that before a blank line or a non-comment entity is found

sort-object-types, sort-objects, sort-classes, ...?

### Describe the problem

When some entity is being sorted, move along with it not only the first comment, but all of them before a blank line or a non-comment entity is found

### Code example

(natural sorting is used)

### Example 1:

```ts
const test = {
  // yes 1
  // yes 2
  name: 'Immanuel',

  // first blablabla
  // second blablabla
  age: 79,

  // what a nice year
  bornYear: 1724,
}
```

to

```ts
const test = {
  // first blablabla
  // second blablabla
  age: 79,

  // what a nice year
  bornYear: 1724,

  // yes 1
  // yes 2
  name: 'Immanuel',
}
```

### Example 2:

```ts
const test = {
  // :)
  /** 
   * yeah
   * works
   * too
   */
  status: 'who_knows',
  key: 'value', // sticked!
  // something more
  c: 'good!',
}
```

to 

```ts
const test = {
  // something more
  c: 'good!',
  key: 'value', // sticked!
  // :)
  /** 
   * yeah
   * works
   * too
   */
  status: 'who_knows',
}
```

### Example 3:

```ts
class Test {
  name: string

  // first blablabla
  // second blablabla
  age: number
}
```

to

```ts
class Test {
  // first blablabla
  // second blablabla
  age: number

  name: string
}
```

### Example 4

```ts
type TTest = {
  b: string

  // wanna
  // be
  a: 123
}
```

to

```ts
type TTest = {
  // wanna
  // be
  a: 123

  b: string
}
```

### Additional comments

Really needed this functionality, so I have hastily created a workaround. Repository: https://github.com/datismoa/eslint-plugin-perfectionist. Does not break any of tests, tho

Would be great if we had an option that regulates such behaviour & the ability to choose in what entities and how (for example, count a multiline comment as an inhibitor) it works

### Validations

hugop95:
### Description

Fixes #97

- [ ] Tests added.

---

- [x] Bug fix


hugop95:
@azat-io RP ready for review!

azat-io (approved):

azat-io:
Wow. It looks great.
Thank you for your work!

PR #263, opened 2024-09-08
feat: `sort-objects`: Missing dependency detections

### LINKED ISSUE #262: Bug: `sort-objects`: Missing dependency detections
### Describe the bug

Very similar to #256.

The following elements are not detected as dependencies and are considered invalid when they should not:

- Negation operator.
- Optional chained elements.
- Non-null assertion operator.
- `as` expressions.
- Type assertion expressions.
- All sorts of ternary expressions.
- Template literal expressions.
- Spread elements.

### Code example

See above

### ESLint version

8.57.0

v3.5.0

### Additional comments

_No response_

### Validations

hugop95:
### Description

Fixes #262.

The dependency detection used is the same as the one used for `sort-variables-declarations`.

Furthermore, function bodies used to be taken into account for dependencies, which is not required for code to be able to compile and run correctly, so this has been removed.

- [x] Bug fix

### DIFF
--- a/test/sort-objects.test.ts
+++ b/test/sort-objects.test.ts
@@ -478,251 +478,651 @@ describe(ruleName, () => {
       },
     )
 
-    ruleTester.run(
-      `${ruleName}(${type}): works with complex dependencies`,
-      rule,
-      {
-        valid: [],
-        invalid: [
-          {
-            code: dedent`
-              let Func = ({
-                a,
-                b = a + c + d,
-                c,
-                d,
-              }) => {
-                // ...
-              }
-            `,
-            output: dedent`
-              let Func = ({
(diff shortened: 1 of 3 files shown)


test/sort-objects.test.ts
```diff
@@ -478,251 +478,651 @@ describe(ruleName, () => {
...
+                  messageId: 'unexpectedObjectsOrder',
+                  data: {
+                    left: 'b',
+                    right: 'c',
+                  },
                 },
-              },
-            ],
-          },
-          {
```
hugop95, line 519: Test removed because function bodies no longer get taken into account for dependencies. See `ignores function body dependencies` test.

PR #265, opened 2024-09-08
fix: #264: Template literal expressions are not detected as dependencies

### LINKED ISSUE #264: Bug: `sort-enums`: Template literal expressions are not detected as dependencies
### Describe the bug

See code below

### Code example

```ts
enum Enum {
  B = 'B',
  A = `${B}`,
}
```
This code is considered invalid and requires `A` to be before `B` when it should not.

### ESLint version

8.57.0

v3.5.0

### Additional comments

_No response_

### Validations

hugop95:
### Description

Fixes #264.

- [x] Bug fix


azat-io:
Thanks!

PR #267, opened 2024-09-13
feat(multiple-rules): #270: add `dependencyOrder` errors

### LINKED ISSUE #270: Feature(all classes supporting dependencies): Add a `dependencyOrderError` error
### Describe the rule

Today, errors are displayed between two adjacent nodes. In a standard alphabetical configuration with the nodes
`a`, `d`, `c`, `b`, the following members will get the following errors:
- c: `Expected "c" to come before "d".`
- b: `Expected "b" to come before "c".`

This works fine when nodes are not dependent on each other.

When a node B depends on a node A, and if B would have been after A if it wasn't depending on it, the algorithm puts B right before A to ensure proper compilation. However, the ESLint errors displayed can be a bit confusing, especially when A and B belong to different groups.

### Code example

```ts
class Class {

  a = { // A very complex object depending on `static z`
         // ...
         attribute: Class.z,
         // ...
         }
  
   // [...] Many other members

  someMethod() {
     // ...
 }

  static z = 0
}
```

This will display the following error on `z`: `Expected "z" (property) to come before "someMethod" (method)`
While technically true, the sole reason why `z` needs to be moved is because it needs to go before `a` for dependency reasons. The addition of the nodes group names in this particular case is also a bit misleading as it's not a question of group sorting here.

A better message would be:
`Expected dependency "z" to come before "a"`.

### Additional comments

_No response_

### Validations

hugop95:
Resolves #270.

Adds a new error for each rule that is impacted by dependency order:
- [x] `sort-classes`: `unexpectedClassesDependencyOrder`
- [x] `sort-enums`: `unexpectedEnumsDependencyOrder'` 
- [x] `sort-objects`: `unexpectedObjectsDependencyOrder` 
- [x] `sort-variable-declarations`: `unexpectedVariableDeclarationsDependencyOrder` 

Code 
```ts
class Class {

  a = { // A very complex object depending on `static z`
         // ...
         attribute: Class.z,
         // ...
         }
  
   // [...] Many other members

  someMethod() {
     // ...
 }

  static z = 0
}
```

#### Before for `z`

image

#### After for `z`

image

### DIFF
--- a/rules/sort-classes.ts
+++ b/rules/sort-classes.ts
@@ -6,7 +6,10 @@ import type {
   Modifier,
   Selector,
 } from './sort-classes.types'
-import type { SortingNodeWithDependencies } from '../utils/sort-nodes-by-dependencies'
+import {
+  getFirstUnorderedDependency,
+  type SortingNodeWithDependencies,
+} from '../utils/sort-nodes-by-dependencies'
 
 import {
   getOverloadSignatureGroups,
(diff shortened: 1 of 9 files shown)


azat-io (approved):

azat-io:
Looks awesome. Thank you!

PR #269, opened 2024-09-13
fix(sort-classes): #268: ignored nodes prevent their adjacent node from triggering linting

### LINKED ISSUE #268: Bug: `sort-classes`: Ignored nodes prevent their adjacent node from triggering linting
### Describe the bug

See code example below.

### Code example

Group configuration:

```
groups: ["property"]
```

```ts
class Class {
    a

    method() {
    }
    
    c
}
```

`method()` is part of the `unknown` group and will remain in place, but prevents `c` from triggering linting to correctly sort the class.

Any other class member that can trigger linting will correctly sort the whole class.

### ESLint version

8.57.0

v3.6.0

### Additional comments

_No response_

### Validations

hugop95:
### Description

Fixes #268.

When implementing the notion of ignored `unknown` group, ignored nodes were set to avoid creating linting errors: it's a bit counter-intuitive to tell the user that an ignored node should be before or after another node if it's supposed to be ignored.

Because linting errors are based on adjacent nodes comparison, classes that **only** have errors between ignored nodes and their adjacent nodes will not raise any error, and thus not trigger linting.

In practice, this likely shouldn't happen often and is fixed as soon as another error is detected in the class, as it will trigger linting and correctly sort it, so I don't believe this is a big issue (I detected it by accident while trying something else).

Custom groups with the `unsorted` sort type are unrelated to this issue. Only cases where a node has the `unknown` group may trigger this behavior.

This PR removes the checks that prevented errors from being raised between an ignored node and its adjacent nodes.

Because those ignored nodes now create ESLint error messages, a test had to be updated as it would now display more errors than before.

#### Possible improvements for another PR

```ts
class Class {
    b

    method() {
    }

    a
}
```

This PR will make the following errors appear:
- on `method`: `Expected "method" (unknown) to come before "b" (property)`
- on `a`: `Expected "a" (property) to come before "method" (unknown)`

The ideal message should appear only on `a` and say `Expected "a" to come before "b"`.
The reasoning behind is that if we compute that a node N1 needs to be placed before an adjacent ignored node NI, then there is at least one other node N2 before NI that should be placed after N1.

- [x] Bug fix

### DIFF
--- a/rules/sort-classes.ts
+++ b/rules/sort-classes.ts
@@ -682,11 +682,6 @@ export default createEslintRule<SortClassesOptions, MESSAGE_ID>({
         pairwise(nodes, (left, right) => {
           let leftNum = getGroupNumber(options.groups, left)
           let rightNum = getGroupNumber(options.groups, right)
-          // Ignore nodes belonging to `unknown` group when that group is not referenced in the
-          // `groups` option.
-          let isLeftOrRightIgnored =
-            leftNum === options.groups.length ||
-            rightNum === options.groups.length
 
           let indexOfLeft = sortedNodes.indexOf(left)
           let indexOfRight = sortedNodes.indexOf(right)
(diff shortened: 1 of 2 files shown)


azat-io:
Could you fix merge conflicts?

azat-io:
Once again please. There are new merge conflict

hugop95:
@azat-io Done.

azat-io (approved):

PR #271, opened 2024-09-14
feat(multiple rules): Do not sort unknown elements if unknown is not referenced in groups

hugop95:
### Description

The large majority of the actual code added involves documentation and tests.

Extends #220 to other rules that handle the `unknown` group.

When the `unknown` group is not entered by the user in the `groups` option, `unknown` group members will not be sorted and will remain where they are rather than go at the very bottom.

- [x] `sort-astro-attributes`
- [x] `sort-interfaces`
- [x] `sort-jsx-props`
- [x] `sort-object-types`
- [x] `sort-objects`
- [x] `sort-union-types`
- [x] `sort-vue-attributes`

The `sort-imports` rule takes into account other complex elements such as newline separation, making it very different than other rules. Left for another PR if needed.

Rules that had a default group configuration equal to `[]` now have `['unknown']` (otherwise, they would not sort anything by default).

Because all those rules rely on the same code to sort groups, a `sortNodesByGroups` function has been extracted, greatly reducing the amount of duplicate code.

No existing unit test was impacted.

- [x] Add tests

- [x] New Feature

### DIFF
--- a/docs/content/rules/sort-astro-attributes.mdx
+++ b/docs/content/rules/sort-astro-attributes.mdx
@@ -138,7 +138,14 @@ Controls whether sorting should be case-sensitive or not.
 
 ### groups
 
-<sub>default: `[]`</sub>
+<sub>
+  default:
+  ```
+  [
+    'unknown',
+  ]
+  ```
+</sub>
 
 Allows you to specify a list of Astro attribute groups for sorting. Groups help organize attributes into categories, prioritizing them during sorting. Multiple groups can be combined to achieve the desired sorting order.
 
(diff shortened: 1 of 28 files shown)


azat-io:
Could you fix merge conflicts?

hugop95:
@azat-io Done!

azat-io:
Once again please. There is new merge conflict.

hugop95:
@azat-io Done.

hugop95:
@azat-io Nope, this PR is changing this behavior!
- If `unknown` is not specified, nodes having the `unknown` group will stay in place and not move around (like `sort-classes` today).
- The default options however always include the `unknown` group.

If this is not the behavior that you want, then this PR can be closed, I will add the code refactor in another PR.

azat-io:
Honestly, I'm leaning towards the old behavior should remain. And the `unknown` group should be added to the end of the list if not explicitly specified.

But your refactoring looks very cool and could improve the project.

hugop95:
@azat-io Understood, I will make a new PR for the refactor in that case.

PR #276, opened 2024-09-15
fix(sort-classes): #275: partitionByComment/partitionByLine takes precedence over dependencies

### LINKED ISSUE #275: Bug(`sort-classes`): `partitionByComment`/`partitionByNewLine` takes precedence over `dependencies`
### Describe the bug

Affects other classes using dependencies and `partitionByComment`/`partitionByNewLine` as well: `sort-objects` and `sort-enums`

### Code example

Configuration: 

```
partitionByComment: "Part*"
```

```ts
class Class {
    b = this.a
    // Part1
    a
}
```

This code is considered valid when it should not because b depends on a.

### ESLint version

8.57.0

v3.6.0

### Additional comments

_No response_

### Validations

hugop95:
Fixes #275 for `sort-classes`, `sort-objects` and `sort-enums`.

The problem comes from the fact that the dependency check is done within each "block" partitioned by comments/line, rather than globally across the whole class/object/enum.

There is very little code change in reality! A large code block has been moved out of a `for` block; so the new indent generates a lot of line changes.

Note: the following code

```ts
class Class {
    b = this.a
    // Part1
    a
}
```

Will get sorted into

```ts
class Class {
    a
    // Part1
    b = this.a
}
```

This means that `b` is now part of the partitioned group defined by `Part1`! Ideally, we would want to get

```ts
class Class {
    a
    b = this.a
    // Part1
}
```

or

```ts
class Class {
    // Part1
    a
    b = this.a
}
```

Unfortunately, the current sorting implementation permutates elements between each others, which is why `a` gets permuted with `b`, and partition comments are never moved around.

This applies for `sort-objects` and `sort-enums` as well.

The current fix should do the job for now.

- [x] Bug fix

### DIFF
--- a/rules/sort-classes.ts
+++ b/rules/sort-classes.ts
@@ -688,42 +688,40 @@ export default createEslintRule<SortClassesOptions, MESSAGE_ID>({
             leftNum === options.groups.length ||
             rightNum === options.groups.length
 
-            let indexOfLeft = sortedNodes.indexOf(left)
-            let indexOfRight = sortedNodes.indexOf(right)
-            let firstUnorderedNodeDependentOnRight =
-              getFirstUnorderedNodeDependentOn(right, nodes)
-            if (
-              firstUnorderedNodeDependentOnRight ||
-              (!isLeftOrRightIgnored && indexOfLeft > indexOfRight)
-            ) {
-              let messageId: MESSAGE_ID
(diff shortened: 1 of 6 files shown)


azat-io:
Looks nice, could you fix merge conflicts?

azat-io (approved):

hugop95:
@azat-io Done!

PR #281, opened 2024-09-22
feat(sort-interfaces): add partitionByComment option

chirokas:
### Additional context

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

- [ ] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [ ] Read [contribution guidelines](https://github.com/azat-io/eslint-plugin-perfectionist/blob/main/contributing.md).

### DIFF
--- a/rules/sort-interfaces.ts
+++ b/rules/sort-interfaces.ts
@@ -202,6 +202,7 @@ export default createEslintRule<Options<string[]>, MESSAGE_ID>({
                 return accumulator
               }
 
+              let lastElement = accumulator.at(-1)?.at(-1)
               let name: string
 
               let { getGroup, defineGroup, setCustomGroups } = useGroups(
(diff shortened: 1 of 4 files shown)


rules/sort-interfaces.ts
```diff
@@ -178,6 +202,13 @@ export default createEslintRule<Options<string[]>, MESSAGE_ID>({
                 return accumulator
               }
 
+              if (partitionComment) {
```
hugop95, line 205: [nit] Really small nitpicking but I would put this check along with the one related to `options.partitionByNewLine` line 249 as those two concepts are related.

rules/sort-interfaces.ts
```diff
@@ -215,11 +238,22 @@ export default createEslintRule<Options<string[]>, MESSAGE_ID>({
...
-                getLinesBetween(sourceCode, lastElement, elementSortingNode)
-              ) {
+              let shouldPartition = false
+              if (partitionComment) {
+                let comments = getCommentsBefore(element, sourceCode)
+                shouldPartition = hasPartitionComment(
+                  partitionComment,
+                  comments,
+                )
+              } else if (options.partitionByNewLine) {
```
hugop95, line 248: I think that this `else if` prevents `partitionByComment` and `partitionByNewLine` from co-existing together, what do you think about

```ts
let lastElement = accumulator.at(-1)?.at(-1)
if (
  (partitionComment &&
    hasPartitionComment(
      partitionComment,
      getCommentsBefore(element, sourceCode),
    )) ||
  (options.partitionByNewLine &&
    lastElement &&
    getLinesBetween(sourceCode, lastElement, sortingNode))
) {
  accumulator.push([])
}
```

hugop95 (approved):

azat-io (approved):

PR #290, opened 2024-09-22
fix(all-rules): Ignores whitespaces in node names

### LINKED ISSUE #181: Bug: [`perfectionist/sort-intersection-types`] indentation can impact rule interpretation
### Describe the bug

In the screenvideo below, you can see 2 strange behaviours:
- the rule works but is broken when changing the indentation
- when inverting 2 types as requested by the error message, a new error message is raised and contradicts the previous one

### Code example

Project (for testing – remove the `eslint-disable` comment): https://github.com/OlivierZal/melcloud-api/blob/fd854ab272e3b818b0570878d9815e938481d7ec/src/facades/interfaces.ts#L61

```
setHolidayMode: (({
    enabled,
    from,
    days,
  }: {
    enabled?: true
    from?: string
    days: number
  }) => Promise<FailureData | SuccessData>) &
    (({
      enabled,
      from,
      to,
    }: {
      enabled?: true
      from?: string
      to: string
    }) => Promise<FailureData | SuccessData>) &
    (({ enabled }: { enabled: false }) => Promise<FailureData | SuccessData>)
```

### ESLint version

v9.7.0

v3.0.0

### Additional comments

Here is a screenvideo:

https://github.com/user-attachments/assets/b8c1cb15-589a-493e-a52f-2f846e006b1f

### Validations

hugop95:
Resolves #181 .

### Description

Today with the `alphabetical` or `natural` sort, the following code is valid

```ts
type T =
{ a: string } &
{ b: string }
```

but this one is not:

```ts
type T =
{ a: string } &
{  b: string }
```

The issue lies in the fact that `{  b` is "lower" than `{ a` because of the whitespace before `b`.
Giving the possibility for users to ignore whitespaces can solve this issue.  

### Changes

This PR changes the following:
- [x] All whitespace-like characters (including `\n`) will be ignored in node names.
- [x] Add tests. Refactored `compare` file to reach 100% code coverage.

- [x] New Feature

### DIFF
--- a/utils/compare.ts
+++ b/utils/compare.ts
@@ -37,24 +37,11 @@ export let compare = (
 ): number => {
   let orderCoefficient = options.order === 'asc' ? 1 : -1
   let sortingFunction: (a: SortingNode, b: SortingNode) => number
-
-  let formatString: (value: string) => string
-  if (options.type === 'line-length') {
-    formatString = (string: string) => string
-  } else {
-    formatString = (value: string) => {
-      let valueToCompare = value
-      if (options.ignoreCase) {
-        valueToCompare = valueToCompare.toLowerCase()
-      }
-      return valueToCompare.replaceAll(/\s/g, '')
-    }
-  }
-
   let nodeValueGetter =
     options.nodeValueGetter ?? ((node: SortingNode) => node.name)
 
   if (options.type === 'alphabetical') {
+    let formatString = getFormatStringFunc(!!options.ignoreCase)
     sortingFunction = (aNode, bNode) =>
       formatString(nodeValueGetter(aNode)).localeCompare(
         formatString(nodeValueGetter(bNode)),
@@ -67,11 +54,13 @@ export let compare = (
       }
       return string
     }
-    sortingFunction = (aNode, bNode) =>
-      naturalCompare(
+    sortingFunction = (aNode, bNode) => {
+      let formatString = getFormatStringFunc(!!options.ignoreCase)
+      return naturalCompare(
         prepareNumeric(formatString(nodeValueGetter(aNode))),
         prepareNumeric(formatString(nodeValueGetter(bNode))),
       )
+    }
   } else {
     sortingFunction = (aNode, bNode) => {
       let aSize = aNode.size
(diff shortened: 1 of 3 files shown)


azat-io:
Are you sure this should be a separate option and not the default behavior?

hugop95:
@azat-io I am trying to find cases where a user would **not** want to ignore whitespaces. It's possible that there are none, in which case this is better to have this as a default option indeed. @OlivierZal, what do you think?

OlivierZal:
I would say it rather should be the default behavior since IMO `perfectionist` should not care about indentation, which belongs to `stylistics` concerns.

hugop95:
> I would say it rather should be the default behavior since IMO `perfectionist` should not care about indentation, which belongs to `stylistics` concerns.

This is unfortunately more than just a stylistic concern: all whitespace are removed, meaning that for

```ts
type T = 
  'I am a string' |
  'I am a         string'
```
the two members of the type are now the same.

I know that my example is not really representative of an issue someone would have in practice, but I'm wondering if we are not potentially impacting something else.

OlivierZal:
Oh I see, so the question is: are there cases where someone wants spaces to be taken into account, and what's the most widly used and accepted practice in the dev community in order to decide which is the default.

hugop95:
@OlivierZal Pretty much. On one end, I think that "stylistic" related issues like

```ts
type T =
{ a: string } &
{  b: string }
```

can be fixed by prettifying tools like `prettier`. On the other, if these tools are not used, this affects sorting rules.

OlivierZal:
My problem is precisely that I use prettier: even if I keep a working  indentation, prettier will change it.

OlivierZal:
Based on our conversation, I would vote for defaulting the behavior to ignore white spaces.

utils/compare.ts
```diff
@@ -38,10 +38,18 @@ export let compare = (
...
+  let formatString: (value: string) => string
+  if (options.type === 'line-length') {
+    formatString = (string: string) => string
+  } else {
+    formatString = (value: string) => {
+      let valueToCompare = value
+      if (options.ignoreCase) {
+        valueToCompare = valueToCompare.toLowerCase()
+      }
+      return valueToCompare.replaceAll(/\s/g, '')
```
OlivierZal, line 50:
```ts
return valueToCompare.replaceAll(/\s/g, '')
```

or just

```ts
return valueToCompare.trim()
```

?

hugop95: Trimming would be the ideal! 

Unfortunately, it doesn't solve your case: `'{  b: string }'.trim()` is still `'{  b: string }'`

OlivierZal: oh, right...

azat-io:
Awesome fix, thank you!

PR #293, opened 2024-09-22
refactor(multiple-rules): adds `sort-nodes-by-groups` function to reduce code duplication

hugop95:
### Description

All rules handling groups (with the exception of `sort-classes` with a new CustomGroup API) do the same thing when it comes to sorting nodes by groups. This PR refactor codes to reduce code duplication for those rules.

- [x] Other


azat-io (approved):

azat-io:
❤️

PR #294, opened 2024-09-22
feat(multiple-rules): #138: Adds `matcher` option

### LINKED ISSUE #138: Feature: `sort-imports` custom-groups Regex match

sort-imports

### Describe the problem

To match imports in custom groups with Regex (instead of / along with) minimatch.
Because minimatch doesn't support wildcard matching for parent paths. So i have to do something like:

```ts
// eslint.config.mjs
const longParentPath =
	"{.," +
	Array.from({ length: 10 }, (_, i) => "../".repeat(i + 1).slice(0, -1)).join(",") +
	"}";

// returns: {.,..,../..,../../..,../../../..,../../../../..,../../../../../..,../../../../../../..,../../../../../../../..,../../../../../../../../..,../../../../../../../../../..}

// ...
"custom-groups": {
	value: {
		"AdmoPro-Basics": `${longParentPath}/Basics/**`,
	},
	type: {
		"AdmoPro-Basics": `${longParentPath}/Basics/**`,
	},
}
// ...
```

to be able to match:
```ts
import { i18n } from "../../../Basics/Language";
// and
import { i18n } from "../../../../../Basics/Language";
```
and other similar imports in files with various depths.

### Code example

Provided.

### Additional comments

_No response_

### Validations

hugop95:
Resolves #138 .

### Description

This PR adds the following to all rules supporting `customGroups` or `partitionByComment`:
- A `matcher: 'minimatch' | 'regex'` option (by default `minimatch`) that dictates what matcher is used for pattern matching for both the `internalPattern` and `customGroups` options.
- For `sort-imports`, the default `ignorePattern` option will be `['^~/.*']` if the matcher is `regex`.

Don't be afraid of the number of line changes, most are tests and documentation.

Rules affected:
- [x] `sort-array-includes`
- [x] `sort-astro-attributes`
- [x] `sort-classes`
- [x] `sort-enums`
- [x] `sort-exports`
- [x] `sort-imports`
- [x] `sort-interfaces`
- [x] `sort-interesection-types`
- [x] `sort-jsx-props`
- [x] `sort-maps`
- [x] `sort-named-exports`
- [x] `sort-object-types`
- [x] `sort-objects`
- [x] `sort-sets`
- [x] `sort-union-types`
- [x] `sort-variable-declarations`
- [x] `sort-vue-attributes`

### Important note

- Entering an invalid RegExp expression will throw an error when a node is evaluated.
- Regex options are not supported: the regex string (`"xxxx"` and not `/xxxx/`) must be entered for serialization reasons.

- [x] New Feature

### DIFF
--- a/test/sort-classes.test.ts
+++ b/test/sort-classes.test.ts
@@ -6646,6 +6646,104 @@ describe(ruleName, () => {
         },
       ],
     })
+
+    ruleTester.run(
+      `${ruleName}: allows to use regex matcher for element names in custom groups with old API`,
+      rule,
+      {
+        valid: [
+          {
+            code: dedent`
+              class Class {
+                iHaveFooInMyName: string
+                meTooIHaveFoo: string
+                a: string
+                b: string
+              }
+            `,
+            options: [
+              {
+                type: 'alphabetical',
+                matcher: 'regex',
+                groups: ['unknown', 'elementsWithoutFoo'],
+                customGroups: {
+                  elementsWithoutFoo: '^(?!.*Foo).*$',
+                },
+              },
+            ],
+          },
+        ],
+        invalid: [],
+      },
+    )
+
+    ruleTester.run(
+      `${ruleName}: allows to use regex matcher for element names in custom groups with new API`,
+      rule,
+      {
+        valid: [
+          {
+            code: dedent`
+              class Class {
+                iHaveFooInMyName: string
+                meTooIHaveFoo: string
+                a: string
+                b: string
+              }
+            `,
+            options: [
+              {
+                type: 'alphabetical',
+                matcher: 'regex',
+                groups: ['unknown', 'elementsWithoutFoo'],
+                customGroups: [
+                  {
+                    groupName: 'elementsWithoutFoo',
+                    elementNamePattern: '^(?!.*Foo).*$',
+                  },
+                ],
+              },
+            ],
+          },
+        ],
+        invalid: [],
+      },
+    )
+
+    ruleTester.run(
+      `${ruleName}: allows to use regex matcher for decorator names in custom groups with new API`,
+      rule,
+      {
+        valid: [
+          {
+            code: dedent`
+              class Class {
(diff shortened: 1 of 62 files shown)


azat-io:
It seems we cannot use regex in ESLint config, config should be serializable.

hugop95:
Hi @azat-io !

Technically yes, you need to make sure that all your options are serializable, so you can not pass functions or complex objects in your config (in reality, passing your functions as string and `eval` them could probably work).

However, you can pass RegExp as strings fine: if we take the example of the issue to match any element like
```ts
import { i18n } from "../../../Basics/Language";
import { i18n } from "../../../../../Basics/Language";
```
a regular expression matching those is `(?:\.\.\/)+Basics\/Language`. You can them stringify it to pass it in the JSON config like so:
```ts
customGroups: {
    value: {
        myCustomGroup: '(?:\\.\\.\\/)+Basics\\/Language',
        },
    },
```
and use the `RegExp` constructor: `return new RegExp(pattern).test(nodeName)`.

While flags (`i`, `u`, `g`, etc) are not directly handled by this method, I don't think this is too much of a problem in reality for our use cases.

What do you think?

azat-io:
@hugop95 Okay, it looks like good idea 👍

azat-io (commented):
Perhaps we should add this option to all rules, is there an option to set `customGroups`?

hugop95:
> Perhaps we should add this option to all rules, is there an option to set `customGroups`?

@azat-io Sure!

test/sort-classes.test.ts
```diff
@@ -7034,30 +7133,59 @@ describe(ruleName, () => {
...
+                  messageId: 'unexpectedClassesOrder',
+                  data: {
+                    left: 'd',
+                    right: 'a',
+                  },
                 },
+              ],
+            },
+          ],
+        })
```
hugop95, line 7160: No changes here, just re-indented.

azat-io:
It looks awesome, thank you!

Issue #296, opened 2024-09-30
Bug: perfectionist/sort-objects customGroups strange behavior


deathemperor:
### Describe the bug

custom groups are not sorted as configured

config:
```js
        "perfectionist/sort-objects": [
          "error",
          {
            ignoreCase: false,
            groups: ["dates"],
            customGroups: {
              dates: ["startDate", "endDate", "start_date", "end_date"],
            },
          },
        ],
```

![image](https://github.com/user-attachments/assets/60bb4a87-5d7f-4de3-b878-5fc129003968)

### Code example

https://github.com/deathemperor/perfectionist-custom-groups

### ESLint version

8.57.0

### ESLint Plugin Perfectionist version

3.6.0

### Validations

- [X] Read the [docs](https://perfectionist.dev).
- [X] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

hugop95:
Hi @deathemperor,

The expected result with your configuration is  
```ts
const test = {
  end_date: "2021-01-02",
  endDate: "2021-01-02",
  start_date: "2021-01-01",
  startDate: "2021-01-01",
};
```

which seems to be working as intended.

What result are you expecting?

deathemperor:
thanks for the explanation, it works. I find it confusing setting up group & customGroups.

What does `top` array mean in this example from the docs?

```ts
{
   groups: [
    'top',
     'unknown',
   ],
  customGroups: {       
    top: ['id', 'name'] 
  }                     
 }
```

hugop95:
@deathemperor I understand that customGroups and groups can be easily misconfused: they represent very different thing.

### groups
`groups` represent the sorting order of the elements: elements matching the `top` group will be put above elements matching the `unknown` group.

### customGroups
`customGroups` are actually **custom groups definitions**. With the exception of `sort-classes` which has a more advanced API, keys of the `customGroups` field are the **names** of groups that you should put in `groups`, and the values are:
- if a string, what the element's name need to match to be part of the group.
- if an array, the element's name need to match at least one of the elements of the array to be part of the group. Matching is by default using https://github.com/isaacs/minimatch.

With the example above, this means that for the object
```js
const a = {
    a: 'a',
    id: 'myId',
    _name: 'name',
    b: 'b'
}
```

- `a` matches the `unknown` group (the default group given when an element doesn't match any group referenced in `groups`).
[...]

Issue #295, opened 2024-09-27
Bug: sort-imports rule does not sort external-type and sibling-type correctly in the same group


jasonneylon:
### Describe the bug

Hi,

When I configure the `sort-imports` rule with a group for type imports where external types come before sibling types the rule is not followed. Instead, eslint reports valid code as violating the rule.

Thanks 🙏 

### Code example

eslint config:

```javascript
    'perfectionist/sort-imports': [
      "error",
      {
        "groups": [
          [
            "builtin-type",
            "external-type",
            "index-type",
            "internal-type",
            "parent-type",
            "sibling-type",
            "type"
          ]
        ]
      }
    ]
```

When I lint the following code with this rule configuration:

```typescript
import type { ReactNode } from 'react';
import type { Test } from './othertypes';
```

I get a lint error as follows:
```
/Users/jneylon/babbel/lessonnine/perfectionisteslint-config-test/src/types.ts
  2:1  error  Expected "./othertypes" to come before "react"  perfectionist/sort-imports

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
```

Given the rule configuration above I would expect this code to be valid.

There is a minimal reproduction in this repo: https://github.com/jasonneylon/perfectionist-eslint-config-test/tree/main

### ESLint version

8.57.1

### ESLint Plugin Perfectionist version

3.7.0

### Validations

- [X] Read the [docs](https://perfectionist.dev).
- [X] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

hugop95:
Hello @jasonneylon !

It seems that there might be a typo in your configuration: you nest your groups in another array
```json
    'perfectionist/sort-imports': [
      "error",
      {
        "groups": [
          [
             ...
          ]
        ]
      }
    ]
```

instead of

```json
    'perfectionist/sort-imports': [
      "error",
      {
        "groups": [
             ...
        ]
      }
    ]
```

What your initial configuration does is put all of your groups into the same group, which means that your two imports are now sorted alphabetically.

jasonneylon:
Hi @hugop95 

What I'm trying to do is sort imports first into a non-types imports and then into type imports, so there is more configuration in our actual use case than included above.

[...]

PR #297, opened 2024-10-01
docs: Improves documentation regarding `groups` and `customGroups` options

hugop95:
### Description

Groups and custom groups behavior are very powerful but complex to understand features. The documentation on those is a bit lacking, as shown by the recent issues #295 and #296.

This PR aims to improve the documentation for all rules regarding the `groups` and `customGroups` options.

Ideally should be included:
- [x] All rules handling `groups` handle the predefined `unknown` group.
- [x] The `unknown` group matches all elements that do not match any group referenced in `groups`.
- [x] If the `unknown` group is not referenced in `groups`, it is automatically added at the end of `groups`.
- [x] `groups` items are `string` or `string[]`. Explain how nested group blocks (`string[]` items) work. Emphasize that the order of items in nested group blocks is irrelevant.
- [x] Keys of `customGroups` are groups names that can then be referenced in `groups`.
- [x] Values of `customGroups` are either `string` or `string[]`. Explain how both work. Emphasize that the order of items in`string[]` values is irrelevant. Add that the `matcher` option allows choosing how to match elements.
- [x] `customGroups` matching **always** takes precedence over predefined groups matching.

This list is non-exhaustive, feel free to add things if needed.

### Example per rule

For the following rules, add a detailed example of a relatively complex, sorted class/object... using `groups` and `customGroups`, preferably with all allowed value types possible. Show what group each element matches in a comment.

- [x] `sort-astro-attributes`
- [x] `sort-interfaces`
- [x] `sort-object-types`
- [x] `sort-objects`
- [x] `sort-union-types`
- [x] `sort-vue-attributes`

### Additional documentation fixes

- Removes some duplicated content related to `groups` for some rules.
- Adds missing documentation related to `partitionByNewLine` for some rules.
- Fixes some typos related to `matcher` option.
- Adds missing default values for options such as `matcher`.

- [x] Documentation update


azat-io (approved):

azat-io:
Looks awesome, thank you!

PR #299, opened 2024-10-03
feat(sort-classes): adds `elementValuePattern` filter for properties

### LINKED ISSUE #298: Feature: property assignment group rule

sort-classes

### Describe the problem

through customGroups, it is possible to create rules that match the name of the property/method/decorators, would it be possible to add a new combination to evaluate the assignment?

developing components in angular, we can create signals, I would like to group them by type of signals and order them alphabetically, as per the example below

### Code example

```typescript
// my signals order sample-> inject -> input -> model -> signal -> computed -> output

class Component {
  anotherService = inject(ServiceA)
  barService = inject(ServiceB)
  fooAService = inject(ServiceC)
  
  anotherSome = input<string>()
  bar = input('default')
  foo = input<number>(0)

  isSeleted= model<boolean>(false)
  
  isLoading = signal<boolean>(false)
    
  hasBoo = computed(() => !!this.bar())
  hasFoo = computed(() => !!this.foo())
  
  click = output<void>()
  close = output<void>()

  //..
}
```

### Additional comments

_No response_

### Validations

hugop95:
### Description

Resolves #298 

### Changes

Adds a `elementValuePattern` option for custom groups, allowing users to match **non-function property** values.

- [x] Add tests
- [x] Update documentation 

- [x] New Feature
- [x] Documentation update

### DIFF
--- a/docs/content/rules/sort-classes.mdx
+++ b/docs/content/rules/sort-classes.mdx
@@ -481,6 +481,7 @@ interface CustomGroupDefinition {
   selector?: string
   modifiers?: string[]
   elementNamePattern?: string
+  elementValuePattern?: string
   decoratorNamePattern?: string
 }
 ```
@@ -497,6 +498,7 @@ interface CustomGroupBlockDefinition {
       selector?: string
       modifiers?: string[]
       elementNamePattern?: string
+      elementValuePattern?: string
       decoratorNamePattern?: string
   }>
 }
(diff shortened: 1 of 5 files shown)


OlivierZal:
Wouldn't `namedFunctionPattern` be more accurate and explicit about the scope of this pattern?

hugop95:
@OlivierZal This feature allows users to target he values of non-function properties. Based on the already existing `elemenNamePattern`, this seems like a better fit.

OlivierZal:
Alrigth 😊

hugop95:
`propertyValuePattern` could also work.

OlivierZal:
Yes, but you're right: `elementValuePattern` is consistent with other options.

azat-io (approved):

PR #304, opened 2024-10-05
feat: adds `sort-decorators`

hugop95:
### Description

A rule to sort decorators exists here: https://github.com/ElsiKora/ESLint-Plugin-Sort-Decorators.

However, it doesn't benefit from the `eslint-plugin-perfectionist` options ecosystem (possibility to group elements, add custom groups, partition by comment, etc.).

Decorators order may sometimes matter (see section below), so giving users the option to keep that order in those cases is necessary.

Additionally, this implementation has some differences related to what we consider an `accessor decorator` to be (explained below as well). 

## Changes

This PR adds a new rule: `sort-decorators`. It can sort:
- Class decorators.
- Class property decorators.
- Method decorators.
- Method parameter decorators.
- Class accessor decorators.

### Options

- `type`
- `order`
- `ignoreCase`
- `groups`
- `customGroups`
- `partitionByComment`
- `matcher`
- `sortOnParameters`: boolean (by default `true`)
- `sortOnProperties`: boolean (by default `true`)
- `sortOnAccessors`: boolean (by default `true`)
- `sortOnMethods`: boolean (by default `true`)
- `sortOnClasses`: boolean (by default `true`)

`partitionByNewLine` was not added as I don't think there is a use case in separating decorators with a new line. This can be easily added if necessary.

There is at the moment only one predefined group: `unknown`.

### Errors

There are 2 errors:
- `unexpectedDecoratorsGroupOrder`: when two adjacent decorators do not belong to the same group and are unordered, with a more detailed error.
- `unexpectedDecoratorsOrder`: when two adjacent decorators belong to the same group but are unordered.

## ⚠️ Should this be included in recommended configs?

Decorators are evaluated from top to bottom. While unusual, decorator order may matter in some projects. Activating these rules without proper configuration **CAN** cause runtime errors.

At the moment, there are 2 other rules that may change behavior unexpectedly through sorting and that are present in recommended configs:
- `sort-maps`
- `sort-sets`

(If users depend on insertion order, although it is unusual).

I think that if we choose not to include this rule into `recommended`, then both of the above rules should not be included as well.

- [x] New Feature
- [x] Documentation update


rules/sort-class-decorators.ts
```diff
@@ -0,0 +1,251 @@
...
+
+      accumulator.at(-1)!.push(sortingNode)
+
+      return accumulator
+    },
+    [[]],
+  )
+
+  let sortedNodes = formattedMembers
+    .map(nodes => sortNodesByGroups(nodes, options))
```
OlivierZal, line 227: `.flatMap()` ?

OlivierZal:
Couldn't there be a conflict of precedence (at least from the user point of view) with `decoratorNamePattern` defined under `sort-classes`.`customGroups`?

Maybe something to pay attention in `sort-classes` and `sort-xxx-decorators`.

rules/sort-class-decorators.ts
hugop95: Thanks, I keep forgetting about that 😄

hugop95:
> Couldn't there be a conflict of precedence (at least from the user point of view) with `sort-classes`.`customGroups` relying on a `decoratorNamePattern`?
> 
> Maybe something to pay attention in `sort-classes` and `sort-xxx-decorators` documentations.
> 
> By the way I ❤️ these new rule proposals.

@OlivierZal This is fine: in `sort-classes`, `decoratorNamePattern` is used to target class members that have at least one decorator with the entered pattern. The order of decorators does not matter.

OlivierZal:
Got it. 

Then config is important because unlike class members, order matters – like it matters for arrays, because decorators apply sequentially.

hugop95:
> Got it. 
> 
> Then config is important because unlike class members, order matters – like it matters for arrays, because decorators apply sequentially. 

Yes, I mention it at the bottom of the PR, regarding whether the rules should be in recommended configs.

OlivierZal:
@hugop95, I know it's a draft. I've read your current tests: I've not found the other decorator syntaxes:

```ts
@a @b
something
```

and

```ts
@a @b something
```

Doesn't it worth to test them? Do you want to allow keeping the original syntax as an option? (not necessarily in the first version of these rules).

hugop95:
@OlivierZal 
This implementation will not affect elements style-wise: decorators are permuted, so no whitespace/newline is added/removed.

However, it can still be interesting to ensure that this does not get broken, so I'll just edit a couple of tests to have an inline syntax and this should be enough.

docs/content/rules/sort-accessor-decorators.mdx
```diff
@@ -0,0 +1,263 @@
+---
+title: sort-accessor-decorators
+description: Enforce sorting of accessor decorators for for improved readability and maintainability. Use this ESLint rule to keep your accessor decorators well-organized
```
OlivierZal, line 3: `for for`

hugop95:
@azat-io 

Should we include this in recommended configs?

Decorators are evaluated from top to bottom. While unusual, decorator order may matter in some projects. Activating these rules without proper configuration **CAN** cause runtime errors.

At the moment, there are 2 other rules that may change behavior unexpectedly through sorting and that are present in recommended configs:
- `sort-maps`
- `sort-sets`

(If users depend on insertion order, although it is unusual).

I think that if we choose not to include this rule into `recommended`, then both of the above rules should not be included as well.

azat-io (changes requested):
Thank you for your contributions!

I think it would be better to combine all these rules into one `sort-decorators` rule?

I think most of the time people don't want to configure each rule separately and for decorators it makes sense to combine everything.

azat-io:
Other than that, I like everything. It's a very cool idea. But it would be nice to combine these rules into one.

hugop95:
Alright, I'll regroup that in one rule and add options to enable/disable it on `methods`, `classes` etc.

OlivierZal (approved):
All good for me @hugop95, great work!

azat-io (commented):
@hugop95 Looks good, awesome work!

1. Could you please also combine test files into one?
2. And we need to add new rule to rule list in `readme.md`

hugop95:
@azat-io Done. Do you think we should release this in a new major version (v4.0.0)? We could add some breaking changes as well in it:

- https://github.com/azat-io/eslint-plugin-perfectionist/issues/309
- Maybe remove the old `customGroups` API in `sort-custom-classes` ?

OlivierZal (approved):

azat-io:
@hugop95 Okay, let's do it in the major version. We'll just have to think about the plan for this release and think about what other changes we want to have.

azat-io:
I have created a `next` branch, could you please direct your pull request to this branch?

hugop95:
@azat-io Done!

azat-io:
Awesome, thank you!

OlivierZal:
Great achievement @hugop95!

hugop95:
> Great achievement @hugop95!

Thank you for the feedback ! Always appreciated.

PR #308, opened 2024-10-09
fix: `partitionByComment` doesn't Respect Comment Boundaries in `sort-union-types`

### LINKED ISSUE #306: Bug:  `partitionByComment` Doesn't Respect Comment Boundaries in `sort-union-types`
### Describe the bug

When using the `partitionByComment: true` option in `sort-union-types`, the union type members are expected to be sorted within groups that are separated by comments. However, the current behavior ignores comment boundaries and sorts the entire union type globally, even across different comment sections.

### Code example

Given this input:

```tsx
type T =
  // Group 1
  | Z
  | B
  // Group 2
  | C
  | A;
```

Expected output (sorting within comment partitions):

```
type T =
  // Group 1
  | B
  | Z
  // Group 2
  | A
  | C;
```

Actual output (comments ignored, globally sorted):

```
type T =
  // Group 1
  | A
  | B
  // Group 2
  | C
  | Z;
```

### ESLint version

8.57.0

3.8.0

### Additional comments

`partitionByComment` works in others, for example `sort-exports`

### Validations

hugop95:
Resolves #306 ... But not completely.

Also applies to `sort-intersection-types`

## The problem

```ts
type T =
  // Group 1
  | Z
  | B
  // Group 2
  | C
  | A;
```
`// Group 1` and `// Group 2` are not detected as partition comments

### Why

Partition comments are comments detected `directly` before a node: here: `Z` and `C`.
However, `|` is a token, so partition comments are not considered comments of `Z` and `C`.

### "True" fix

Write code as follows:

```ts
type T =
  // Group 1
  Z |
  B |
  // Group 2
  C |
  A;
```

However, some tools will make automatically format this as what we had in the first place.

## Proposed solution in this PR

When detecting comments for a given node:
- If the node has comments immediately before, OK, behavior as usual.
- If the node has a `|` token (`&` for `sort-interesection-types`), get comments before that token rather than the node.

## The problems with this solution

Unfortunately, doing this has consequences on other comments-related features (only if writing `|` before the node):
Comments associated to a node will not be moved around. This means that

```ts
type T =
  // Not a partition comment, follow Z!
  | Z
  | B
```
gets autofixed to
```ts
type T =
// Not a partition comment, follow Z!
| B  
| Z
```

The reason is that nodes being re-ordered do **not** include the `|` token with them (and as such, comments above that token).

Is it possible to make a fix to include those tokens and their comments in the sorting algorithm? Sure, but in my opinion, it's not worth the potential things that can break (especially if mixing both writing styles in the same union/intersection types).

- [x] Bug fix


hugop95:
@azat-io For info!

hugop95:
I realize this is not a full fix, but that's the simplest thing I can think of at the moment.

rules/sort-intersection-types.ts
```diff
@@ -261,7 +261,7 @@ export default createEslintRule<Options, MESSAGE_ID>({
             (partitionComment &&
               hasPartitionComment(
                 partitionComment,
-                getCommentsBefore(type, sourceCode),
+                getCommentsBefore(type, sourceCode, '&'),
```
azat-io, line 264: I think it looks like a hack. It might be worth refactoring in the future.

azat-io (approved):

PR #321, opened 2024-10-15
release v4.0.0

### LINKED ISSUE #341: Release: v4.0.0
We have a new major release planned. Let's discuss the plans here.

- [x] Remove Rules for Astro, Vue and Svelte
- [x] New API for Sorting Classes Custom Groups
- [x] New Default Value for Groups in sort-classes
- [x] New Rules
- [x] Regex vs Minimatch?
- [x] Problems Comparing Languages in localCompare
- [x] TS Config Support
- [x] Use TS Config When Defining Import Groups?

#321
## Remove Rules for Astro, Vue and Svelte

We deprecated the following rules:

- `sort-vue-attributes`
- `sort-svelte-attributes`
- `sort-astro-atttributes`

This was done as part of task #175. They should be removed in the next release.

Additional good news, we will have a big reduction in the size of JS scripts on the documentation site. Because Shiki loads large scripts for parsing these languages.

## New API for Sorting Classes Custom Groups

Support for the new API was added a long time ago. In this version, we remove the deprecated API for installing custom class groups.

This task was done by @hugop95, thank you! #331

The API now looks like this:

```js
customGroups: [                               
	[                                           
	    {
		groupName: 'constructor',
	        selector: 'constructor',
	    },
	    {                                   
		groupName: 'input-properties',
	        selector: 'property',
	        modifiers: ['decorated'],
	        decoratorNamePattern: 'Input',
	    },
	    {                                         
	        groupName: 'output-properties',
	        selector: 'property',
	        modifiers: ['decorated'],
	        decoratorNamePattern: 'Output',
	    },
    ],
]
```

## New Default Value for Groups in `sort-classes`

We've greatly improved how the new default grouping of class members will look.

#320

```js
groups: [
	'index-signature',
    ['static-property', 'static-accessor-property'],
    ['static-get-method', 'static-set-method'],
    ['protected-static-property', 'protected-static-accessor-property'],
    ['protected-static-get-method', 'protected-static-set-method'],
    ['private-static-property', 'private-static-accessor-property'],
    ['private-static-get-method', 'private-static-set-method'],
    'static-block',
    ['property', 'accessor-property'],
    ['get-method', 'set-method'],
    ['protected-property', 'protected-accessor-property'],
    ['protected-get-method', 'protected-set-method'],
    ['private-property', 'private-accessor-property'],
    ['private-get-method', 'private-set-method'],
    'constructor',
    ['static-method', 'static-function-property'],
    ['protected-static-method', 'protected-static-function-property'],
    ['private-static-method', 'private-static-function-property'],
    ['method', 'function-property'],
    ['protected-method', 'protected-function-property'],
    ['private-method', 'private-function-property'],
    'unknown',
]
```

## New Rules

New rules have been added:

- `sort-decorators` #304
- `sort-heritage-clauses` #334

These rules look really cool. Thank you so much @hugop95

## Regex vs Minimatch?

We have two matchers to describe patterns for comment-splitters and for defining custom groups.

**minimatch:**

```js
{
	matcher: 'minimatch',
	customGroups: {
		value: {
			reactLibraries: ['react-*']
		}
	}
}
```

**regex:**

```js
{
	matcher: 'regex',
	customGroups: {
		value: {
			reactLibraries: ['^react-.*']
		}
	}
}
```

Minimatch seems more convenient, but it also brings additional dependency to the project, which is not a good thing.

I would like to explore how other ESLint plugins work and official ESLint rules, what do they use in similar cases, regular expressions or something like minimatch?

Then our plugin could look consistent within the plugin list.

Perhaps we should delete minimatch. Or make regex the default, and recognize minimatch as deprecated. Or leave everything as it is now.

## Problems Comparing Languages in `localeCompare`

@fkworld came to us and highlighted a very important problem that arises when comparing strings in different languages.

I think that this problem should be fixed and a feature for selecting the project language should be added.

#328

Thank you!

## TS Config Support

I added type support in the 3.0.0 release, with @JoshuaKGoldberg additionally helping me with this. 

#90

However, people continue to run into problems when using the plugin inside eslint.config.ts. #338

The reason for the problem is due to the fact that we are using TypeScript ESLint types instead of ESLint types:

```ts
import type {
	ClassicConfig,
	FlatConfig,
} from '@typescript-eslint/utils/ts-eslint'
```

I think this should definitely be dealt with and test using the plugin with configs that use TypeScript.

## Use TS Config When Defining Import Groups?

We had @gavvvr come to us with a suggestion to use tsconfig.json to detect whether an import is internal or external.

#317

But I'm afraid it might affect the performance of the plugin. I think we should do some research on this.

## Do you have any ideas?

If you have any additional ideas, I'd love to hear them. 

Thank you to all the contributors, you are incredibly awesome! 💜

azat-io:
Close #341


hugop95:
@azat-io What do you think about removing support for the old `sort-classes` `customGroups` API? (using objects to define custom groups)

azat-io:
@hugop95 Yeah, maybe that makes sense.

PR #324, opened 2024-10-15
feat(multiple-rules): adds `newlinesBetween`

### LINKED ISSUE #146: Feature: `sort-object-types` add support for `newlines-between`

sort-object-types

### Describe the problem

There is no support for `newlines-between`, so it's not possible to automatically add newlines between custom groups.

### Code example

For example, we have a type like this:
```
type Person = {
  onBirthday: () => void;
  name: string;
  onAgeChange: (age: number) => void;
  age?: number;
  surname: string;
  birthday?: Date;
}
```

I expect it to be sorted like this using custom groups:
```
type Person = {
  name: string;
  surname: string;
  
  age?: number;
  birthday?: Date;
  
  onAgeChange: (age: number) => void;
  onBirthday: () => void;
}
```

### Additional comments

_No response_

### Validations

### LINKED ISSUE #329: Bug(sort-imports): Invalid newlines auto-fix under very special conditions
### Describe the bug

Newline auto-fixes can in very specific configurations result in code that does not compile. See code example.

### Code example

Configuration:

`"newlinesBetween": "always"` or `"never"`.

Code:

```ts
import a from 'a'

 import b from 'b'
```

Notice the whitespace before the second import.

The resulting code is:
```ts
import a from 'a' import b from 'b'
```

### ESLint version

8.57.0

3.9.1

### Additional comments

_No response_

### Validations

hugop95:
Resolves #146
Fixes #329

### Description

This PR aims to extend the `newlinesBetween` option of `sort-imports` to other rules.

Affected rules:
- [x] `sort-interfaces`
- [x] `sort-object-types`
- [x] `sort-objects`
- [x] `sort-union-types`

### Compatibility with `partitionByNewline`

`partitionByNewLine` and `newlinesBetween` are incompatible:

Only one option at a time can be active.

An error if thrown otherwise: `The 'partitionByNewLine' and 'newlinesBetween' options cannot be used together`

- [x] New Feature
- [x] Documentation update

### DIFF
--- a/docs/content/rules/sort-interfaces.mdx
+++ b/docs/content/rules/sort-interfaces.mdx
@@ -214,6 +214,18 @@ interface User {
 
 Each group of members (separated by empty lines) is treated independently, and the order within each group is preserved.
 
+### newlinesBetween
+
+<sub>default: `'ignore'`</sub>
+
+Specifies how new lines should be handled between interface groups.
+
+- `ignore` — Do not report errors related to new lines between interface groups.
+- `always` — Enforce one new line between each group, and forbid new lines inside a group.
+- `never` — No new lines are allowed between interface members.
+
+This options is only applicable when `partitionByNewLine` is `false`.
+
 ### groupKind
 
 <sub>default: `'mixed'`</sub>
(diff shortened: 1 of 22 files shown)


azat-io:
Thank you! Let's make all pull requests that aren't related to ASAP bug fixes into the `next` branch? In order for them to be included in the 4.0.0 release.

azat-io (approved):

PR #327, opened 2024-10-16
fix(sort-switch-case): fixes expressions being ignored

### LINKED ISSUE #326: Bug(`sort-switch-case`): Switch cases cannot be sorted when switch expression isn't a value
### Describe the bug

First off, thank you for addressing the other issue I raised so quickly! When testing the new release with the fix, I encountered a new issue with `sort-switch-case`: if the switch expression is anything other than a simple value or variable (basically anything that requires evaluating such as a function call, accessing an object property, or a math expression), the cases are not recognized as being out of order.

### Code example

**Input**
```js
switch ("a".concat("", "b")) {
  case "b":
    break;
  case "c":
    break;
  case "ab":
    break;
}

const obj = {
  letter: "a",
};

switch (obj.letter) {
  case "b":
    break;
  case "c":
    break;
  case "a":
    break;
}

switch (1 + 2) {
  case 3:
    break;
  case 1:
    break;
  case 2:
    break;
}
```

**Expected Output**
```js
switch ("a".concat("", "b")) {
  case "ab":
    break;
  case "b":
    break;
  case "c":
    break;
}

const obj = {
  letter: "a",
};

switch (obj.letter) {
  case "a":
    break;
  case "b":
    break;
  case "c":
    break;
}

switch (1 + 2) {
  case 1:
    break;
  case 2:
    break;
  case 3:
    break;
}
```

**Actual Output**
(No change)
```js
switch ("a".concat("", "b")) {
  case "b":
    break;
  case "c":
    break;
  case "ab":
    break;
}

const obj = {
  letter: "a",
};

switch (obj.letter) {
  case "b":
    break;
  case "c":
    break;
  case "a":
    break;
}

switch (1 + 2) {
  case 3:
    break;
  case 1:
    break;
  case 2:
    break;
}
```

### ESLint version

9.12.0

3.9.0

### Additional comments

_No response_

### Validations

hugop95:
Fixes #326 

### Description

This PR makes sure that `switch(true)` is still disabled, but enables other expressions to be entered.

- [x] Bug fix

### DIFF
--- a/rules/sort-switch-case.ts
+++ b/rules/sort-switch-case.ts
@@ -93,7 +93,8 @@ export default createEslintRule<Options, MESSAGE_ID>({
 
       let sourceCode = getSourceCode(context)
 
-      let isDiscriminantIdentifier = node.discriminant.type === 'Identifier'
+      let isDiscriminantTrue =
+        node.discriminant.type === 'Literal' && node.discriminant.value === true
       let isCasesHasBreak = node.cases
         .filter(caseNode => caseNode.test !== null)
         .every(
@@ -107,7 +108,7 @@ export default createEslintRule<Options, MESSAGE_ID>({
             ),
         )
 
-      if (isDiscriminantIdentifier && isCasesHasBreak) {
+      if (!isDiscriminantTrue && isCasesHasBreak) {
         let nodes = node.cases.map<SortSwitchCaseSortingNode>(
           (caseNode: TSESTree.SwitchCase) => {
             let name: string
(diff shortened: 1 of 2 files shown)


azat-io (approved):

azat-io:
LGTM. Do you think it makes sense to create a 3.9.1 release or include this in version 4.0.0?

hugop95:
@azat-io 3.9.1 I think:

This is in my opinion a fix that every user should have, including the ones that will not directly move to 4.0.0.

azat-io:
Thanks!
Released in [v3.9.1](https://github.com/azat-io/eslint-plugin-perfectionist/releases/tag/v3.9.1).

PR #328, opened 2024-10-18
feat: add locales option for all rules

fkworld:
### Description

My colleague and I encountered differences when using the sort-objects rule.

After troubleshooting, we found that the discrepancies occurred during the **localeCompare** process. My VSCode display language is **English**, while my colleague's VSCode display language is **Simplified Chinese**.

A simple reproduction example:

**1. add test.ts**
```
const locale = new Intl.DateTimeFormat().resolvedOptions().locale;
console.log('locale', locale);

const result = 'CPO'.localeCompare('运营商');
console.log('"CPO" localeCompare "运营商" result', result);
```

**2. vscode config display language Simplified Chinese, run test.ts in vscode terminal**
output:
```
locale zh-CN
"CPO" localeCompare "运营商" result 1
```

**3. vscode config display language English, run test.ts in vscode terminal**
output:
```
locale en-US
"CPO" localeCompare "运营商" result -1
```

**4. run test.ts in system terminal, not vscode terminal**
output:
```
locale zh-CN
"CPO" localeCompare "运营商" result 1
```

### Additional context

**I'm not sure if it's reasonable to use the locale from the operating environment.**

My perspective is: when I use sort-objects, I want it to return the same order in any environment. Therefore, I added a fixed locales parameter of `en-US`.

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/utils/common-json-schemas.ts
+++ b/utils/common-json-schemas.ts
@@ -13,6 +13,21 @@ export let orderJsonSchema: JSONSchema4 = {
   type: 'string',
 }
 
+export let localesJsonSchema: JSONSchema4 = {
+  oneOf: [
+    {
+      type: 'string',
+    },
+    {
+      items: {
+        type: 'string',
+      },
+      type: 'array',
+    },
+  ],
+  description: 'Specifies the sorting locales.',
+}
+
 export let ignoreCaseJsonSchema: JSONSchema4 = {
   description: 'Controls whether sorting should be case-sensitive or not.',
   type: 'boolean',
--- a/docs/content/guide/getting-started.mdx
+++ b/docs/content/guide/getting-started.mdx
@@ -145,6 +145,7 @@ In settings you can set the following options:
 - `ignoreCase` — Ignore case when sorting.
 - `ignorePattern` — Ignore sorting for elements that match the pattern.
 - `specialCharacters` — Control whether special characters should be kept, trimmed or removed before sorting. Values can be `'keep'`, `'trim'` or `'remove'`.
+- `locales` - The locales of sorting. Values can be a string with a BCP 47 language tag, or an array of such strings. See [String.prototype.localeCompare() - locales](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#locales).
 - `partitionByComment` — Partition the sorted elements by comments. Values can be `true`, `false` or regexp pattern.
 - `partitionByNewLine` — Partition the sorted elements by new lines. Values can be `true` or `false`.
 
--- a/docs/content/rules/sort-array-includes.mdx
+++ b/docs/content/rules/sort-array-includes.mdx
@@ -157,6 +157,16 @@ Controls whether special characters should be trimmed, removed or kept before so
 - `'trim'` — Trim special characters when sorting alphabetically or naturally (e.g., “_a” and “a” are the same).
 - `'remove'` — Remove special characters when sorting (e.g., “/a/b” and “ab” are the same).
 
+### locales
+
+<sub>default: `'en-US'`</sub>
+
+Specifies the sorting locales. See [String.prototype.localeCompare() - locales](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#locales).
+
+- `string` — A BCP 47 language tag (e.g. `'en'`, `'en-US'`, `'zh-CN'`).
+- `string[]` — An array of BCP 47 language tags.
+
+**Note:** Only effective when `type` is `alphabetical`.
 
 ### groupKind
 
--- a/docs/content/rules/sort-classes.mdx
+++ b/docs/content/rules/sort-classes.mdx
@@ -192,6 +192,17 @@ Controls whether special characters should be trimmed, removed or kept before so
 - `'trim'` — Trim special characters when sorting alphabetically or naturally (e.g., “_a” and “a” are the same).
 - `'remove'` — Remove special characters when sorting (e.g., “/a/b” and “ab” are the same).
 
+### locales
+
+<sub>default: `'en-US'`</sub>
+
+Specifies the sorting locales. See [String.prototype.localeCompare() - locales](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#locales).
+
+- `string` — A BCP 47 language tag (e.g. `'en'`, `'en-US'`, `'zh-CN'`).
+- `string[]` — An array of BCP 47 language tags.
+
+**Note:** Only effective when `type` is `alphabetical`.
+
 ### partitionByComment
 
 <sub>default: `false`</sub>
--- a/docs/content/rules/sort-decorators.mdx
+++ b/docs/content/rules/sort-decorators.mdx
@@ -186,6 +186,17 @@ Controls whether special characters should be trimmed, removed or kept before so
 - `'trim'` — Trim special characters when sorting alphabetically or naturally (e.g., “_a” and “a” are the same).
 - `'remove'` — Remove special characters when sorting (e.g., “/a/b” and “ab” are the same).
 
+### locales
+
+<sub>default: `'en-US'`</sub>
+
+Specifies the sorting locales. See [String.prototype.localeCompare() - locales](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#locales).
+
+- `string` — A BCP 47 language tag (e.g. `'en'`, `'en-US'`, `'zh-CN'`).
+- `string[]` — An array of BCP 47 language tags.
+
+**Note:** Only effective when `type` is `alphabetical`.
+
 ### partitionByComment
 
 <sub>default: `false`</sub>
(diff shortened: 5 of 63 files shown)


hugop95:
Hi @fkworld! Thank you for this PR. You put a finger on an important issue.

I have a different approach toward it, however:

- I think that fixing a given locale for everyone by default is a breaking change, so this should probably be put for the next major version.
- This should be accompanied by the possibility for users to:
  - Override the default locale.
  - Go back to an operating environment-based sorting behavior.

I think it's interesting to examine the behavior for the `natural-sort` type as well, and add tests for those.

azat-io:
Thank you for your pull request! It's important feature.

I agree with @hugop95.

Could you please:

1. Add locale to `settings` (https://perfectionist.dev/guide/getting-started#settings) with default value `en-US`
2. Add docs
3. Add some tests
4. Move PR from `main` to `next` branch

fkworld:
@hugop95 @azat-io I understand, I'm working on it.

fkworld:
# Simple or Complex?

I agree that this is a breaking change and should be put for the next major version.

However, when I tried to implement locale as settings/options, I discovered it would introduce a significant level of complexity.

The complexity manifests in:

1. Almost all rules need `locale` option.
2. Almost all rules need to have tests written for at least two different locales.
    - For unit testing, it is usually a combination of alphabetical + asc and line-length + desc, not all combinations.
    - Or only write unit tests for the `compare` method.
3. An invalid `locale` will cause `localeCompare` to throw an error.
4. `locale` does not affect `naturalCompare`.
5. Conflict with the `specialCharacters` option, when `specialCharacters` is set to `trim`/`remove`, the `locale` cannot take effect in practice.
6. Conflict with `naturalCompare`

```ts
["z", "ä"].sort((a, b) => a.localeCompare(b, "en-US"))        // ['ä', 'z']
["z", "ä"].sort((a, b) => naturalCompare(a,b))                // ['z', 'ä']
["z", "ä"].sort((a, b) => a.charCodeAt(0) - b.charCodeAt(0))  // ['z', 'ä']
```

---

I propose three solutions

## solution-1: Add locale option for all rules. (complex)

1. Add `locale` option for all rules.
2. Add unit test for all rules. Or add unit test for `compare` method.
3. Add docs for all rules.

## solution-2: Hardcode the parameter 'en-US'. (simple)

1. Hardcode localeCompare with parameter 'en-US'.
5. Add docs for `recommended-alphabetical`. Inform developers that when implementing alphabetical sorting, the underlying method `localeCompare("...", "en-US")` is used.

## solution-3: Implement an alphabetical sorting method, like `naturalCompare`. (simple)

I have reviewed the source code of the natural-compare-lite package, which compares characters based on their charCodeAt, essentially the UTF-16 code unit.

1. Implement an alphabetical sorting method that also uses `charCodeAt` for comparison.
2. Add docs for `recommended-alphabetical`. Inform developers that when implementing alphabetical sorting, the underlying method `charCodeAt` is used.

---

I prefer solution-3 for the following reasons:

1. Sorting non-English alphabets is not a high-frequency operation.
2. Most developers will not pay too much attention to the locale.
3. It follows the same implementation principle as `naturalCompare`.

I am eager to hear your thoughts on this matter! @hugop95 @azat-io

hugop95:
@fkworld Thank you for this very detailed explanation. I personally think that solution 3 is solution 2 in a superior way, so let's focus on solutions 1 and 3.

In my opinion, this all depends on whether "Sorting non-English alphabets is a high-frequency operation or not" as you mention. I assume that it's not, but can't tell from experience, so I'll trust your words on that!

In that case, fixing a default locale (or solution 3) for everyone is fine. And if needed tomorrow, this can be expanded by rethinking about how to handle locale-based sorting.

Solution 3 then sounds the easiest way to accomplish that, while having the benefit to be similar to the `natural` compare implementation.

I have written below some comments about the points you mention, and a [Question] at the bottom.

# Complexity

## Almost all rules need to have tests written for at least two different locales

I think that unit tests in `compare` are not enough if we want to make sure that all rules support this option. 

All possible combination of options are not tested as you said, but this is fine (and make tests actually readable!). The objective is really to test everything "globally" while making sure edge cases are covered. So in theory, for each rule, we would only need tests for the `alphabetical` sorting type:
  - A test verifying that the `locale` option is supported.
  - A test verifying that we have two different sorting results with two different locales. 
 
I agree that this still a lot to do though, but if we decide to allow users to choose their locales tomorrow, we will have to go through all of that.

## An invalid locale will cause localeCompare to throw an error.

I think this is fine, an error is thrown if users enter an invalid regex with the `matcher: regex` option. It's the user's responsibility to enter a valid locale.

## locale does not affect naturalCompare.

I don't think we can do much here, other than searching for an alternative or implementing that ourselves. As long as it's mentioned in the documentation, this should be fine.

## Conflict with the `specialCharacters` option, when `specialCharacters` is set to trim/remove, the locale cannot take effect in practice.

We can make this combination incompatible and thrown an error in that case.

# Solution 1

## Add docs for all rules.

I think most users would want to set up a locale globally for all rules, so it's probably not useful to mention that this option exists per-rule level.

# Solution 3

If I understand correctly, this would be similar to fixing the locale for everyone, but through a `charCodeAt` sorting method rather than `localeCompare`. 

## Implement an alphabetical sorting method that also uses `charCodeAt` for comparison.

[Question] What do you think would be general impact for english-language based codebases? Are we expecting almost no change in comparison to the current `localeCompare` sorting method for this?

hugop95:
@fkworld Thank you for your investigation.

What I'm going to say is just how I see it, @azat-io may I have a different point of view.

I would personally push for option 1 if we had more visibility on the return on investment, but I know it's obviously quite a bit more work than option 2 and understand if you don't want to/have time to spend on implementing and testing all of this.

On one hand, I think it's neat for a sort-focused plugin to offer those powerful options. On the other hand, I don't know if it would be really used.

If setting a default value (`en-US`) doesn't affect the codebase of any user (to the point where they would want to fix a particular locale that is), then I think it's fine to go with option 2.

However, I think we should be ready to revert this until option 1 is implemented if it turns out that this affects more users than we expected.

azat-io:
I generally agree with @hugop95.

I find option 1 more interesting. And if you have the desire and strength, we can implement it.

Nevertheless, option 2 also has a place. At least as a temporary solution.

Which option are you leaning towards?

fkworld:
@hugop95 @azat-io Thank you for describing your perspectives. I will work on the solution-1 plan, which is expected to take 3 to 5 days.

I will do:

- [x] Add `locale` to `Settings`
- [x] Add `locale` to `AlphabeticalCompareOptions`
- [] Add `localeJsonSchema`
- [ ] Add `locale` to all rule `Options`, `defaultOptions` and `schema`
- [ ] Add all rule test with option `{ type: 'alphabetical', ignoreCase: false, order: 'asc', locale: 'zh-CN' }`
- [ ] Add `locale` to all rule docs

utils/common-json-schemas.ts
```diff
@@ -13,6 +13,11 @@ export let orderJsonSchema: JSONSchema4 = {
   type: 'string',
 }
 
+export let localeJsonSchema: JSONSchema4 = {
+  description: 'Specifies the sorting locale.',
+  type: 'string',
+}
```
OlivierZal, line 19: Don't we want to be stricter here?

```
export let localeJsonSchema: JSONSchema4 = {
  description: 'Specifies the sorting locale.',
  type: 'string',
  pattern: '^[a-z]{2}(-[A-Z]{2})?$'
}
```

hugop95 (approved):
Looks great! 💯

OlivierZal:
Awesome

OlivierZal (approved):

fkworld:
Hi @hugop95 @azat-io, I've finished the work on this feature and would appreciate your review on the PR.

Here are some notes:

1. `locale` to `locales`, consistent with localeCompare - locales parameters.
2. `locales` type is `NonNullable<Intl.LocalesArgument>`.
    1. `locales` type is `string | string[]` in es5, and `Intl.LocalesArgument` in es2020. I choose es2020.
    2. Without `NonNullable` will cause a type error. Because `Intl.LocalesArgument` includes `undefined`, will be lost by Required<Partial<...>>.

You can reproduce it in the rules/sort-heritage-clauses.ts file:
image

Or you can reproduce it with the following simple example:
```ts
type Obj = {
  a?: string
  b: string | undefined
}

type PartialA = Partial<Obj>
// {
//   a?: string | undefined;
//   b?: string | undefined;
// }

type RequiredA = Required<Obj>
// {
//   a: string;
//   b: string | undefined;
// }

type RequiredPartialA = Required<Partial<Obj>>
// {
//   a: string;
//   b: string;
// }
```

3. I use `['你好', '世界', 'a', 'A', 'b', 'B']` for unit test. 你好 means hello, and 世界 means world. This is simple data to differentiate between localeCompare en-US, localeCompare zh-CN, and charCodeAt.
```ts
const resultCharCodeAt = ['你好', '世界', 'a', 'A', 'b', 'B'].toSorted(
  (a, b) => a.charCodeAt(0) - b.charCodeAt(0),
)
// ['A', 'B', 'a', 'b', '世界', '你好']

const resultEnUS = ['你好', '世界', 'a', 'A', 'b', 'B'].toSorted(
  (a, b) => a.localeCompare(b, 'en-US'),
)
// ['a', 'A', 'b', 'B', '世界', '你好']

const resultZhCN = ['你好', '世界', 'a', 'A', 'b', 'B'].toSorted(
  (a, b) => a.localeCompare(b, 'zh-CN'),
)
// ['你好', '世界', 'a', 'A', 'b', 'B']
```

azat-io:
Looks awesome! Thank you! ❤️

azat-io:
Released in [v4.0.0](https://github.com/azat-io/eslint-plugin-perfectionist/releases/tag/v4.0.0).

If you'd like, you can support the release with a retweet:
https://x.com/azat_io_en/status/1858853714238271530

PR #331, opened 2024-10-18
feat(sort-classes)!: removes support for old `customGroups` API

hugop95:
### Description

`sort-classes`'s `customGroups` option currently supports:
- an object (old API): like other rules, each key is the group's name, and its value the pattern to match the value of the class's member.
- an array (new API) with more options.

Migrating from the old to the new API is very straightforward.

Old API:
```ts
{
  "key1": "value1",
  "key2": "value2"
}
```

New API:
```ts
[
  {
    "groupName": "key1",
    "elementNamePattern": "value1"
  },
  {
    "groupName": "key2",
    "elementNamePattern": "value2"
  }
]
```

This PR removes support for the old API and the code associated and adds a section to the documentation to help users who have not yet switched to the new API to migrate.

- [x] Documentation update
- [x] Other

### DIFF
--- a/docs/content/rules/sort-classes.mdx
+++ b/docs/content/rules/sort-classes.mdx
@@ -14,6 +14,7 @@ keywords:
 ---
 
 import CodeExample from '../../components/CodeExample.svelte'
+import Important from '../../components/Important.astro'
 import CodeTabs from '../../components/CodeTabs.svelte'
 import { dedent } from 'ts-dedent'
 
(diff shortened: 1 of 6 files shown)


docs/content/rules/sort-classes.mdx
```diff
@@ -472,6 +473,34 @@ abstract class Example extends BaseExample {
 
 ### customGroups
 
+<Important title="ℹ️ Migrating from the old API">
```
azat-io, line 476: Let's remove the emoji from the documentation site after all, shall we? They're not used anywhere else.

Everything else looks good to me.

azat-io (changes requested):

docs/content/rules/sort-classes.mdx
hugop95: @azat-io Fixed in b9d52b44ef5f298de55225a526e8566511db702b.

azat-io (approved):

PR #337, opened 2024-10-23
fix: Improve ending commas/semicolon behavior

### LINKED ISSUE #336: Bug(multiple-rules): Inline element sorting may result in fix that breaks compilation/runtime
### Describe the bug

This issue is related to inline sorting, and affects:
- `sort-classes`
- `sort-interfaces`
- `sort-object-types`
- `sort-imports`
- `sort-exports`
- `sort-switch-case` 

### Code example

```js
type Type = {
  b, a
}
```

This results in the following autofix, which breaks compilation:

```js
type Type = {
  a b,
}
```

Inlining `}` with the last member will make the rule work however:

```js
type Type = {
    a, b }
```
Even worse:

```js
type Type = {
  b,a
}
```

will get sorted into

```js
type Type = {
  ba,
}
```

Other example with `sort-classes`:

```ts
class Class {

    method(){}property: string

}
```

This is ugly, but it compiles and runs! It gets formatted into

```ts
class Class {

    property: stringmethod(){}

}
```

### ESLint version

8.57.0

3.9.1

### Additional comments

_No response_

### Validations

hugop95:
Fixes #336.

Very little code changes outside of tests: 
image

## Affected rules

- `sort-classes`
- `sort-interfaces`
- `sort-object-types`
- `sort-imports`
- `sort-exports`
- `sort-switch-case` 

## Description

Today, the way commas and semicolons are handled for a given node actually depends on multiple factors:
-  It depends on the rule.
    - A node in `sort-objects` will never include `,`, while a node in `sort-interfaces` will. This is how the AST parser works. This means that rules that natively ignore commas and semicolons are not affected (`sort-enums` for example).
    - In `sort-object-types` and `sort-interfaces`, nodes will include `;` and `,`. 
    - In `sort-object-types`, commas and semicolons are ignored for computing the length of a node with the `line-length` sorting type, but not in `sort-interfaces`. https://github.com/azat-io/eslint-plugin-perfectionist/blob/b2afcfd8c1b39d99d1a0613a6a5a0a6c15bccf8e/rules/sort-object-types.ts#L252-L260
- To determine if a given node will move with its associated semicolon/comma (if it has one), we check the token after that semicolon/comma. If it exists and is on the same line as the node, the semicolon/comma associated to the node will not move.

The last point is problematic, because it fixes a problem in very specific cases (and not the one mentioned in the issue, for example).

## Proposal

This PR does the following:
- Make the `line-length` sorting type always ignore ending semicolons and commas for all rules, similarly to what `sort-object-types` does today.
- Adds a semicolon to nodes that would **potentially** require one during inline sorting (see the logic below).

## How to know if we need to add a semicolon to a node after sorting

Let's say that a node **B** needs to take the place of a node **A** after sorting.

First, **B** should indicate whether it **requires an ending semicolon if inline with a following node**. 
In `sort-classes` for example, `properties` require an ending semicolon if inline, while `non-abstract methods` don't.

If **B** requires an ending semicolon, we should check whether if ends with a semicolon (or a comma) or not (as that semicolon will be brought with it when moving).

If **B** does not end with a semicolon/comma, check whether **A** (the place that **B** will take) has at least one inline token (other than a semicolon/comma) after it. If it does, this means that we should add a semicolon after **B** for safety. 

There are cases where this process adds a semicolon when it is not actually necessary, but this does not break compilation and doesn't affect runtime. Furthermore, this is likely to be autocorrected by another ESLint rule and is easy to implement, so that's a quick-win in my book.

Also, let's not forget that although these bugs can be very problematic (breaking compilation/runtime), they are not likely to be reproduced that often in reality due to all the conditions required for those bugs to appear, so let's keep it not too complicated!

## `sort-classes`: Types of nodes and whether they require an ending semicolon or not when inline with a following node

- Methods:
  - If abstract: ✅
  - Else: ❌
- Properties/Accessors: ✅ (including function properties!)
- TS Index signature: ✅
- Static block: ❌

## Examples

<details>

#### Example 1

❌
```ts
abstract class Class {
    myFunction(){}static{}abstract myAbstractFunction()
}
```
✅
```ts
abstract class Class {
    static{}abstract myAbstractFunction();myFunction(){}
}
```

#### Example 2

❌
```ts
abstract class Class {
    constructor(){}[key: string]: string
}
```
✅
```ts
abstract class Class {
    [key: string]: string;constructor(){}
}
```

</details>

## Consequences

- Only one test (duplicated 2 times, for a different sorting type) in `sort-object-types` had its output fixed (See commit n°2: 0e79823b77a841598416bf49df1276e1ac3bd6f5)

## Additional changes

- [Refactor] Uses `sourceCode.getText()` rather than slicing.
- Avoids making a permutation autofix when a node is already at its expected place.

## Remaining to do

Add inline tests for all rules! (even the ones not affected by #336)

- [x] `sort-array-includes`
- [x] `sort-classes`
- [x] ~~`sort-decorators`~~ No separator
- [x] `sort-enums`
- [x] `sort-exports`
- [x] `sort-heritage-clauses`
- [x] `sort-imports`
- [x] `sort-interfaces`
- [x] ~~`sort-jsx-props`~~ No separator
- [x] `sort-named-exports`
- [x] `sort-named-imports`
- [x] `sort-object-types`
- [x] `sort-objects`
- [x] `sort-sets`
- [x] `sort-switch-case`
- [x] `sort-union-types`
- [x] `sort-variable-declarations`

- [x] Bug fix

### DIFF
--- a/rules/sort-array-includes.ts
+++ b/rules/sort-array-includes.ts
@@ -178,7 +178,7 @@ export let sortArray = <MessageIds extends string>(
             name:
               element.type === 'Literal'
                 ? `${element.value}`
-                : sourceCode.text.slice(...element.range),
+                : sourceCode.getText(element),
             size: rangeToDiff(element, sourceCode),
             node: element,
             group,
--- a/rules/sort-array-includes.ts
+++ b/rules/sort-array-includes.ts
@@ -179,7 +179,7 @@ export let sortArray = <MessageIds extends string>(
               element.type === 'Literal'
                 ? `${element.value}`
                 : sourceCode.text.slice(...element.range),
-            size: rangeToDiff(element.range),
+            size: rangeToDiff(element, sourceCode),
             node: element,
             group,
           }
--- a/rules/sort-classes.ts
+++ b/rules/sort-classes.ts
@@ -608,7 +608,7 @@ export default createEslintRule<SortClassesOptions, MESSAGE_ID>({
               node: member,
               dependencies,
               name,
-              requiresEndingSemicolonWhenInline: requiresEndingSemicolon,
+              addSafetySemicolonWhenInline: requiresEndingSemicolon,
               dependencyName: getDependencyName(
                 name,
                 modifiers.includes('static'),
--- a/rules/sort-classes.ts
+++ b/rules/sort-classes.ts
@@ -384,7 +384,7 @@ export default createEslintRule<SortClassesOptions, MESSAGE_ID>({
             } else if (member.key.type === 'Identifier') {
               ;({ name } = member.key)
             } else {
-              name = sourceCode.text.slice(...member.key.range)
+              name = sourceCode.getText(member.key)
             }
 
             let isPrivateHash =
--- a/rules/sort-classes.ts
+++ b/rules/sort-classes.ts
@@ -602,8 +602,8 @@ export default createEslintRule<SortClassesOptions, MESSAGE_ID>({
 
             let sortingNode: SortingNodeWithDependencies = {
               size: overloadSignatureGroupMember
-                ? rangeToDiff(overloadSignatureGroupMember.range)
-                : rangeToDiff(member.range),
+                ? rangeToDiff(overloadSignatureGroupMember, sourceCode)
+                : rangeToDiff(member, sourceCode),
               group: getGroup(),
               node: member,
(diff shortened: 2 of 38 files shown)


azat-io:
Awesome, thank you!

PR #345, opened 2024-10-29
feat(sort-imports): resolve aliased imports through tsconfig.json

### LINKED ISSUE #317: Feature: distinguish TypeScript non-relative imports relative to 'baseSrc' as a separate group in 'sort-imports' rule

sort-imports

### Describe the problem

In TypeScript, instead of `import`ing files relative to your current `.ts` file ('parent', 'sibling'), you can also use non-relative import style (which will be relative to a directory specified in `baseUrl` in `tsconfig.json`).

Since eslint-plugin-perfectionist project [also has `baseUrl` specified](https://github.com/azat-io/eslint-plugin-perfectionist/blob/v3.8.0/tsconfig.json#L15), then instead of `import sortVariableDeclarations from './rules/sort-variable-declarations'`, the source code might say: `import sortVariableDeclarations from 'rules/sort-variable-declarations'`

The problem is that TypeScript non-relative imports are treated as those belonging to 'external' group.

Would be great to distinguish them as a separate group. I would prefer to group 'ts-non-relative-imports' together with 'parent', 'sibling' and 'index'.

### Code example

To see in practice what I mean, do the following

1. Clone this repo on the latest version `git clone --branch v3.8.0 https://github.com/azat-io/eslint-plugin-perfectionist`
2. Navigate to `./index.ts` and convert one relative import to non-relative one: (for example, replace `from './rules/sort-variable-declarations'` with `from 'rules/sort-variable-declarations'`, notice the `./` got removed)
3. Apply eslint `--fix` on this file and you'll see that `from 'rules/sort-variable-declarations'` import will belong to a separate group. 

You can easily ensure that it gets treated as 'external' if you 

1) add a couple more different imports, for example:
```typescript
import {readFileSync} from 'node:fs' // 'builtin'
import naturalCompare from 'natural-compare-lite' // 'external'
```
2) and also configure `perfectionist/sort-imports` rule to split 'external' and 'builtin' into separate groups

If you configure 'builtin' and 'external' to belong separate groups, then `from 'rules/sort-variable-declarations'` will be grouped with `from 'natural-compare-lite'` (which is 'external' for this project)

### Additional comments

_No response_

### Validations

hugop95:
Resolves #317

## Description

Adds a `tsconfigRootDir?: string` option (can be changed if needed). When passed by the user, each time a file is linted, the nearest `tsconfig.json` file will be searched, read, and used to determine if an import is internal or not.

Will try to dynamically load (and cache) `typescript` so this should not impact non-typescript users.

The `tsconfig.json` search and cache algorithm was taken from https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/typescript-estree/src/parseSettings/getProjectConfigFiles.ts 🙏. 

##  Process

- Linting is run for a given file.
- Check if `tsconfigRootDir` is entered. If not, sort as usual.
- Ensures that `typescript` is present. If not, sort as usual.
- If it is, search and read the closest `tsconfig.json` file and get the compiler options.
- If no file is found, sort as usual. If an error occurs (config file is invalid, for example), throw an error.
- For each import, use the compiler options to determine if that import can be marked as `internal` or `internal-type` (for the `groups` option).

## Assigning the `internal` / `internal-type` / `external` / `external-type` group to an import

Reminder: currently, `internal` and `internal-type` are **only** assigned to groups that match the `internalPattern` option. Non-external imports are not automatically matched with the `internal` / `internal-type` group.

I have chosen to keep this, and add the following:
If an import is considered 'internal' according to the user's `tsconfig`, that import will be matched with the `internal` / `internal-type` group as if it was matching `internalPattern`.

Everything below also applies for `type` imports with `internal-type` and `external-type`.

For a given import:

- If `typescript` is not present, use the standard process: the import is `external` if it starts with `.` or `/`.
- Else: If `typescript.isExternalModuleNameRelative(import)`, do not assign `internal` nor `external` at this stage.
- If `tsconfigRootDir` is not entered or if no `tsconfig.json` file was found, assign `external`.
- Get typescript information about the import using `typescript.resolveModuleName`.
- If the resolution fails, assign `external`.
- If the resolution succeeds, assign `external` if `resolution.resolvedModule.isExternalLibraryImport === true`, otherwise `internal`.

## Performance

Tests on a ~10k files codebase, 1 `tsconfig.json`.

#### Without caching `resolveModuleName` attempts

image

#### With caching `resolveModuleName` attempts

image

#### Without `tsconfigRootDir` option (what we have today)

(Some other rules were disabled)

image

#### Bottlenecks

Tests made while caching `resolveModuleName` attempts (~950ms)

- Without searching `tsconfig.json`: ~900ms
- Without `typescript.resolveModuleName`: ~620ms

#### Conclusion

Some significant performance hit despite caching.

- Searching for the appropriate `tsconfig.json`: ~+15%
- `typescript.resolveModuleName`: ~+75%
- Reading and parsing the config: ~300ms? 

- [x] New Feature

### DIFF
--- a/docs/content/rules/sort-imports.mdx
+++ b/docs/content/rules/sort-imports.mdx
@@ -201,7 +201,7 @@ Specifies a maximum line length for sorting imports. When the line length exceed
 
 This option is only available when the type is set to `'line-length'`.
 
-### tsConfigRootDir
+### tsconfigRootDir
 
 <sub>default: `undefined`</sub>
 
--- a/docs/content/rules/sort-imports.mdx
+++ b/docs/content/rules/sort-imports.mdx
@@ -201,6 +201,12 @@ Specifies a maximum line length for sorting imports. When the line length exceed
 
 This option is only available when the type is set to `'line-length'`.
 
+### tsConfigRootDir
+
+<sub>default: `undefined`</sub>
+
+Specifies the  directory of the root `tsconfig.json` file (ex: `.`). This is used for marking aliased imports as `internal` or `internal-type` in the `groups` option.
+
 ### groups
 
 <sub>
--- a/rules/sort-imports.ts
+++ b/rules/sort-imports.ts
@@ -13,7 +13,9 @@ import {
   typeJsonSchema,
 } from '../utils/common-json-schemas'
 import { validateGroupsConfiguration } from '../utils/validate-groups-configuration'
+import { readClosestTsConfigByPath } from '../utils/read-closest-ts-config-by-path'
 import { getOptionsWithCleanGroups } from '../utils/get-options-with-clean-groups'
+import { getTypescriptImport } from '../utils/get-typescript-import'
 import { sortNodesByGroups } from '../utils/sort-nodes-by-groups'
 import { getCommentsBefore } from '../utils/get-comments-before'
 import { makeNewlinesFixes } from '../utils/make-newlines-fixes'
@@ -70,6 +72,7 @@ export type Options<T extends string[]> = [
     environment: 'node' | 'bun'
     internalPattern: string[]
     sortSideEffects: boolean
+    tsConfigRootDir?: string
     maxLineLength?: number
     order: 'desc' | 'asc'
     ignoreCase: boolean
@@ -121,6 +124,10 @@ export default createEslintRule<Options<string[]>, MESSAGE_ID>({
             type: 'integer',
             minimum: 0,
             exclusiveMinimum: true,
+          },
+          tsConfigRootDir: {
+            description: 'Specifies the tsConfig root directory.',
+            type: 'string',
           },
           groups: groupsJsonSchema,
           customGroups: {
@@ -198,6 +205,7 @@ export default createEslintRule<Options<string[]>, MESSAGE_ID>({
       sortSideEffects: false,
       newlinesBetween: 'always',
       maxLineLength: undefined,
+      tsConfigRootDir: undefined,
       groups: [
         'type',
         ['builtin', 'external'],
@@ -269,6 +277,13 @@ export default createEslintRule<Options<string[]>, MESSAGE_ID>({
         ...Object.keys(options.customGroups.value ?? {}),
       ],
     )
+
+    let compilerOptions = options.tsConfigRootDir
+      ? readClosestTsConfigByPath({
+          filePath: context.physicalFilename,
+          tsConfigRootDir: options.tsConfigRootDir,
+        })
+      : null
 
     let isSideEffectOnlyGroup = (
       group: undefined | string[] | string,
@@ -364,7 +379,7 @@ export default createEslintRule<Options<string[]>, MESSAGE_ID>({
 
       let { getGroup, defineGroup, setCustomGroups } = useGroups(options)
 
-      let isInternal = (value: string) =>
+      let matchesInternalPattern = (value: string) =>
         options.internalPattern.length &&
         options.internalPattern.some(pattern => matches(value, pattern))
 
@@ -390,12 +405,47 @@ export default createEslintRule<Options<string[]>, MESSAGE_ID>({
         )
       }
 
-      let isExternal = (value: string) =>
-        !(value.startsWith('.') || value.startsWith('/'))
+      let getInternalOrExternalGroup = (
+        value: string,
+      ): 'internal' | 'external' | null => {
+        let typescriptImport = getTypescriptImport()
+        if (!typescriptImport) {
+          return !value.startsWith('.') && !value.startsWith('/')
+            ? 'external'
+            : null
+        }
+
+        let isRelativeImport =
+          typescriptImport.isExternalModuleNameRelative(value)
+        if (isRelativeImport) {
+          return null
+        }
+        if (!compilerOptions) {
+          return 'external'
+        }
+
+        let resolution = typescriptImport.resolveModuleName(
+          value,
+          context.filename,
+          compilerOptions,
+          typescriptImport.sys,
+        )
+        // If the module can't be resolved, assume it is external
+        if (resolution.resolvedModule?.isExternalLibraryImport === undefined) {
+          return 'external'
+        }
+
+        return resolution.resolvedModule.isExternalLibraryImport
+          ? 'external'
+          : 'internal'
+      }
 
       if (node.type !== 'VariableDeclaration' && node.importKind === 'type') {
         if (node.type === 'ImportDeclaration') {
           setCustomGroups(options.customGroups.type, node.source.value)
+          let internalExternalGroup = matchesInternalPattern(node.source.value)
+            ? 'internal'
+            : getInternalOrExternalGroup(node.source.value)
 
           if (isIndex(node.source.value)) {
             defineGroup('index-type')
@@ -409,7 +459,7 @@ export default createEslintRule<Options<string[]>, MESSAGE_ID>({
             defineGroup('parent-type')
           }
 
-          if (isInternal(node.source.value)) {
+          if (internalExternalGroup === 'internal') {
             defineGroup('internal-type')
           }
 
@@ -417,7 +467,7 @@ export default createEslintRule<Options<string[]>, MESSAGE_ID>({
             defineGroup('builtin-type')
           }
 
-          if (isExternal(node.source.value)) {
+          if (internalExternalGroup === 'external') {
             defineGroup('external-type')
           }
         }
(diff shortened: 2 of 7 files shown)


rules/sort-imports.ts
```diff
@@ -1,5 +1,10 @@
 import type { TSESTree } from '@typescript-eslint/types'
 
+import {
+  isExternalModuleNameRelative,
+  resolveModuleName,
+  sys,
+} from 'typescript'
```
azat-io, line 7: Wouldn't this break the use of the plugin in JavaScript projects?

hugop95: @azat-io True! I can make the imports dynamic so the tsconfig check isn't done in non-typescript projects. What do you think?

OlivierZal: I don't see any other way...

hugop95: It's actually a bit more annoying than I thought: dynamic imports are async, and there can be no async operation within the ESLint ecosystem.

OlivierZal: Using `createRequire`?

hugop95: @OlivierZal I wasn't using `createRequire`! I was going for the `await import` route, but this seems way better, thanks!

docs/content/rules/sort-imports.mdx
```diff
@@ -201,6 +201,12 @@ Specifies a maximum line length for sorting imports. When the line length exceed
 
 This option is only available when the type is set to `'line-length'`.
 
+### tsConfigRootDir
+
+<sub>default: `undefined`</sub>
```
OlivierZal, line 206: what about defaulting to something else, like:
- `languageOptions.parserOptions.tsconfigRootDir` (<=> `undefined` if not defined)
- `import.meta.dirname` (with the possibility to override with `null` if someone isn't interested in such an import sorting)

see https://typescript-eslint.io/getting-started/typed-linting/

OlivierZal, line 206: @hugop95, what do you think about defaulting to something else, like e.g.:
- `languageOptions.parserOptions.tsconfigRootDir` (<=> `undefined` if not defined)
- `import.meta.dirname` (with the possibility to override with `null` if someone isn't interested in such an import sorting)

see https://typescript-eslint.io/getting-started/typed-linting/

docs/content/rules/sort-imports.mdx
```diff
@@ -201,6 +201,12 @@ Specifies a maximum line length for sorting imports. When the line length exceed
 
 This option is only available when the type is set to `'line-length'`.
 
+### tsConfigRootDir
```
OlivierZal, line 204: `tsconfigRootDir` (with a small `c`) would be more consistent with the already existing eslint config option: https://typescript-eslint.io/getting-started/typed-linting/

hugop95: @OlivierZal Changed in e81054b79bd07ce2977806ddd39f0590592fbf9d.

docs/content/rules/sort-imports.mdx
hugop95: @OlivierZal The option I set is a bit of a placeholder at the moment, but those are great suggestions.

hugop95: On the other hand, I don't think we should activate the feature by default (and try importing typescript + searching for a `tsconfig.json` file). So I'm not fully sure about option 2. 🧐

utils/read-closest-ts-config-by-path.ts
```diff
@@ -0,0 +1,86 @@
...
+  )
+  if (configFileRead.error) {
+    throw new Error(
+      'Error reading tsconfig file: ' + JSON.stringify(configFileRead.error),
+    )
+  }
+  let parsedContent = typescriptImport.parseJsonConfigFileContent(
+    configFileRead,
+    typescriptImport.sys,
+    './',
```
hugop95, line 71: Not fully sure if this is what should be entered here.

image

utils/read-closest-ts-config-by-path.ts
```diff
@@ -0,0 +1,86 @@
...
+  )
+  let compilerOptionsConverted =
+    typescriptImport.convertCompilerOptionsFromJson(
+      parsedContent.raw.config.compilerOptions,
+      './',
```
hugop95, line 76: [note] Same here: not fully sure if this is what needs to be entered (`./`)

utils/read-closest-ts-config-by-path.ts
OlivierZal: It's rather something like `path.dirname(filePath)`

hugop95: @OlivierZal That makes sense, updated: path.dirname(filePath)

utils/read-closest-ts-config-by-path.ts
hugop95: Updated to `path.dirname(filePath)`

hugop95:
I've run some performance tests, you can see the results in the PR description.

Long story short, there is going to be a performance hit:
- Searching for the appropriate `tsconfig.json` file is fast (<100ms in my test), but note that I tested it in a large project with a single file. Multiple `tsconfig.json` shouldn't affect the performance at this stage too much, though.
-  Reading/parsing `tsconfig.json`. I haven't tested that single stage too far, but I wouldn't be surprised if this takes 200+ms for large projects, or even more.
- `typescript.resolveModuleName`: the function used to know whether an import is external or not according to typescript. Even with cache, this takes a non-negligible amount of time, probably ~300ms in my tests.

Without the option activated, the total time for everything was about 350ms.

hugop95:
I would actually love to be able to get feedback on whether this is going to be an issue or not.

@gavvvr Would you be okay for testing this in your project(s)? This will also be a great opportunity to make sure that the new algorithm works as expected in practice.

docs/content/rules/sort-imports.mdx
OlivierZal: EDIT: we probably should disable by default to avoid any performance issue.

gavvvr:
> @gavvvr Would you be okay for testing this in your project(s)? This will also be a great opportunity to make sure that the new algorithm works as expected in practice.

Hello @hugop95. I don't have much code to test this. There are only several absolute imports in [fd8e9a2ec](https://github.com/gavvvr/obsidian-imgur-plugin/tree/fd8e9a2ec4858606d9e240f4d4401514711e2ffc) snapshot of my open-source project. Right after this commit I switched to relative imports, because I wanted sorting/grouping by perfectionist plugin to work as expected :)

Those absolute imports appeared in my codebase only because I used vscode's hotkey to add missing imports. If you are still interested, that probably will be faster for you to test it (now you know where the codebase is). Unfortunately, I'm not much experienced with testing things, which are not yet on npm registry.

hugop95:
@gavvvr No worries, thanks!

hugop95:
@azat-io @OlivierZal  
I have finished my work on this PR, performance concerns are addressed at the bottom of the PR description.

Overall, I have mixed feelings on this:

The good points:
- Code specific to handling `typescript` import, searching for the `tsconfig.json` files, etc... are isolated from the rule, so not that much code added in the rule itself.
- Performance impact is **ok**. It does make the rule take a longer time, but it's still acceptable.   

The bad:
- It's still a lot of code for one option in one rule.
- I haven't been able to test this in a large repository with multiple configs (real use-case), so even if we have tests, they don't replicate complex use cases and we might still have bugs.
-  I don't know if it will be used: will it be worth maintaining in the end?

Still to discuss:
- Are we keeping `tsconfigRootDir ` as the option's name (https://github.com/azat-io/eslint-plugin-perfectionist/pull/345#discussion_r1827315159)? What about the default value?

OlivierZal (approved):

OlivierZal:
Current implementation looks good to me:
- `tsconfigRootDir` name is consistent with existing eslint options (we could eventually also authorize the value `true`, like this is sometimes the case with some eslint options – e.g. `projectService` –, which would read `languageOptions.parserOptions.tsconfigRootDir`)
- not enabled by default, so no impact on performances if the user doesn't explicitly fill it

azat-io:
Looks good. 👍 
But I'd like to take some more time to test. Because the task is difficult.

hugop95:
@azat-io I aggree, no rush!

azat-io:
It works!

Perhaps we should also consider support for aliases from `paths` in `tsconfig.json` in the future. Thank you!

Issue #338, opened 2024-10-25
Bug: TypeScript error during setup


BenJackGill:
### Describe the bug

I am using Eslint with a TypeScript config file (`eslint.config.ts`) as described [here](https://eslint.org/docs/latest/use/configure/configuration-files#typescript-configuration-files) in the Eslint docs.

But it is producing TypeSciprt errors during setup.

### Code example

**ERROR EXAMPLE 1** - Using the plugins setup method as described [here](https://perfectionist.dev/guide/getting-started) in the perfectionist docs:

The `eslint.config.ts` file:

```TypeScript
import js from "@eslint/js";
import type { Linter } from "eslint";
import perfectionist from "eslint-plugin-perfectionist";

export default [
  js.configs.recommended,
  {
    plugins: {
      perfectionist, // <-- ERROR SHOWS HERE
    },
  },
] satisfies Linter.Config[];
```

The TypeScript error:

```TypeScript
Type '{ configs: { 'recommended-alphabetical-legacy': Config; 'recommended-line-length-legacy': Config; 'recommended-natural-legacy': Config; 'recommended-alphabetical': Config; 'recommended-line-length': Config; 'recommended-natural': Config; }; rules: { ...; }; name: string; }' is not assignable to type 'Plugin'.
  Types of property 'configs' are incompatible.
    Type '{ 'recommended-alphabetical-legacy': Config; 'recommended-line-length-legacy': Config; 'recommended-natural-legacy': Config; 'recommended-alphabetical': Config; 'recommended-line-length': Config; 'recommended-natural': Config; }' is not assignable to type 'Record<string, Config<RulesRecord> | Config<RulesRecord>[] | LegacyConfig<RulesRecord, RulesRecord>>'.
      Property ''recommended-alphabetical-legacy'' is incompatible with index signature.
        Type 'Config' is not assignable to type 'Config<RulesRecord> | Config<RulesRecord>[] | LegacyConfig<RulesRecord, RulesRecord>'.
          Type 'Config' is not assignable to type 'LegacyConfig<RulesRecord, RulesRecord>'.
            Types of property 'overrides' are incompatible.
              Type 'ConfigOverride[] | undefined' is not assignable to type 'ConfigOverride<RulesRecord>[] | undefined'.
                Type 'ConfigOverride[]' is not assignable to type 'ConfigOverride<RulesRecord>[]'.
[...]

PR #346, opened 2024-10-29
fix: make types compatible with eslint types

azat-io:
### Description

Currently there are problems with types, because TypeScript ESLint types are not exactly identical to ESLint types. This pull request fixes that, but it does it in a slightly non-natural way.

```ts
// eslint.config.ts
import type { Linter } from "eslint";

import perfectionist from "eslint-plugin-perfectionist";

export default [
  perfectionist.configs['recommended-line-length'],
] satisfies Linter.Config[];
```

### Additional context

#338

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1,7 +1,4 @@
-import type {
-  ClassicConfig,
-  FlatConfig,
-} from '@typescript-eslint/utils/ts-eslint'
+import type { ESLint, Linter, Rule } from 'eslint'
 
 import sortVariableDeclarations from './rules/sort-variable-declarations'
 import sortIntersectionTypes from './rules/sort-intersection-types'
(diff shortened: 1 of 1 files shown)


hugop95 (approved):
Nice! 

cc @OlivierZal 

For information: https://github.com/typescript-eslint/typescript-eslint/issues/9724#issuecomment-2270409598

`typescript-eslint` says:
> - our intention is to accept `@types/eslint types`
> - our intention is not to be assignable to `@types/eslint types`

but recommends using their types internally (https://github.com/typescript-eslint/typescript-eslint/issues/8613#issuecomment-1984076953), so casting our types to satisfy ESLint's types seems like the right approach 👍

azat-io:
Thank you for the review!

PR #347, opened 2024-10-30
fix(sort-switch-case): fixes block case statements ignoring break/return

### LINKED ISSUE #348: Bug(sort-switch-case): Empty last case is sorted when it should not
### Describe the bug

See code example

### Code example

```ts
function funct(x) {
    switch(x) {
        case "b":
            // Do something
            break;
        case "a":
    }
}
```

This will result in

```ts
function funct(x) {
    switch(x) {
        case "a":
        case "b":
            // Do something
            break;
    }
}
```

### ESLint version

8.57.0

3.9.1

### Additional comments

_No response_

### Validations

hugop95:
### Description

Fixes #348

Despite being wrapped in `{ }`, block case statements need `break`.

### Changes

- Fixes `break` and `return` not being considered necessary for block case statements. This means that block statements without `break/return` might still get sorted when they should not. 1 test duplicated 3 times had to be updated (commit 1: 15e07e3ef82dfe43425f1d713e7cf5370a05861b)
- Fixes an issue with the new feature, allowing sorting nodes that don't have break/return: if the **last** node has no `break/return`: the whole cases group associated must stay in place. 
- Fixes #348.

### Checklist

- [x] Update documentation for #348
- [x] Add tests 

#### Footer

- [x] Bug fix
- [x] Documentation update

### DIFF
--- a/docs/content/rules/sort-switch-case.mdx
+++ b/docs/content/rules/sort-switch-case.mdx
@@ -27,6 +27,10 @@ This practice contributes to a more readable and maintainable codebase, allowing
 
 By integrating this rule into your ESLint configuration, you can focus on the functionality of your code, confident that your switch statements are consistently structured and easy to manage.
 
+<Important>
+It is highly recommended to disable the [`no-useless-return`](https://eslint.org/docs/latest/rules/no-useless-return) rule to avoid conflicts (see [this issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues/348)).
+</Important>
+
 ## Try it out
 
 <CodeExample
--- a/docs/content/rules/sort-switch-case.mdx
+++ b/docs/content/rules/sort-switch-case.mdx
@@ -27,10 +27,6 @@ This practice contributes to a more readable and maintainable codebase, allowing
 
 By integrating this rule into your ESLint configuration, you can focus on the functionality of your code, confident that your switch statements are consistently structured and easy to manage.
 
-<Important>
-It is highly recommended to disable the [`no-useless-return`](https://eslint.org/docs/latest/rules/no-useless-return) rule to avoid conflicts (see [this issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues/348)).
-</Important>
-
 ## Try it out
 
 <CodeExample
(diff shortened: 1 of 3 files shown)


docs/content/rules/sort-switch-case.mdx
```diff
@@ -27,6 +27,10 @@ This practice contributes to a more readable and maintainable codebase, allowing
 
 By integrating this rule into your ESLint configuration, you can focus on the functionality of your code, confident that your switch statements are consistently structured and easy to manage.
 
+<Important>
+It is highly recommended to disable the [`no-useless-return`](https://eslint.org/docs/latest/rules/no-useless-return) rule to avoid conflicts (see [this issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues/348)).
```
azat-io, line 31: Thank you. I need time to reproduce locally the problem with `no-useless-return` and `perfectionist/sort-switch-case`.

hugop95: @azat-io No worries, here is a simple configuration if that can save you some time (legacy config with `"eslint": "^8.57.0"`):

.eslintrc.json
```json
{
  "extends": [
    "eslint:all",
    "plugin:@typescript-eslint/strict",
    "plugin:perfectionist/recommended-alphabetical-legacy"
  ]
}
```

Code
```ts
function funct(x) {
    switch(x) {
        case "b":
            break;
        case "a":
            return;
    }
}
```

azat-io: That's an interesting observation. Thank you, the problem is indeed a curious one.

However, it seems to me that this case is very rare, and the `no-useless-return' rule is quite popular. I use it myself in my work. Also this rule is included in popular ESLint configs:

- `eslint-config-standard`
- `eslint-config-airbnb`
- `eslint-config-eslint`
- etc.

Maybe we should remove this warning? Because the case is really rare.

hugop95: @azat-io Good news, I have run additional tests on this version (next) and can't reproduce it, it seems that the previous PR fixed it. I reverted the documentation commit.

hugop95: Alright, I understand why: the issue actually comes from the fact that

```ts
function funct(x) {
    switch(x) {
        case "b":
            let a;
            break;
        case "a":
    }
}
```

will get sorted to

```ts
function funct(x) {
    switch(x) {
        case "b":
            let a;
            break;
        case "a":
    }
}
```

`no-useless-return` doesn't actually affect us! This PR fixes it, I'll add a test.

azat-io (approved):

PR #349, opened 2024-10-31
feat!: remove minimatch in favor of regexp

azat-io:
### Description

Almost all ESLint uses regular expressions. I think this will be a good change to keep ESLint configs consistent.

### Additional context

We can discuss about it in #341.

---

- [ ] Bug fix
- [x] New Feature
- [x] Documentation update
- [ ] Other


docs/content/rules/sort-interfaces.mdx
```diff
@@ -177,7 +177,7 @@ Controls whether special characters should be trimmed, removed or kept before so
 
 Allows you to specify names or patterns for interfaces that should be ignored by this rule. This can be useful if you have specific interfaces that you do not want to sort.
 
-You can specify their names or a glob pattern to ignore, for example: `'Component*'` to ignore all interfaces whose names begin with the word “Component”.
+You can specify their names or a regexp pattern to ignore, for example: `'Component*'` to ignore all interfaces whose names begin with the word “Component”.
```
OlivierZal, line 180: @azat-io, `'Component*'` example in this sentence is outdated

docs/content/rules/sort-jsx-props.mdx
```diff
@@ -178,7 +178,7 @@ Controls whether special characters should be trimmed, removed or kept before so
 
 Allows you to specify names or patterns for JSX elements that should be ignored by this rule. This can be useful if you have specific components that you do not want to sort.
 
-You can specify their names or a glob pattern to ignore, for example: `'Table*'` to ignore all object types whose names begin with the word Table.
+You can specify their names or a regexp pattern to ignore, for example: `'Table*'` to ignore all object types whose names begin with the word Table.
```
OlivierZal, line 181: same than [here](https://github.com/azat-io/eslint-plugin-perfectionist/pull/349/files#r1825255558)

docs/content/rules/sort-objects.mdx
```diff
@@ -253,7 +253,7 @@ Determines whether this rule should be applied to styled-components like librari
 
 Allows you to specify names or patterns for object types that should be ignored by this rule. This can be useful if you have specific objects that you do not want to sort.
 
-You can specify their names or a glob pattern to ignore, for example: `'User*'` to ignore all object types whose names begin with the word “User”.
+You can specify their names or a regexp pattern to ignore, for example: `'User*'` to ignore all object types whose names begin with the word “User”.
```
OlivierZal, line 256: same than [here](https://github.com/azat-io/eslint-plugin-perfectionist/pull/349/files#r1825255558)

azat-io:
@OlivierZal Thank you for the review. All of your comments have been fixed. What do you think of the idea itself?

OlivierZal:
I totally agree with you: having only one matcher simplifies, and `regex` provides more flexibility (`minimatch` is good to match files).

hugop95 (approved):
@azat-io LGTM! I think that the power of `minimatch` came into play for `sort-imports`. I personally wasn't using that feature in this rule, but this is one case where I think that `minimatch` is generally easier to use than regex. But all `minimatch` expressions can be converted to a regex one, so we theoretically lose no functionality here.

azat-io:
@hugop95 Thank you! I would like it to work the same way in all rules.

Issue #350, opened 2024-11-02
Feature: How to sort by function in sort-classes rule?


jon9090:
### What rule do you want to change?

perfectionist/sort-classes

### Describe the problem

I have this class. Is it possible to order the members by function type, in this sequence: `inject`, `input`, `output`, variables, `rxState`, `injectQuery`, `injectMutation`, and then the rest?

Thanks for the help!

### Code example

```typescript
class Foo {
   api = inject(ApiService);
   foo = input(..);
   fooChanged = output(..);

   disabledStatus = true;

   state = rxState(() => {...});

   userQuery = injectQuery(() => { ... });
   clientQuery = injectQuery(() => { ... });
      
   userMutation = injectMutation(() => { ... });
   clientMutation = injectMutation(() => { ... });

   open() {....} // other functions here
}
```

### Validations

- [X] Read the [docs](https://perfectionist.dev).
- [X] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

jon9090:
@hugop95  Thank you for the quick reply!

jon9090:
Hi @hugop95, I checked this, and something isn't working as expected.

I made an example where you can run `npm run lint` to see that there’s no change in `index.ts`. These are the options I used, as you suggested, but the sorting isn’t coming out correctly. 

For example, lint is producing this order:
```typescript
#usersApi = inject(UsersApiService);
breadcrumbs = injectBreadcrumbs();
metadata = injectQuery(() => ({
  queryKey: ['reports', 'metadata'] as const,
  queryFn: () => this.#reportApi.get(),
}));
querystring = injectQueryString({});
state = rxState<{
```

But it should be:
```typescript
#usersApi = inject(UsersApiService);
breadcrumbs = injectBreadcrumbs();
querystring = injectQueryString({});
state = rxState<{
metadata = injectQuery(() => ({
  queryKey: ['reports', 'metadata'] as const,
  queryFn: () => this.#reportApi.get(),
}));
```

It seems the sorting is not respecting the order set in the `groups` configuration:
```javascript
groups: [
  'inject',
  'input',
  'output',
  'rxState',
  'injectQuery',
  'injectMutation',
  'property',
  'unknown',
]
```

Here's a StackBlitz example for reference: https://stackblitz.com/edit/stackblitz-starters-duzwun

the fullcode:

lint:
```
 

[...]

PR #354, opened 2024-11-04
feat(sort-classes): adds `partitionByNewLine` and `newlinesBetween` options

hugop95:
The majority of lines added are tests and documentation.

### Description

Simply adds those two options to `sort-classes`.

### Changes

- Adds `partitionByNewLine` option. `partitionByComment` already exists and probably is a better use case, but supporting `partitionByNewline` is a matter of a couple of lines. Might as well add it!
- Adds `newlinesBetween` option (requested here: https://github.com/azat-io/eslint-plugin-perfectionist/issues/350#issuecomment-2453047795)
- Creates a `partitionByNewLineJsonSchema` JSON schema to reduce code duplication.

As with other rules using this option, `partitionByNewLine` and `newlinesBetween` can not be used together.

- [x] New Feature

### DIFF
--- a/docs/content/rules/sort-classes.mdx
+++ b/docs/content/rules/sort-classes.mdx
@@ -213,6 +213,39 @@ Allows you to use comments to separate the class members into logical groups. Th
 -	`false` — Comments will not be used as delimiters.
 - string — A regexp pattern to specify which comments should act as delimiters.
 
+### partitionByNewLine
+
+<sub>default: `false`</sub>
+
+When `true`, the rule will not sort the members of a class if there is an empty line between them. This can be useful for keeping logically separated groups of members in their defined order.
+
+```ts
+class User {
+  // Group 1
+  firstName: string;
+  lastName: string;
+
+  // Group 2
+  age: number;
+  birthDate: Date;
+
+  // Group 3
+  address: {
+    street: string;
+    city: string;
+  };
+  phone?: string;
+
+  // Group 4
+  updateAddress(address: string) {}
+  updatePhone(phone?: string) {}
+
+  // Group 5
+  editFirstName(firstName: string) {}
+  editLastName(lastName: string) {}
+};
+```
+
 ### groups
 
 <sub>
--- a/docs/content/rules/sort-classes.mdx
+++ b/docs/content/rules/sort-classes.mdx
@@ -246,6 +246,18 @@ class User {
 };
 ```
 
+### newlinesBetween
+
+<sub>default: `'ignore'`</sub>
+
+Specifies how new lines should be handled between class member groups.
+
+- `ignore` — Do not report errors related to new lines between object type groups.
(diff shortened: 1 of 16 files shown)


docs/content/rules/sort-classes.mdx
```diff
@@ -213,6 +213,51 @@ Allows you to use comments to separate the class members into logical groups. Th
...
+  updateAddress(address: string) {}
+  updatePhone(phone?: string) {}
+
+  // Group 5
+  editFirstName(firstName: string) {}
+  editLastName(lastName: string) {}
+};
+```
+
+### newlinesBetween
```
OlivierZal, line 249: default is `always` for imports, shouldn't we align with it since it's common practice to separate class members with new lines?

docs/content/rules/sort-classes.mdx
```diff
@@ -213,6 +213,51 @@ Allows you to use comments to separate the class members into logical groups. Th
...
+
+<sub>default: `'ignore'`</sub>
```
OlivierZal, line 251: default is `always` for imports, shouldn't we align with it since it's a common practice to separate class members with new lines?

However we could argue about the opportunity to have such an option since `@stylistic/lines-between-class-members` already exists (unlike for imports).

OlivierZal (approved):

hugop95:
@OlivierZal I've seen your comment 👀

I'm posting my answer here because it also includes some thought about how I can see option evolve in the future:


I added this option because it's there for the majority of rules already, and it was very simple to plug it to this one as well.
The more options we can have in common in all rules the better in my opinion, even if technically, other plugins might handle something similar.

`newlinesBetween` separates groups defined in the `groups` option, which is innate to `perfectionist`.

> default is always for imports, shouldn't we align with it since it's a common practice to separate class members with new lines?

I don't think we should for the following reasons:
- The number of newlines will heavily depend on the user's `groups` (which, in `sort-classes`, can be very large). Some users might want to separate `public`/`protected`/`private` properties, but not include newlines between them, so I think that users choosing to activate this option will probably want to modify their `groups` configuration as well to have the right amount of group separation and newlines.
- Although this will be included in a new major version, I'm not sure if the majority of users would want to have this enabled by default.
- All other rules handling `newlinesBetween`, other than `sort-imports`, have it set as `ignore`.
 
In the context of `sort-classes`, I see this option as a potential pre-work for something better: the possibility to configure newlines at a deeper level. For example:
- Requiring newlines between members of specific groups.
- Requiring newlines above specific groups, and not for others.

The question is to know whether this is something that users really want or not, and I think that a "basic" option like `newlinesBetween` today can serve as a way to measure interest in that.

OlivierZal:
Exactly, totally agree that's why I removed my comment. I also read too fast and made a confusion with the `@stylistic/lines-between-class-members`, which doesn't target the same "lines".

azat-io (approved):

PR #358, opened 2024-11-06
feat: adds ESLint disable directives handline

hugop95:
## Description

ESLint offers multiple ways to [disable rules](https://eslint.org/docs/latest/use/configure/rules#disabling-rules).

When a line is disabled for a rule, ESLint suppresses the error that would appear on that line for that rule. This works well for rules that fix a single error, such as adding a missing semicolon.

However, `perfectionist` does not operate on single elements: an error is necessarily associated to **2** elements: the 2 nodes that should switch place together. While we display a single error on node `A` in that case (`A` should be before `B`), there technically is a second one on node `B` (`B` should be before `A`).

In practice, this means that we never fix a single error at a time: we "fix" all errors at once (which makes sense in our context).

### The problem

The problem is that adding a `// eslint-disable` comment to suppress an error will not prevent other errors in that file from fixing the error that was associated with disabled comment.

This means that users have no way to put a node in its place for good while still being able to sort other nodes.

## Proposal

This PR implements the following:
- Any node that is affected by `eslint-disable`-like comments will be considered `disabled`:
  - ESLint takes care of suppressing the error.
  - We make sure that this node does not enter the sorting algorithm.

### Technical note

We are not suppressing the error on our own! This is important, because ESLint may report a warning or an error if an `eslint-disable`-like comment is placed but no error was expected, so it is **crucial** to make sure that our rules still report errors (that ESLint will suppress) while we ignore the node internally.

### Dependencies

Some rules require some nodes to be placed before other nodes due to dependencies. The user is free to `eslint-disable` nodes to bypass this, at their own risk.

## Checklist

- [ ] sort-array-includes
- [ ] sort-classes
- [ ] sort-decorators
- [ ] sort-enums
- [ ] sort-exports
- [ ] sort-heritage-clauses
- [ ] sort-imports
- [ ] sort-interfaces
- [ ] sort-intersection-types
- [ ] sort-jsx-props
- [ ] sort-maps
- [ ] sort-named-exports
- [ ] sort-named-imports
- [ ] sort-object-types
- [ ] sort-objects
- [ ] sort-sets
- [ ] sort-switch-case
- [ ] sort-union-types
- [ ] sort-variable-declarations

- [x] New Feature


hugop95:
Actually, waiting for #361 to be merged first.

hugop95:
I have finished handling of `/* eslint-disable */` as well. Need to add some tests for that. Will be done today.

rules/sort-classes.ts
```diff
@@ -618,30 +626,47 @@ export default createEslintRule<SortClassesOptions, MESSAGE_ID>({
...
-            isNodeIgnored: sortingNode =>
-              getGroupNumber(options.groups, sortingNode) ===
-              options.groups.length,
-          }),
+
+      let sortNodesIgnoringEslintDisabledNodes = (
+        ignoreEslintDisabledNodes: boolean,
+      ) =>
+        sortNodesByDependencies(
+          formattedNodes.flatMap(nodes =>
```
OlivierZal, line 634: You used `flatMap` 😉

OlivierZal (approved):

azat-io (approved):

Issue #359, opened 2024-11-06
Feature: Sorting modules


mnaoumov:
### Describe the rule

It should sort members of the modules similarly to the [sort-classes](https://perfectionist.dev/rules/sort-classes) rule

### Code example

See https://stackblitz.com/edit/stackblitz-starters-p1d3rv?file=index.ts which has [recommended-alphabetical](https://perfectionist.dev/configs/recommended-alphabetical) config enabled

I expect

```ts
export function c() {
  a();
}
export function b() {
  a();
}
function a() {}
```

after executing this rule to become

```ts
export function b() {
  a();
}
export function c() {
  a();
}
function a() {}
```

as I think exported functions (like `public` in classes) should be higher than normal functions (like `private` in classes)

### Additional comments

I am surprised I couldn't find any related feature requests. This seems like a must-have rule

### Validations

- [X] Read the [docs](https://perfectionist.dev).
- [X] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

azat-io:
Duplicate #53.

It's an interesting idea, but we don't seem to have a way to do it safely.

hugop95:
I have started a prototype of such rule a while back, there are issues that need to be discussed:

## Type of elements sorted

My prototype is called `sort-files`: it doesn't just sort exports, it sorts all elements that you may find in a file: `namespace`, `interface`, `type`, `module`, `class`, `type`, `enum` and `function` (note that for `class`, `enum`, `interface` and `type`, it doesn't sort their content, as we have rules for them, but their declaration position).

## Safety concerns

Safety concerns: as @azat-io says, we need to ensure that compilation succeeds and runtime behavior is left unchanged. We manage this pretty well in `sort-classes` now. We've had a number of issues, but slowly got rid of all the possible cases. 

I think that eventually we can reach that with `sort-files`, but it may take some time and we may encounter unexpected difficulties.
Classes, enums and variables order matter. And you need to consider that you may have to do the sorting at multiple levels, such as within a `namespace`.

## Overlap with other rules

[...]

Issue #261, opened 2024-09-07
Feature: order function


splincode:
### Describe the rule

before

```ts
export function b() {
 // ..
}

export function c() {
 // ..
}

export function a() {
 // ..
}
```

### Code example

after

```ts
export function a() {
 // ..
}

export function b() {
 // ..
}

export function c() {
 // ..
}
```

### Validations

- [X] Read the [docs](https://perfectionist.dev).
- [X] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

hugop95:
I like the idea but think that it be extended to choosing how to sort file members in general rather than just functions (types, interfaces, classes etc).

OlivierZal:
I agree with @hugop95 it could be with anything: constants, arrow functions, etc.

The difficulty is to pay attention whether they are dependent one with the others, to avoid calling "something" before its declaration.

azat-io:
It seems to me that this would be extremely difficult to implement.

Issue #53, opened 2023-08-04
Feature: (sort-exports) Support sorting exports of `const` / `function` / `interface` / `type` and other similar exports.


u3u:
### What rule do you want to change?

sort-exports

### Describe the problem

Same as https://github.com/jrdrg/eslint-plugin-sort-exports

I think [`eslint-plugin-sort-exports`](https://github.com/jrdrg/eslint-plugin-sort-exports) is great, and I have been using it all along. However, it has a fatal flaw: if the exported variables depend on another variable, sorting them may result in accessing undefined variables. Currently, I can only disable it using `// eslint-disable-next-line sort-exports/sort-exports` (even if referenced within an exported function or an arrow function defined with `export const`, there won't be any issues because of variable hoisting). It would be great if it could be optimized and fixed based on this!

### Code example

**Input**

```ts
export const z = { x: 0 };

// line comment
export const a = () => {
  return b;
};

// Due to the dependency on constant `z`, `b` should be placed after `z`.
export const b = z.x;

export enum E {};

export type B = '';

/** block comment */
export interface C {};
```

**Output**

If the `sortExportKindFirst` option is set to `type`, then the `export interface` and `export type` need to be placed at the top for separate sorting, with other export statements placed below them for separate sorting. I think this should be the default value because it is more in line with the intuitive definition of React components.

> **Note**: The comment above the `export` statement should be sorted together with it.

```ts
export type B = '';

/** block comment */
export interface C {}

export enum E {}

// line comment
export const a = () => {
  return b;
};

export const z = { x: 0 };

// Due to the dependency on constant `z`, `b` should be placed after `z`.
export const b = z.x;
```

### Validations

- [X] Read the [docs](https://eslint-plugin-perfectionist.azat.io).
- [X] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

azat-io:
It's a cool idea. But I think it's hard to implement.

azat-io:
Thank you for your issue!

Unfortunately, I don't see any way to do this feature at the moment.Perhaps we'll revisit this idea in the future.

There are no plans to do so in the near future.

PR #360, opened 2024-11-06
New rule: `sort-modules`

hugop95:
I started this a month or two ago, but there are still a lot of things that need to be discussed.
See https://github.com/azat-io/eslint-plugin-perfectionist/issues/359#issuecomment-2459077550.

### Description

There have been a lot of requests to be able to sort various declaration types in modules for a while now (#53, #261 and #359 to only name a few).

It's probably going to be the most complex rule that we have, but I think it's worth giving it a full look and see how far we can go.

This rule is very similar to `sort-classes` in many ways, in terms of what it aims to do, the issues we will likely encounter, and the implementation I have in mind:
- It will sort elements which may not have the same types (interface declarations, class declarations, enum declarations, etc).
-  Dependencies have to be taken into account.
- There are plenty of tokens to take into account with the `groups` option (`declare`, `export`, etc).

All of what is presented below is a proposal, nothing is fixed. Feel free to suggest anything.

## What should it sort

- `interface` declarations.
- `type` declarations.
- `class` declarations.
- `function` declarations.
- `enum` declarations.
- ~~`namespace` declarations~~ (complex, see below `Dependencies`).
- ~~`module` declarations~~ (complex, see below `Dependencies`).

This rule should sort elements within a `module` or `namespace` as well.

### What it can not sort

- variables (`const`, `let` ...), as this deeply changes runtime behavior, except function declarations. This is complex and can be studied later on if needed.
- `import` statements: handled by `sort-imports`.
- "From" exports (`export * from ...` or `export { a } from ...`): handled by `sort-exports`.

## Proposal

Follow an approach identical to `sort-classes` with `selectors` and `modifiers` to generate groups.

### Selectors and modifiers

- `interface`: `declare`, `default`, `export`. 
- `type`: `declare`, `export`.
- `class`: `declare`, `default`, `export`, `decorated`.
- `function`: `declare`, `default`, `export`, `async` (Could be used in `sort-classes` as well for consistency)
- `enum`: `declare`,  `export`. 
- ~~`namespace`: `declare`, `export`~~.
- ~~`module`: `declare`, `export`~~.

## Dependencies

The following elements must be declared before they are used:
- `class` elements.
- `enum` elements.
- `module` and `namespace` elements: this is where things get tricky: JavaScript/TypeScript hoists declarations within `namespace`/`module` but doesn't assign them to the namespace object until execution.
  - This means that
    ```ts
    module Test {
      test()
      export function test() {
          console.log("test")
      }
    }
    ```
    works but
    ```ts
    Test.test()
    module Test {
      export function test() {
          console.log("test")
      }
    }
    ```
    doesn't.

### Consequences

It is probably better not to sort `namespace` and `module` elements for now (their content can be sorted, but not their declarations).

## Partition elements

The following elements have to act as partition elements in order to ensure compilation and runtime safety. This means that when one of those elements is encountered, members below that element and above that element will not be sorted together.

- `namespace`/`module` declarations.
- any variable declaration or method call.

## Custom Group API

This rule uses the same `customGroups` API as in `sort-classes`, with one exception: the `elementValuePattern` option is not present.

## Default groups

```ts
  groups: [
    'declare-enum',
    'export-enum',
    'enum',
    ['declare-interface', 'declare-type'],
    ['export-interface', 'export-type'],
    ['interface', 'type'],
    'declare-class',
    'export-class',
    'class',
    'declare-function',
    'export-function',
    'function'
  ]
```

- [x] New Feature


hugop95:
Hi @OlivierZal 👋

You've greatly contributed to the `sort-classes` redesign in #196, I think you will be interested in this!

OlivierZal:
Hi @hugop95, I was already looking at your PR, but didn't react yet since it's a draft ;-)

Will have deeper look at it!

hugop95:
@OlivierZal The code is a very early WIP now, so no need to take a look at it right now, we're still in the specification phase 😁

OlivierZal:
I started to perform some tests to get more familiar with what could be expected.

hugop95:
I should have a first fully working build tomorrow.

hugop95:
I have currently fixed the default groups to:

```ts
  groups: [
    'declare-enum',
    'export-enum',
    'enum',
    ['declare-interface', 'declare-type'],
    ['export-interface', 'export-type'],
    ['interface', 'type'],
    'declare-class',
    'export-class',
    'class',
    'declare-function',
    'export-function',
    'function',
    'unknown',
  ],
```

Don't hesitate to improve it! This is a bit of a placeholder.

hugop95:
We have a first stable build now. Tested on a large project and it compiles OK. 

Don't hesitate to build and test this in one of your projects and give some feedback on what you think could be improved!

hugop95:
I've added all tests related to dependency detection. Use the latest build if possible, a few things were fixed. I'm about 50% done on all unit tests.

OlivierZal:
For now, I think everything makes sense.

rules/sort-modules.types.ts
```diff
@@ -0,0 +1,210 @@
...
+  'function',
+  'interface',
+  // 'module',
+  // 'namespace',
+  'type',
+  'class',
+]
+
+export const allModifiers: Modifier[] = [
+  'async',
```
OlivierZal, line 153: If I well understand, this will also be added to `sort-class` – what is cool.

OlivierZal, line 153: thanks for having added this

hugop95:
The first version of this rule is done (at last)!

I'm sure there are rooms for improvements, but that will come with user feedback.

The documentation is **mostly** copy-pasted from `sort-classes` (they work the same, after all). I haven't spent too much time on it, though, so it's possible that some things might need to be corrected.

I'm tempted to add a disclaimer saying that things might change a bit at the beginning: the rule is fresh, so heavy (read: breaking) changes can be expected.

docs/content/rules/sort-modules.mdx
```diff
@@ -0,0 +1,592 @@
...
+                'error',
+                {
+                  type: 'alphabetical',
+                  order: 'asc',
+                  ignoreCase: true,
+                  specialCharacters: 'keep',
+                  partitionByComment: false,
+                  partitionByNewLine: false,
+                  newlinesBetween: 'ignore',
+                  groups: [
```
OlivierZal, line 514: To open a discussion, maybe I would propose the following order:

```
'declare-enum',
'enum',
'export-enum',
['declare-interface', 'declare-type'],
['interface', 'type'],
['export-interface', 'export-type'],
'declare-class',
'class',
'export-class',
'declare-function',
'function'
'export-function',
```

because it seems more logical, e.g. to avoid dependency issues like this:
```
export interface A {}

interface B {}

export interface C {}
```

while all the exports grouped at the end would sort without care about such dependencies:

```
interface B {}

export interface A {}

export interface C {}
```

docs/content/rules/sort-modules.mdx
```diff
@@ -0,0 +1,592 @@
...
+
+- Selector: `class`.
+- Modifiers: `declare`, `default`, `decorated`, `export`.
+- Example: `declare-class`, `export-default-decorated-class` or `class`.
+
+#### Functions
+
+- Selector: `function`.
+- Modifiers: `declare`, `default`, `async`, `export`.
+- Example: `async-function`, `export-default-function` or `function`.
```
OlivierZal, line 344: That's not the purpose of this PR, but could we also add `async` to `sort-classes`?

docs/content/rules/sort-modules.mdx
hugop95: @OlivierZal `interfaces` and `types` do not suffer dependency issues like `classes` and `enums` have (they are eliminated at runtime and typescript does not require interfaces or types to be ordered for compilation).

The question is valid for `class` and `enum` though, and I think that we have two schools:
- Put `export-ed` elements at the end to ensure that no dependency issues occur (what you propose).
- Put `export-ed` elements first (they are generally above in the codebases I have seen), and rely on our dependency detection system to place `class`/`enum` above the exported element if needed.

I'm personally for the second one: I trust our dependency detection system (even though it might not be bulletproof at first, `sort-class` is now very reliable on that matter). But it's really a matter of personal preference in the end.

hugop95: Functions on the other hand don't depend on other functions (or at least, it's not required at runtime to have them ordered), so I propose to do a bit of both schools:

- Place `type`, `interface`, `class` exports at the end.
- Place `export function` first (Industry standard?)
- What about enums?

docs/content/rules/sort-modules.mdx
hugop95: @OlivierZal I have it planned! It will come in a separate PR.

docs/content/rules/sort-modules.mdx
OlivierZal: Yes, take `interface` as an example was not smart 😅 but you got my point, and I've no strong option since default can be overridden with a custom configuration according to user's preferences.

So I'm also ok with option 2, it was just to open a discussion.

hugop95: Your argument regarding interfaces/types/classes having sub-types/interfaces/classes makes a lot of sense, so I'm actually more incline toward the hybrid proposal than option 2 now!

hugop95: I'll try to search for official guidelines on this and see what comes out.

readme.md
```diff
@@ -185,6 +185,7 @@ module.exports = {
 | [sort-intersection-types](https://perfectionist.dev/rules/sort-intersection-types)       | Enforce sorted intersection types             | 🔧  |
 | [sort-jsx-props](https://perfectionist.dev/rules/sort-jsx-props)                         | Enforce sorted JSX props                      | 🔧  |
 | [sort-maps](https://perfectionist.dev/rules/sort-maps)                                   | Enforce sorted Map elements                   | 🔧  |
+| [sort-modules](https://perfectionist.dev/rules/sort-modules)                             | Enforce sorted modules and namespaces         | 🔧  |
```
OlivierZal, line 188: do you finally want to mention namespaces? maybe you should explain in the documentation how they will be handled?

hugop95: @OlivierZal I agree that it's not mentioned in the documentation as I thought it would be a bit confusing:
the rule sorts `modules` and `namespaces` but not together:

```ts
function a() {}
 function b() {} // We are sorted!


namespace B {
	function a() {}
    function b() {} // We are sorted!
}

namespace A { // namespaces are not sorted together
	function a() {}
    function b() {} // We are sorted!
}
```

So I think we could say it sorts modules and namespaces **content** instead?

OlivierZal: Looks good to me

docs/content/rules/sort-modules.mdx
hugop95: I'm wondering if we're not missing a use-case:
A generic `types.ts` file exporting

```ts
export type A = {
    attribute1: D
}
type D = "d"

export type B = {
    attribute1: C
}
type C = "c"
``` 

The sorted result is
```ts
export type A = {
    attribute1: D
}
export type B = {
    attribute1: C
}

type C = "c"
type D = "d"
```

I think that some users would prefer to have the 1st sort: it's essentially the same as if we activated the dependency system for `types` (or `interfaces`), but reversed: dependencies are put right after their `parent` rather than before.

Note: if we allow users to enable this (through some sort of `takeTypesDependenciesIntoAccount` option), it's important to also allow them to choose what types of dependencies to take into account:

```ts
export MyEnum {}

export type A = {
    attribute1: MyEnum
}
```

Users might not want to put `MyEnum` after `A`.

I was thinking of an API like this in options:

```ts
{
  typeDependencies: "ignore" | { position: "before" | "after"; types: ('type' | 'interface' | 'enum' | 'class')[]
}
```

OlivierZal: Yes, could be postponed. 

I like your suggestion (ignore, before, after), but I'm not sure to see why options should separate options for types and interfaces.

OlivierZal: I identify multiple `export` use cases (with or without `type`):

```
export = App

export { App }

export default App

export { default as App }

export { App } from './some-import.js'
```

hugop95: I've updated the default groups order to place `export-class` after `class`. It looks like placing exported interfaces and types first seem standard in the codebases I have seen though. I'm sure we are going to be improving this over time, but there are too many variables to take into account that I don't objectively know what would be the best approach to take regarding this, so I think we can wait for some user feedback on that.

hugop95:
@azat-io @OlivierZal I am done regarding this, feel free to suggest any other change, but otherwise considering this ready to be merged for me. 😁

readme.md
hugop95: Updated here f954b6d1e91009e58e96e2d3f2d74a96306b2b4c

docs/content/rules/sort-classes.mdx
```diff
@@ -766,6 +767,7 @@ Example:
                 partitionByComment: false,
                 partitionByNewLine: false,
                 newlinesBetween: 'ignore',
+                ignoreCallbackDependenciesPatterns
```
azat-io, line 770: Missed value?

azat-io:
Looks nice. I want to test this on a few big projects today.

docs/content/rules/sort-classes.mdx
hugop95: @azat-io Oops, fixed here 5247c9b3b1c0ce0df23c74f5347d6de98855d083

OlivierZal (approved):
Thanks a lot @hugop95, looks good to me too. It seems to be a very good start for this new rule.

azat-io:
I've tested this rule on several projects and it looks amazing. I would like to merge it.

azat-io (approved):

hugop95:
@OlivierZal Thanks for all the help and feedback!

PR #361, opened 2024-11-07
feat(sort-imports)!: adds `partitionByNewLine` and `partitionByComment`

hugop95:
### Description

`sort-imports` doesn't have `partitionByNewLine` and `partitionByComment` options.
However, the rule acts as if `partitionByComment: true` was set up by default: all comments will partition imports.

## Changes

- Adds `partitionByNewLine` (by default `false`) option. As with other rules, this option is incompatible with `newlinesBetween`.
- Adds `partitionByComment` (by default `false`) option.
-  [BREAKING CHANGE] Removes the current behavior that acts as if `partitionByComment: true` was set by default. The reasoning is that all other rules have `partitionByComment: false` by default.

One test duplicated 3 time was removed (see commit **1**), and replaced with other tests.

### Additional fix

- Fixes missing config field in the docs of `sort-classes`.

- [x] New Feature


OlivierZal (approved):

test/sort-imports.test.ts
```diff
@@ -623,41 +623,6 @@ describe(ruleName, () => {
...
-    ruleTester.run(`${ruleName}(${type}): doesn't break user comments`, rule, {
-      valid: [
-        {
-          code: dedent`
-            import { b1, b2 } from 'b'
-
-            /**
-             * Comment
-             */
-
```
azat-io, line 635: If there is no blank line between the comment and the import, it will consider this comment as a comment to the import and move it along with it when sorting, right?

hugop95: @azat-io Yes, you can see in 307f8596f308215fe2d5ba96338cc8d15b25a8d1, line 2150, an example where `// Not partition comment` is moved around with its import.

azat-io (approved):

PR #365, opened 2024-11-11
refactor: improve code quality

azat-io:
### Description

I plan to make a more rigorous ESLint config soon. This pull request is a preparation for that. This is where I'm doing code refactoring:
- typing of function results
- regular expressions improvement
- variable naming improvement
- code reduction
- microoptimizations in the documentation code, reducing the number of renderers

None of the tests have been changed.

### Additional context

N/A.

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other

### DIFF
--- a/test/get-eslint-disabled-rules.test.ts
+++ b/test/get-eslint-disabled-rules.test.ts
@@ -2,7 +2,7 @@ import { describe, expect, it } from 'vitest'
 
 import { getEslintDisabledRules } from '../utils/get-eslint-disabled-rules'
 
-const eslintDisableDirectives = [
+let eslintDisableDirectives = [
   'eslint-disable',
   'eslint-enable',
   'eslint-disable-line',
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -34,7 +34,7 @@ export default defineConfig({
       },
     }),
     sitemap({
-      filter: page => !new RegExp(`^${site}/guide$`).test(page),
+      filter: page => !new RegExp(`^${site}/guide$`, 'u').test(page),
     }),
     mdx(),
   ],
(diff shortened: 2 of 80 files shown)


test/get-eslint-disabled-rules.test.ts
```diff
@@ -2,7 +2,7 @@ import { describe, expect, it } from 'vitest'
 
 import { getEslintDisabledRules } from '../utils/get-eslint-disabled-rules'
 
-const eslintDisableDirectives = [
```
hugop95, line 5: [question] @azat-io I know that we use https://www.npmjs.com/package/eslint-plugin-prefer-let, but should top-level constant declarations use `let` as well?

hugop95 (approved):
🚀

test/get-eslint-disabled-rules.test.ts
azat-io: Just personal preference.

We use `eslint-plugin-prefer-let` to encourage the semantic use of `let` and `const` for clarity and readability. However, top-level constant declarations should still use `const` when they represent true, immutable constants — typically values that do not depend on other variables or computations, such as mathematical constants (`PI`, `E`) or other universally fixed values.

In our project, `const` is reserved for these true constants and is written in `UPPER_CASE` to distinguish it visually and semantically from other variables. Using `const` for these values emphasizes that they are intended to remain unchanged globally, reinforcing their immutability to readers and developers.

For other variables and bindings, even at the top level of a module, we prefer `let` to align with our approach that prioritizes code readability. This choice allows `const` to retain its specific meaning in our codebase and helps maintain a consistent and predictable style throughout the project.

There is also nice video ("`const` was a mistake"):

https://www.youtube.com/watch?v=dqmtzHB2zTM

Issue #368, opened 2024-11-11
Bug: (perfectionist/sort-switch-case) default and 'default' cases are separated


gajus:
### Describe the bug

default and 'default' get sorted and produce invalid code

### Code example

```
switch (variant) {
  default:
  case 'default':
    return 1;

  case 'userList':
    return 2;
}
```

gets fixed to:

```
  switch (variant) {
    case 'default':
      return 1;
    case 'userList':
      return 2;

    default:
  }
```

which produces broken code

### ESLint version

v9.14.0

### ESLint Plugin Perfectionist version

v3.9.1

### Validations

- [X] Read the [docs](https://perfectionist.dev).
- [X] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

azat-io:
Hello! Thank you for your issue!
Sounds strange. It seems we have already fixed this issue in #229.

I'll try to reproduce the problem.

azat-io:
I understand your problem. This is a different problem. At the moment we expect the default case to be at the end.

To be honest, this is the first time I see this style of writing switch-case. This is a bug, we will fix it. Thank you!

hugop95:
This issue is fixed in `next`: the sorting result is

```ts
switch (variant) {
    case 'userList':
        return 2;
    case 'default':

    default:
        return 1;
}
```

azat-io:
@hugop95 Awesome. I unfortunately could not find a corresponding test case in `test/sort-switch-case.test.ts`. Perhaps if it is not there, it should be added.

hugop95:
The issue example can be simplified to

```ts
 switch (variant) {
    default:
    case 'default':
        return 1;
}
```

which gets sorted to

```ts
switch (variant) {
    case 'default':
        return 1;
    default:
}
```

@azat-io Good idea!

azat-io:
Fixed in 960e21c49f56155a13f1bccc47ab35d649bc67e4.
Released in [v4.0.0](https://github.com/azat-io/eslint-plugin-perfectionist/releases/tag/v4.0.0).

If you'd like, you can support the release with a retweet:

PR #369, opened 2024-11-12
test(sort-switch-case): adds unit test for #368

hugop95:
Issue #368 is already fixed in `next`, but this PR adds a dedicated unit test.

- [x] Other


azat-io:
Thanks!

PR #370, opened 2024-11-12
fix(sort-classes): adds missing safety semicolon for `declare class` inline methods

hugop95:
### Description

`declare class` methods act like `abstract` methods and can't have an implementation. As such, they need a safety semicolon when declared inline.

### Example

```ts
declare class Class {
  b(); a()
}
```

gets sorted to

```ts
declare class Class {
  a() b();
}
```

instead of

```ts
declare class Class {
  a(); b();
}
```
This PR fixes this.

### Changes

- Fixes the bug (one line fix).
- Adds a unit test.

- [x] Bug fix


azat-io (approved):

azat-io:
👍

PR #371, opened 2024-11-14
feat: allows users to enter a number for `partitionByNewline`

hugop95:
### Description

Today, `partitionByNewLine` will partition elements with a single newline. In some rules, however (typically `sort-classes`), it's very common to separate `methods` with a newline without wanting to partition them.

This PR allows users to enter a `number` for the `partitionByNewLine` option. That number represents the number of newlines required to partition elements (a common use-case would be `2`).

- Almost no code change (one added function), most lines changed are tests and updated documentation.
- No test impacted.
- `boolean` still supported.
- `default` behavior and value unchanged.

### Remaining to do

- Add tests for each rule.

- [x] New Feature

### DIFF
--- a/docs/content/rules/sort-array-includes.mdx
+++ b/docs/content/rules/sort-array-includes.mdx
@@ -191,9 +191,16 @@ Allows you to use comments to separate the members of arrays into logical groups
 
 ### partitionByNewLine
 
+<sub>
+  type: `boolean | number`
+</sub>
 <sub>default: `false`</sub>
 
-When `true`, the rule will not sort the members of an array if there is an empty line between them. This can be useful for keeping logically separated groups of members in their defined order.
+The rule will not sort the members of an array if:
+- `true` — There is an empty line between them.
+- `number` — There are at least `number` empty lines between them.
+
+This can be useful for keeping logically separated groups of members in their defined order.
 
 ```ts
 if ([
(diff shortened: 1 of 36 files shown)


azat-io:
Hi! I don't really understand if this feature is needed. Could you please tell me why you need it, some use cases?

hugop95:
@azat-io My use case is mostly in `sort-classes`, but can also be applied to `sort-modules`:

In my case, I like to separate all `methods` in `sort-classes` with a newline. But I still would like to be able to use `partitionByNewLine` if I put 2 newlines.

For all other rules though, it's not going to be as useful.

If you feel that this isn't adding value, we can dismiss that PR, there is little code added so I can maintain that on my own side for my personal use.

azat-io:
All right. I just use Prettier and having multiple blank lines in my projects is impossible. I think we can accept this, but I'm not sure if this option will be popular.

I'd like to hear a third opinion. Perhaps @OlivierZal can say something about this.

OlivierZal:
@azat-io, I wouldn't use this rule option for myself but I can imagine that this feature could interest few developers. The design looks good to me, and doesn't hurt, so it seems fine to me.

OlivierZal:
I use prettier too 😜

azat-io:
Okay, let's add this.

PR #375, opened 2024-11-18
feat!: move from natural-compare-lite to natural-orderby

azat-io:
### Description

Hi. This is probably the last breaking change in the upcoming release. But I'm not entirely sure we need it. I'd like to discuss it.

This pull request changes the current library for natural sorting from  [natural-compare-lite](https://github.com/litejs/natural-compare-lite) to [natural-orderby](https://github.com/yobacca/natural-orderby).

#### Pros

- **Supported Library**. The last release of the library we are currently using was 9 years ago. We can't influence its development in any way, we can't inform the author of the library about any problems. At the same time, the last release of the new library was today.

- **Support numeric separators**. The old library came out quite a long time ago and therefore did not support numeric separators (`1_000_000`). The new library started to support this out of the box: https://github.com/yobacca/natural-orderby/pull/231

- **Ability to select locale**. We added new option to select locale for alpabetical sorting (https://github.com/azat-io/eslint-plugin-perfectionist/pull/328). The new library support this also out of the box: https://github.com/yobacca/natural-orderby/pull/240

- **More functionality**. The new library takes into account not only numbers, but can also work with dates, IP addresses and complex numbers.

#### Cons

- **The weight of the library.** The new library is not huge, but it weighs noticeably more than the old one. Is it critical for us?

```
du -sh ./node_modules/natural-compare-lite/
 12K	./node_modules/natural-compare-lite/

du -sh ./node_modules/natural-orderby/
100K	./node_modules/natural-orderby/
```

- **The order is a little different**. You can see the changes in the tests of the current pull-request.

I'd like to hear your opinion, is it worth replacing the old library with the new one? Or is the current library completely satisfied with us and the advantages are not so significant?

@hugop95 @OlivierZal What do you think?

@JoshuaKGoldberg I don't use natural sorting, but I know you do. I would be glad if you could give me your opinion.

### Additional context

N/A.

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/docs/content/rules/sort-array-includes.mdx
+++ b/docs/content/rules/sort-array-includes.mdx
@@ -126,7 +126,7 @@ This rule accepts an options object with the following properties:
 Specifies the sorting method.
 
 - `'alphabetical'` — Sort items alphabetically (e.g., “a” < “b” < “c”) using [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare).
-- `'natural'` — Sort items in a [natural](https://github.com/litejs/natural-compare-lite) order (e.g., “item2” < “item10”).
+- `'natural'` — Sort items in a [natural](https://github.com/yobacca/natural-orderby) order (e.g., “item2” < “item10”).
 - `'line-length'` — Sort items by the length of the code line (shorter lines first).
 
 ### order
--- a/docs/content/rules/sort-classes.mdx
+++ b/docs/content/rules/sort-classes.mdx
@@ -161,7 +161,7 @@ This rule accepts an options object with the following properties:
 Specifies the sorting method.
 
 - `'alphabetical'` — Sort items alphabetically (e.g., “a” < “b” < “c”) using [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare).
-- `'natural'` — Sort items in a [natural](https://github.com/litejs/natural-compare-lite) order (e.g., “item2” < “item10”).
+- `'natural'` — Sort items in a [natural](https://github.com/yobacca/natural-orderby) order (e.g., “item2” < “item10”).
 - `'line-length'` — Sort items by the length of the code line (shorter lines first).
 
 ### order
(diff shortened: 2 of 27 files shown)


OlivierZal:
Hi @azat-io, thanks for this PR.

For me all your pros win against your cons: maybe using a lib actively maintained is very important.

I don't mind if the order is slightly different, provided there is a logic behind the sorting engine.

OlivierZal:
Maybe the biggest change is for `'~/'`, but I don't think that absolute paths are a good practice.

hugop95:
@azat-io Hi 👋

I'm 100% onboard with replacing the old library! I only see pros, and I don't think that users would be bothered by additional elements being logically ordered (if they even notice that the library changed in the first place).

The only con is the size, which in my opinion isn't important that big of a deal here.

JoshuaKGoldberg:
Heh, just seeing this now - looks like a good change to me. If anything comes up I'll file an issue. Thanks for the ping! 🙂

hugop95:
@azat-io It seems that this change particularly affects the `sort-imports` rule: https://github.com/azat-io/eslint-plugin-perfectionist/issues/384.

In particular, `.` is considered lower than `-` with `natural-compare-lite`, but not with `natural-orderby`. This can potentially affect a lot of imports.

What do you think about re-adding the type under a deprecated `natural-legacy` until the next major version? This would also allow use to investigate solutions to keep the previous behavior if needed.

We could choose not to add the dependency ourselves, but to require users to install it themselves if they want to use it. What do you think?

azat-io:
Maybe we should create an issue or correct this in `natural-orderby`?

hugop95:
@azat-io We may do that, but here is what I suspect:
-Theoretically, `natural-orderby` offers this feature through `Custom compare functions` (https://github.com/yobacca/natural-orderby#examples). It would be ideal to allow users to define their own custom functions, but ESLint configs need to be serializable, so this is currently not really possible.
- I haven't looked too much deeper into `natural-compare-lite`, but I'm not sure its exact behavior can be replicated through default options of `natural-orderby`, some parts are probably opinionated.

PR #385, opened 2024-11-19
fix: fixes runtime errors in non typescript-eslint projects 

hugop95:
### Description

Some runtime errors happen in projects that do not user `typescript-eslint` parser.

For example, decorators are not detected (and the `decorators` property will not exist on nodes, even though `typescript-eslint` types will say it necessarily exists).

This PR fixes them and adds for each rule some basic tests to ensure that the rule works fine in a non-typescript environment, using `eslint`'s Rule Tester rather than `typescript-eslint`'s Rule Tester.

- [x] Bug fix


azat-io:
Awesome, thanks a lot!

azat-io (approved):

PR #393, opened 2024-11-23
feat(sort-objects): adds `objectDeclarations ` and `destructuredObjects ` options

### LINKED ISSUE #392: Feature: `sort-objects` Add an option to disable destructured objects sorting

sort-objects

### Describe the problem

When using `sort-objects` rule with `customGroups`, the rule is also applied to destructured objects. I need custom groups to be used only for objects and no sorting (or also customizable one) for destructured objects.

### Code example

Given the config:

```js
import perfectionist from 'eslint-plugin-perfectionist'

export default {
  plugins: {
    perfectionist
  },
  rules: {
    'perfectionist/sort-objects': ['error', {
      customGroups: {
        id: '^id$'
      },
      groups: [
        'id',
        'unknown'
      ],
      type: 'natural'
    }]
  }
}
```

✅ Object is sorted perfectly fine with `id` being first:

```js
const movie = {
  id: 1,
  createdAt: 2024,
  title: 'movie'
}
```

❌ However, function with destructured object also follows the same rules and will give an error:

```js
// Expected "id" (id) to come before "createdAt" (unknown).
function execute({ createdAt, id, title }) {}
``` 

Same goes for React functional components where using destructured `props` object is a common pattern:

```jsx
// Expected "id" (id) to come before "children" (unknown).
export default function Task({ children, id, title }) {
  return (
    <p>hello darkness my old friend</p>
  )
}
```

### Additional comments

There is an option `destructureOnly` which applies the rule only to destructured objects and I kinda need the opposite.

I've also used [eslint-plugin-sort-destructure-keys](https://github.com/mthadley/eslint-plugin-sort-destructure-keys) for sorting destructured objects so it would be a perfect combo.

### Validations

hugop95:
Resolves #392

## Changes

- Deprecates the `destructureOnly` option in the documentation (still usable).
- Adds a `objectDeclarations: boolean` option (by default `true`): when false, will not sort object declarations. 
- Adds a `destructuredObjects: boolean | { groups: boolean }` option (by default `true`): when false, will not sort destructured objects. When `{ groups: false }`, will not use groups to sort destructured objects.

## Advantages of using `{ groups: boolean }` rather than `{ disableGrouping: boolean }`

- Uses the same existing option name `groups`, making it clear that this sub-option is linked with the parent `groups` option.
- Gives flexibility later if users require specifying specific grouping for destructured objects: `{ groups: boolean }` can be extended to `{ groups: boolean | string[] | string[][] }`.

- [x] New Feature

### DIFF
--- a/docs/content/rules/sort-objects.mdx
+++ b/docs/content/rules/sort-objects.mdx
@@ -264,11 +264,29 @@ Allows you to specify names or patterns for object types that should be ignored 
 
 You can specify their names or a regexp pattern to ignore, for example: `'^User.+'` to ignore all object types whose names begin with the word “User”.
 
-### destructureOnly
+### [DEPRECATED] destructureOnly
 
 <sub>default: `false`</sub>
 
-Allows you to sort only objects that are part of a destructuring pattern. When set to `true`, the rule will apply sorting exclusively to destructured objects, leaving other object declarations unchanged.
+Use the [objectDeclarations](#objectdeclarations) and [destructuredObjects](#destructuredobjects) options instead.
+
+Allows you to only sort objects that are part of a destructuring pattern. When set to `true`, the rule will apply sorting exclusively to destructured objects, leaving other object declarations unchanged.
+
+### objectDeclarations
+
+<sub>default: `true`</sub>
+
+Allows you to choose whether to sort standard object declarations.
+
+### destructuredObjects
+
+<sub>
+  type: `boolean | { groups: boolean }`
+</sub>
+<sub>default: `true`</sub>
+
+Allows you to choose whether to sort destructured objects.
+The `groups` attribute allows you to specify whether to use groups to sort destructured objects.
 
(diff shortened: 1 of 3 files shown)


hugop95:
@OlivierZal If you want to review this 👀

OlivierZal (approved):
Looks good!

azat-io (approved):

azat-io:
Nice, thank you!

hckhanh:
@hugop95 How can I replace `destructureOnly` when this property is deprecated? I don't know why.
I don't see any alternative way to do it

hugop95:
Hi @hckhanh ,

Can you try with 
`objectDeclarations: false` and tell me if this solves your problem? If not, please give me an example of what you are trying to do.

hckhanh:
Hi @hugop95 
Thank you. I misunderstood the usage of these options.

PR #396, opened 2024-11-25
feat: improves groups/custom groups for `sort-interfaces` and `sort-object-types`

### LINKED ISSUE #146: Feature: `sort-object-types` add support for `newlines-between`

### LINKED ISSUE #217: Bug: `sort-object-types` Optionality ignores the notion of group
### Describe the bug

Optionality seems to be ignoring the notion of group:

### Code example

Configuration:

```
"perfectionist/sort-object-types": ["error", {
          "groupKind": "optional-first",
          "groups": [
            "callback",
            "unknown"
          ],
          "customGroups": {
            "callback": "on*"
          }
        }],
```

The following code below is invalid (`onClick` belongs to a group before `a`):
```ts
type A =  {
  a

  onClick
}
```

while this one is valid

```ts
type A =  {
  a?

  onClick
}
```

Is optionality supposed to be ignoring groups?

### ESLint version

8.57.0

3.2.0

### Additional comments

_No response_

### Validations

### LINKED ISSUE #395: Feature: Ignore index signature while sorting types

sort-object-types, sort-interfaces

### Describe the problem

I prefer the order like this:

- required 
- optional
- index signature

but the plugin treats index signatures as a required property.

Maybe we can add an options to ignore index signature.

Or is there any options can already make this happen?

### Code example

```ts
export type Foobar = {
  foo: string
  bar?: string
  [key: string]: any
}
```

Config is rule `perfectionist/sort-object-types` with `groupKind: 'required-first'`.

### Additional comments

_No response_

### Validations

hugop95:
Resolves #395.
Resolves #146.
Resolves #217.

## Description

Approximate lines change without tests:
image

Adds the grouping system of `sort-classes` and `sort-modules` to `sort-interfaces` and `sort-exports`.

`sort-interfaces` and `sort-object-types` sort the same things, and they now have the same options.

Deprecates but keeps the `groupKind` option in favor of `groups` (`required-member` and `optional-member` for example).

### Selectors

Ordered from most to least important.
- index signatures: `index-signature`, `member`.
- methods: `method`, `member`.
- properties: `property`, `member`.
- `multiline`: This selector is deprecated: it allows users to keep using the `multiline` group, but the `multiline` modifier should be used instead (`multiline-member` group, for example).

### Modifiers

`required`, `optional`, `multiline`.

This allows for `optional-multiline-method` for example.

## Other features

- **[FEAT]** Adds the `ignorePattern` option to `sort-object-types`: this option was available in `sort-interfaces`.
- **[REFACTOR]** Makes `sort-interfaces` use the same algorithm as `sort-object-types` to sort elements, greatly removing code duplication.
- **[REFACTOR]** We now have 4 rules using the new grouping system: adds some common JSON schemas and functions.

## Affected tests

`sort-interfaces` and `sort-object-types` don't exactly compute node names in the same way: one rule includes the type of methods, the other doesn't.

In this PR, the way both rules compute the name of a node has been uniformized:
- It won't consider the optional (`?`) character in the name of the node.
- For methods, the name will not include the type.

This only affects the nodes names appearing in the error message.

You can see the affected unit tests in this commit: f69c0073418da45c73162083fbc3a5d70f983ed5.

## Commit split

- Ignore [coverage: ignore code that is possibly impossible to cover](https://github.com/azat-io/eslint-plugin-perfectionist/pull/396/commits/c93e9a1c0cd423eb313842cb678a898e7dd1df70): mistake during a rebase.

- [x] New Feature


hugop95:
@OlivierZal Feel free to take a look and give feedback if you wish!

OlivierZal (approved):
Thanks for this feature!

azat-io:
Could you resolve merge conflicts?

hugop95:
@azat-io Done!

rules/sort-interfaces.ts
```diff
@@ -81,252 +26,8 @@ let defaultOptions: Required<Options<string[]>[0]> = {
   groups: [],
 }
 
-export default createEslintRule<Options<string[]>, MESSAGE_ID>({
-  create: context => ({
-    TSInterfaceDeclaration: node => {
```
azat-io, line 86: I'm not sure we should get rid of code duplication here. Could this cause us any problems in the future?

hugop95: @azat-io

> Could this cause us any problems in the future?

There shouldn't be:
- Interface nodes (`TSInterfaceDeclaration`) have a `body.body` field which is typed as `TypeElement[]`.
- Object type nodes (`TSTypeLiteral`) have `members` field which is typed as `TypeElement[]`.

Each object type/interface member acts the same way on a ESLint parsing level.

If we decide to add a feature solely for one rule and not the other (I don't have a use case for this in mind), then we can also just add a parameter to the common `sortObjectTypeElements` function.

I can still copy-paste the code in each rule if you feel like it's better, it's just that we need not to forget to update it in two places when adding a feature to one of the rules.

rules/sort-object-types.ts
```diff
@@ -63,269 +62,388 @@ interface SortObjectTypesSortingNode extends SortingNode<TSESTree.TypeElement> {
...
+    customGroups: {
+      oneOf: [
+        customGroupsJsonSchema,
+        buildCustomGroupsArrayJsonSchema({ singleCustomGroupJsonSchema }),
+      ],
+    },
+    groupKind: {
+      enum: ['mixed', 'required-first', 'optional-first'],
+      description: 'Specifies top-level groups.',
+      type: 'string',
```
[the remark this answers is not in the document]
hugop95: @azat-io I wish we could, but the JSON Schema library that we are using (`import type { JSONSchema4 } from '@typescript-eslint/utils/json-schema'`) does not handle the `deprecated` attribute typescript-wise, so I'm not sure it's supported yet.

azat-io:
Okay, thank you for your answers!

azat-io:
I want to fix some issues in the readme file. I will make a release with this feature tomorrow.

PR #398, opened 2024-11-25
fix: removes `baseUrl` from `tsconfig.json`

hugop95:
### Description

Reduces the potential of import errors by removing the `baseUrl: "."` option in `tsconfig.json`.

`import { pairwise } from 'utils/pairwise'` or similar imports will now raise errors.

I don't think there should be any side effect in this.

- [x] Other


azat-io (approved):

azat-io:
Thanks!

Issue #384, opened 2024-11-19
Bug: Upgrading to v4, named imports are getting rearranged


aeharding:
### Describe the bug

---

With the following v3.9.1:

```
"perfectionist/sort-named-imports": [
  "warn",
  { ignoreCase: false, type: "natural", ignoreAlias: false },
],
```

```
import { LOCALSTORAGE_KEYS, get } from "#/features/settings/syncStorage";
```

✅ Valid

---

v4.0.2:

❌ `Expected "get" to come before "LOCALSTORAGE_KEYS".eslint[perfectionist/sort-named-imports](https://perfectionist.dev/rules/sort-named-imports)`

---

### Code example

See above

### ESLint version

v9.15.0

### ESLint Plugin Perfectionist version

v4.0.2

### Validations

- [X] Read the [docs](https://perfectionist.dev).
- [X] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

hugop95:
Hi @aeharding,

I haven't taken a deeper look yet, but I believe that this is likely due to the following breaking change:
- Move from `natural-compare-lite` to `natural-orderby`.

The implementation of natural sort is different. You may play with `ignoreCase` to try to get your previous order.

hugop95:
`natural-orderby`:
```ts
console.log(naturalCompare('LOCALSTORAGE_KEYS', 'get')) // 1
```

`natural-lite`:
```ts
console.log(naturalCompare('LOCALSTORAGE_KEYS', 'get')) // -1
console.log(naturalCompare('lOCALSTORAGE_KEYS', 'get')) // 1
```

azat-io:
This seems to be a question for [natural-orderby](https://github.com/yobacca/natural-orderby).
Perhaps we should create an issue about case sensitivity support for this project.

Sparticuz:
Another difference that came up for me was that it seemed to put special characters before, so `eslint-comments.js` was moved before `eslint.js`

Before:
```ts
import eslintJSConfig from "./configs/eslint.js";
import eslintCommentsConfig from "./configs/eslint-comments.js";
```

After:
```ts
import eslintCommentsConfig from "./configs/eslint-comments.js";
import eslintJSConfig from "./configs/eslint.js";
```

sthenault:
ooops wrong ping BTW, rather meant to call for @aeharding feedback

[...]

PR #404, opened 2024-12-01
feat: adds custom sort type through custom alphabet

hugop95:
Related to #384.

## Description

The recent removal of the previous `natural-compare-lite` library for the new `natural-orderby` has raised the question of trying to provide users with a way to enter their own custom sort.

This PR offers a way to resolve it.

## Constraints

Here is the main constraint regarding this feature:

ESLint configs must be **serializable**, so allowing users to enter a **sorting function** will not work. The alternative is to let users pass an **ordered list of characters** (that we wil be calling **alphabet**) representing which characters are considered more important than others.

Example:
`aAbBcCdDeEfF...` 

An alphabet can't obviously be restrained to 26 letters, however, and it would be interesting to give a way for users to simply build a custom alphabet.

As such, the PR will be built around **2** main features:

- Allowing users to enter their custom ordered alphabet.
- Giving users a simple way for users to **build** that custom alphabet.

# Feature 1: Allowing users to enter their custom ordered alphabet

This requires:
- Adding an additional `type` on top of the existing `alphabetical`, `natural` and `line-length` ones: `custom`.
- Adding an option for all rules: `alphabet: string`, representing the ordered list of characters of the user's alphabet.

Important notes:
- As mentioned above, an alphabet is more than just letters, numbers and some special characters. By following the [Unicode standard](https://en.wikipedia.org/wiki/Plane_(Unicode)), we can say that a complete alphabet should have **~300000** characters, even though likely 99.999% of characters used in a codebase could be regrouped in less than a couple of hundred characters. More information on this in the **Giving a simple way for users to build that custom alphabet** section.
- ⚠️ Ensure that no performance issue is noticed with large alphabets.
- Generating large alphabets is not something that can be done in the Legacy ESLint JSON format, so this feature is mostly targeting users using javascript-based configs. 
- It is very likely that users would enter an alphabet common to all rules, so there should be a way to enter that alphabet in common ESLint `settings`.
- Characters that are not in an alphabet will have the lowest priority by default (highest if the order is `desc`).

# Feature 2: Giving a simple way for users to build that custom alphabet.

This feature would be much handier if we gave users a way to easily build customizable alphabets. a `Alphabet` class has been created for that purpose. It has multiple utility functions to quickly build an alphabet.

## Alphabet size

`Alphabet.generateRecommendedAlphabet()` will generate ~196000 characters and should very certainly cover all use cases. It will only take into account the Unicode Planes 0 and 1.

`Alphabet.generateCompleteAlphabet()` will generate ~450000 characters. It will take into account the Unicode Plane 0, 1, 2, and 3.

I think it would be interesting in the future to find a way to generate an alphabet with just a couple thousand characters that would still be enough for 99.99% of users. 

### Example 1: Rebuild the old `natural-lite` sort

```ts
import eslintPluginPerfectionist from 'eslint-plugin-perfectionist';
import naturalCompare from 'natural-compare-lite';

const alphabet = eslintPluginPerfectionist.Alphabet
  .generateRecommendedAlphabet()
  .sortBy((a, b) => naturalCompare(a, b))
  .getCharacters();
```

### Example 2: custom `localeCompare` sort

```ts
const alphabet = eslintPluginPerfectionist.Alphabet
  .generateRecommendedAlphabet()
  .sortByLocaleCompare('en-US')
  .prioritizeCase('uppercase') // Will ensure that each character that has a uppercase different than lowercase will have its uppercase first
  .reverse() // `desc` sort order
  .placeCharacterBefore({ characterBefore: '-', characterAfter: '/' }) // `-` should always be more important than `/` 
  .getCharacters();
```

### Example 4: Extremely simple custom `natural` (new library) sort

```ts
const alphabet = eslintPluginPerfectionist.Alphabet
  .generateFrom('abcdefgh...z') // I don't care about other characters, don't consider them when sorting!
  .sortByNaturalSort('en-US')
  .getCharacters();
```

## List of utility functions

All those functions are available in the `Alphabet` class.

#### - `static generateRecommendedAlphabet(): Alphabet`

Generates an alphabet containing relevant characters from the Unicode standard (~196000 characters). Uses the Unicode planes 0 and 1.

#### - `static generateCompleteAlphabet(): Alphabet`

Generates an alphabet containing all characters from the Unicode standard except for irrelevant Unicode planes (~450000 characters). Uses the Unicode planes 0, 1, 2 and 3.

#### - `static generateFrom(values: string[] | string): Alphabet`

Generates an alphabet from the given characters.

#### - `sortByCharCodeAt(): this`

Sorts the alphabet by the character code point.

#### - `sortByNaturalSort(locale?: string): this`

Sorts the alphabet by the natural order of the characters, using `natural-orderby`.

#### - `sortByLocaleCompare(locales?: Intl.LocalesArgument): this`

Sorts the alphabet by the locale order of the characters.

#### - `sortBy(sortingFunction: (characterA: string, characterB: string) => number): this`

Sorts the alphabet by the sorting function provided.

#### - `pushCharacters(values: string[] | string): this`

Adds specific characters to the end of the alphabet.

#### - `removeCharacters(values: string[] | string): this`

Removes specific characters from the alphabet.

#### - `removeUnicodeRange({ start: number, end: number }): this`

Removes specific characters from the alphabet by their range.

#### - `prioritizeCase(casePriority: 'lowercase' | 'uppercase'): this`

For each character with a lower and upper case, permutes the two cases so that the alphabet is ordered by the case priority entered.

```ts
Alphabet.generateFrom('aAbBcdCD')
  .prioritizeCase('uppercase') // Returns 'AaBbCDcd'
```

#### - `placeCharacterBefore({ characterBefore: string; characterAfter: string }): this`

Places a specific character right before another character in the alphabet.

```ts
Alphabet.generateFrom('ab-cd/ef')
.placeCharacterBefore({ characterBefore: '/', characterAfter: '-' }) // Returns 'ab/-cdef'
```

#### - `placeCharacterAfter({ characterBefore: string; characterAfter: string }): this`

Places a specific character right after another character in the alphabet.

```ts
Alphabet.generateFrom('ab-cd/ef')
.placeCharacterAfter({ characterBefore: '/', characterAfter: '-' }) // Returns 'abcd/-ef'
```

#### - `placeAllWithCaseBeforeAllWithOtherCase(caseToComeFirst: 'uppercase' | 'lowercase'): this`

Permutes characters with cases so that all characters with the entered case are put before all the characters with the other case.

#### - `getCharacters(): string`

Retrieves the characters from the alphabet. Called at the very end.

## Limits

It would be nice for users to be able to do
```ts
import { Alphabet } from 'eslint-plugin-perfectionist';
```

Unfortunately, we currently use a `default export` in the plugin, and we can't add a named export `Alphabetical` on top of that. The solution is to add an `Alphabet` field to that default export. Not as clean, but it will work.

Feel free to add comments or suggest changes to these specifications, the code added is WIP.

- [x] New Feature


hugop95:
Hi @OlivierZal @azat-io 👋

Here is a proposal to add a way for users to enter their own custom sort. Don't hesitate to say what you think of the idea!

This branch should be fully buildable, and you should be able to use `type: custom` and `alphabet: <YOUR_ALPHABET>` should you want to try it out.

@sthenault Pinging you as well in case you are interested. 🙂

azat-io:
Wow. That looks very impressive. It's a cool feature. I'll look into the code tomorrow. 👍

OlivierZal:
@hugop95, that's huge and awesome!

OlivierZal:
@hugop95, given the nature, power and usefulness of the `Alphabet` class, couldn't it be a separate module imported as a dependency?

It's probably easier to add it in perfectionist's utils for a POC and test purposes, but maybe you should consider since it's not directly about sorting.

hugop95:
@OlivierZal I think that this is something that can possibly come up with time. It will be easier to adapt the class internally to fit the needs of users who use it for the purpose of `perfectionist` for now. 🙂

OlivierZal:
Sure, that was my guess and it's important the class doesn't miss its target.

hugop95:
@azat-io @OlivierZal I am done with this, feel free to give feedback or test it out!

Don't hesitate to ask if you have any question.

OlivierZal:
Maybe I need to find some time to play with it and appreciate how handy the use of `Alphabet` concretely is, but the code looks great.

azat-io:
I want to take a few days to test it too.

docs/content/configs/recommended-custom.mdx
```diff
@@ -0,0 +1,98 @@
+---
```
azat-io, line 1: Can you also add the config to `docs/pages/configs/index.astro`?

docs/content/configs/recommended-custom.mdx
```diff
@@ -0,0 +1,98 @@
...
+title: recommended-custom
```
azat-io, line 2: And also add config to `docs/utils/pages.ts`?

docs/content/configs/recommended-custom.mdx
```diff
@@ -0,0 +1,98 @@
...
+<CodeTabs
+  code={[
+    {
+      source: dedent`
+        // eslint.config.js
+        import perfectionist from 'eslint-plugin-perfectionist'
+        import naturalCompare from 'natural-compare-lite';
+
+        const myCustomAlphabet = perfectionist
+          .generateRecommendedAlphabet()
```
azat-io, line 49: I don't think the documentation is comprehensive or obvious enough. Maybe we should describe in this file all methods that are available?

docs/content/configs/recommended-custom.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+<CodeTabs
+  code={[
+    {
+      source: dedent`
+        // eslint.config.js
+        import perfectionist from 'eslint-plugin-perfectionist'
+        import naturalCompare from 'natural-compare-lite';
+
+        const myCustomAlphabet = perfectionist
```
azat-io, line 48: Did you mean `perfectionist.Alphabet`?

index.ts
```diff
@@ -135,5 +142,10 @@ export default {
       type: 'natural',
       order: 'asc',
     }),
+    'recommended-custom': createConfig({
+      type: 'custom',
+      order: 'asc',
+    }),
   },
-} as PluginConfig
+  Alphabet,
```
azat-io, line 150: Perhaps this shouldn't be in the default export?

Wouldn't it be better to use a named export?

utils/compare.ts
```diff
@@ -32,70 +33,142 @@ interface NaturalCompareOptions<T extends SortingNode>
...
+      // eslint-disable-next-line no-undefined
+      if (indexOfA === undefined) {
+        indexOfA = Infinity
+      }
+      // eslint-disable-next-line no-undefined
+      if (indexOfB === undefined) {
+        indexOfB = Infinity
+      }
+      if (indexOfA !== indexOfB) {
+        return indexOfA - indexOfB > 0 ? 1 : -1
```
azat-io, line 140: Perhaps we should create a utility in the future:

```ts
let convertBooleanToSign = (boolean: boolean): number => boolean ? 1 : -1
```

Because we have these numbers repeating in a lot of places.

utils/compare.ts
```diff
@@ -32,70 +33,142 @@ interface NaturalCompareOptions<T extends SortingNode>
...
+      indexByCharacters.set(character, index)
     }
+    alphabetCache.set(alphabet, indexByCharacters)
+  }
+  return (aNode: T, bNode: T) => {
+    let aValue = formatString(nodeValueGetter(aNode))
+    let bValue = formatString(nodeValueGetter(bNode))
+    // Iterate character by character
+    // eslint-disable-next-line unicorn/no-for-loop
+    for (let i = 0; i < aValue.length; i++) {
```
azat-io, line 126:
```suggestion
    for (let [i] of aValue.entries()) {
```

utils/compare.ts
```diff
@@ -32,70 +33,142 @@ interface NaturalCompareOptions<T extends SortingNode>
...
+      let indexOfA = indexByCharacters.get(aCharacter)
+      let indexOfB = indexByCharacters.get(bCharacter)
+      // eslint-disable-next-line no-undefined
+      if (indexOfA === undefined) {
+        indexOfA = Infinity
+      }
+      // eslint-disable-next-line no-undefined
+      if (indexOfB === undefined) {
+        indexOfB = Infinity
+      }
```
azat-io, line 138:
```suggestion
      indexOfB ??= Infinity
```

utils/alphabet.ts
```diff
@@ -0,0 +1,446 @@
...
+    }
+    return new Alphabet(
+      arrayValues.map(value =>
+        Alphabet._convertCodepointToCharacter(value.codePointAt(0)!),
+      ),
+    )
+  }
+
+  /**
+   * Generates an alphabet containing relevant characters from the Unicode standard
```
azat-io, line 62: It's not critical, but I'd like us to try to stay within the 80 character limit for string length.

This is necessary for the convenience of working with the code on laptops with a small screen.

I use the following setting in VS Code:

```json
{
  "editor.rulers": [80]
}
```

azat-io:
Do you want `Alphabet` to be capitalized?

utils/compare.ts
hugop95: @azat-io Unfortunately, this won't work: `aValue` is of type `string`, so it doesn't have `entries` like arrays.

index.ts
hugop95: @azat-io I mentioned this limit at the end of the PR description in the `Limits` section:

It would be nice for users to be able to do

`import { Alphabet } from 'eslint-plugin-perfectionist';`

Unfortunately, we currently use a default export in the plugin, and we can't add a named export Alphabetical on top of that.

Changing the default export to a named export would be a breaking change, as users would not be able to do `import perfectionist from 'eslint-plugin-perfectionist'`.

If you manage to find a solution, feel free to suggest it!

docs/content/configs/recommended-custom.mdx
hugop95: @azat-io This was added in f40baa2e3659992b04010b91ad7094765f170227, unless I missed something 🙂

docs/content/configs/recommended-custom.mdx
hugop95: @azat-io I was hesitating about doing that, but you're right, I'll add a section for that.

docs/content/configs/recommended-custom.mdx
hugop95: @azat-io Added in 8e64e71fdfdd8a244200dab9b80d894c33d3e658.

docs/content/configs/recommended-custom.mdx
hugop95: @azat-io Thanks! Fixed in 3c6b360f77bc220a164febe35e7dbc737f5531d9.

utils/compare.ts
hugop95: @azat-io 👍 I added the function in 19eec7b7c59855ab70176a18f12d5efa89f26667.

utils/alphabet.ts
hugop95: @azat-io My bad! I fixed it in 39b649f859e9b79acc7bac7ebe702e70224e7b80. I try to remember it but `prettier` doesn't auto-fix comments unfortunately.

hugop95:
@azat-io 

> Do you want Alphabet to be capitalized?

This is a class, so in my opinion it should be to make it clear for users.

index.ts
OlivierZal: @hugop95 here is the syntax:

```ts
export class Alphabet { ... }

export default { ... }
```

```ts
class Alphabet { ... }

export default { ... }
export { Alphabet }
```

or

```ts
class Alphabet { ... }
const plugin = {}

export { plugin as default, Alphabet }

Comment (hugop95) [reply index.ts:150]: @OlivierZal

The

```ts
import perfectionist from 'eslint-plugin-perfectionist'
```
will not work the same anymore: this will break ESLint for users:

- Current value of `perfectionist`: 
Screenshot 2024-12-08 at 13 47 05

- Value after the change:
Screenshot 2024-12-08 at 13 46 25

OlivierZal: Mmmmmm, I thought it would be fine, I need to perform some tests.

OlivierZal: Ok, under node_modules the plugin is in commonjs and module.exports doesn't support named exports. For the moment I have issues to convert into pure esm – which should support named exports besides default export.

OlivierZal: Ok it's not as simple as I thought. The simplest would probably be a separate module as suggested in https://github.com/azat-io/eslint-plugin-perfectionist/pull/404#issuecomment-2510861165

docs/content/configs/recommended-custom.mdx
hugop95: Done in 8f2ef4c5daae0f0f6c3dc18c7da02f3a25fb9e83.

index.ts
azat-io: Is the problem related to CJS?

How about we make an export from a separate file and add that to `exports` in `package.json`?

hugop95: @azat-io 

> Is the problem related to CJS?

I'll let @OlivierZal answer this 🙂

> How about we make an export from a separate file and add that to exports in package.json?

Added in a100c46845db9a8440205aa94064c1e2c8aebec3:

Users can now import the `Alphabet` class through `import { Alphabet }  from 'eslint-plugin-perfectionist/alphabet'`.

azat-io (approved):
Look good! Thank you, @hugop95.

Before the merge, I'd like to hear @OlivierZal opinion.

OlivierZal:
Hi @azat-io, I had a last look: I think the code is great and the feature is awesome. It’s a go for me!

OlivierZal (approved):

PR #410, opened 2024-12-04
feat: improves function detection

### LINKED ISSUE #409: Bug: issue when migrating to the new `groups` for `sort-object-types` / `sort-object-interfacs`
### Describe the bug

You can see in the commit below that despite the following `groups`:
```ts
  groups: [
    'required-index-signature',
    'optional-index-signature',
    'required-property',
    'optional-property',
    'required-method',
    'optional-method',
  ],
```

- `onSync?` is moved ahead of required methods, although it's an optional method: [here](https://github.com/OlivierZal/melcloud-api/commit/777fce9deea2018089209b00a86d3833bd568ecf#diff-e0899f9bf34e50cd996d55c63b887d2fbd99d1b9cf18264928709c28f33700baR86)
- `tiles` is a required method but is moved ahead of all the other required methods: every occurrence, e.g. [here](https://github.com/OlivierZal/melcloud-api/commit/777fce9deea2018089209b00a86d3833bd568ecf#diff-caf90ffb40aa6dfd6a5b98f50a7b43825066d1eafc25be1b2a4ba7103cca40cdR48)

It seems they are both considered to be properties instead of methods, probably because:
- the first is typed with an alias (`OnSyncFunction`)
- the second includes an intersection in its type

### Code example

https://github.com/OlivierZal/melcloud-api/commit/777fce9deea2018089209b00a86d3833bd568ecf

### ESLint version

v9.16.0

v4.2.0

### Additional comments

_No response_

### Validations

hugop95:
Fixes #409.

### Description

Allows union and intersection types resulting in functions to be detected, such as

```ts
(((v: false) => 'false') | ((v: true) => 'true')) & ((v: any) => any)
```

- [x] Bug fix

### DIFF
--- a/rules/sort-object-types.ts
+++ b/rules/sort-object-types.ts
@@ -27,6 +27,7 @@ import { singleCustomGroupJsonSchema } from './sort-object-types.types'
 import { isNodeEslintDisabled } from '../utils/is-node-eslint-disabled'
 import { allModifiers, allSelectors } from './sort-object-types.types'
 import { hasPartitionComment } from '../utils/is-partition-comment'
+import { isNodeFunctionType } from '../utils/is-node-function-type'
 import { sortNodesByGroups } from '../utils/sort-nodes-by-groups'
 import { getCommentsBefore } from '../utils/get-comments-before'
 import { makeNewlinesFixes } from '../utils/make-newlines-fixes'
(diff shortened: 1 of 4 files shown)


azat-io (approved):

azat-io:
Nice, thank you!

azat-io:
Added in ac1e7c4b5168ecab0028c0880f0fd750e95f3441.
Released in [v4.3.0](https://github.com/azat-io/eslint-plugin-perfectionist/releases/tag/v4.3.0).

PR #411, opened 2024-12-05
feat(sort-objects): handles context-based configurations

### LINKED ISSUE #148: Feature: Context-aware Rules

perfectionist/sort-objects

### Describe the problem

This not strictly limited to the /sort-objects rule, but I'll use it as an example.

Often, the most intuitive ordering is contingent on the _other_ keys present in an object.

For example if an object is intended to represent a color or a quaternion, alphabetical or natural sorting doesn't align with convention.

It's challenging to describe these context-driven rules using `groups` and `custom-groups` without interfering with more general cases, since each key is considered atomically.

### Code example

Natural or alphabetical sort is makes sense in most cases:

```ts
const multipleChoice = {
	a: true,
	b: false,
	c: false
};
```

But the canonical ordering of certain data structures is different, for example:

```ts
// Desired:
const quaternion = {
	x: 0, 
	y: 0,
	z: 0,
	w: 0
};

// Perfectionist:
const quaternion = {
	w: 0
	x: 0, 
	y: 0,
	z: 0,
};
```

Or:
```ts
// Desired:
const color = {
	r: 0.8,
	g: 0.5,
	b: 0.5,
	a: 0.25
};

// Perfectionist:
const color = {
	a: 0.25
	b: 0.5,
	g: 0.5,
	r: 0.8,
};
```

I've come up with a [rather sketchy rule configuration](https://github.com/kitschpatrol/svelte-tweakpane-ui/blob/main/.eslintrc.cjs#L3-L48) that can cover the `quaternion` and `color` and a couple other cases, but still ends up breaking in the `multipleChoice` case... but can't figure out how to cover all cases without some awareness of the what the entire set of keys to be sorted is.

### Additional comments

I'm not sure what the best way to expose this functionality might be, perhaps `groups` could become a record of arrays keyed with some kind of predicate that is run over each key in the object to identify if it's a candidate for a particular group of sorting rules.  

It's also possible I'm just missing something about the possibilities already available through `groups` and `custom-groups`.

Thanks for a great tool.

### Validations

### LINKED ISSUE #408: Feature: Support Logical Ordering for Specific String Literal Types
### Describe the rule

In some cases, string literal types in TypeScript are better represented in a specific logical order rather than an alphabetical one. For example:

```tsx
type Breakpoint = 'xs' | 'sm' | 'md' | 'lg' | 'xl'; // Logical order
```

In this case, the progression from '`xs`' (extra small) to '`xl`' (extra large) follows a clear semantic or domain-specific logic. However, the current sorting rules in `eslint-plugin-perfectionist` would reorder these alphabetically:

```tsx
type Breakpoint = 'lg' | 'md' | 'sm' | 'xl' | 'xs'; // Alphabetical order
```

This disrupts the intended meaning and reduces the readability of such types, which are often used in design systems, configuration objects, or domain-specific types.

### Code example

1. Design Systems: Breakpoints in CSS frameworks ('`xs`', '`sm`', '`md`', '`lg`', '`xl`').

2. User States: User states in a workflow ('pending', 'active', 'completed', 'archived').

3. Error Severity Levels: Severity levels ('low', 'medium', 'high', 'critical').

4. Task Priorities: Task priority levels ('low', 'normal', 'high', 'urgent').

5. Font size: `thin`, `light`, `regular`, `bold`.

### Additional comments

_No response_

### Validations

hugop95:
Resolves #148.
Resolves #408.

## Description

Users have been requesting a way to have different sorting rules depending on context. For example:
- Sizes `xs`, `s`, `m`, `l`, `xl`.
- color: `r`, `g`, `b`.

While custom groups may solve this, having many contextual custom groups that are unrelated to each other makes the configuration hard to read and more prone to errors.

This PR proposes to use the native ESLint feature allowing users to enter multiple option configuration to solve the problem:
- Adds a `useConfigurationIf: { allNamesMatchPattern?: string }` (by default `{}`) to `sort-objects`.

If the `allNamesMatchPattern ` is provided by the user, the entered options configuration will be used if **all** the object's keys match the entered pattern.

## Example

- Detects color objects and sort them by `r`, `g`, `b`.
- Detect size objects and sort them by `xs`, `s`, `m`, `l`, `xl`.
- Sort anything else alphabetically.

Configuration:
```ts
{
    'error',
    {
      groups: ['r', 'g', 'b'],
      customGroups: {
        r: 'r',
        g: 'g',
        b: 'b',
      },
      useConfigurationIf: {
        allNamesMatchPattern: '^r|g|b$',
      },
    },
    {
      groups: ['xs', 's', 'm', 'l', 'xl'],
      customGroups: {
        xs: 'xs',
        s: 's',
        m: 'm',
        l: 'l',
        xl: 'xl',
      },
      useConfigurationIf: {
        allNamesMatchPattern: '^xs|s|m|l|xl$',
      },
    },
    {
      type: 'alphabetical' // Fallback options
    }
  ],
}
```

### Extensibility

The `useConfigurationIf` option is easily extensible with additional options if requested tomorrow.

### Other rules

This option can theoretically be applied to other rules, although the use case is not clear for rules other than `sort-objects`.

- [x] New Feature

### DIFF
--- a/rules/sort-objects.ts
+++ b/rules/sort-objects.ts
@@ -5,6 +5,7 @@ import type { SortingNodeWithDependencies } from '../utils/sort-nodes-by-depende
 import {
   partitionByCommentJsonSchema,
   partitionByNewLineJsonSchema,
+  useConfigurationIfJsonSchema,
   specialCharactersJsonSchema,
   newlinesBetweenJsonSchema,
   customGroupsJsonSchema,
@@ -20,6 +21,7 @@ import {
 } from '../utils/sort-nodes-by-dependencies'
 import { validateNewlinesAndPartitionConfiguration } from '../utils/validate-newlines-and-partition-configuration'
 import { validateGroupsConfiguration } from '../utils/validate-groups-configuration'
+import { getMatchingContextOptions } from '../utils/get-matching-context-options'
 import { getEslintDisabledLines } from '../utils/get-eslint-disabled-lines'
 import { isNodeEslintDisabled } from '../utils/is-node-eslint-disabled'
 import { hasPartitionComment } from '../utils/is-partition-comment'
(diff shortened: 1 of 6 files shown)


rules/sort-objects.ts
```diff
@@ -42,28 +44,29 @@ import { complete } from '../utils/complete'
...
-     */
-    destructureOnly: boolean
-    ignorePattern: string[]
-    order: 'desc' | 'asc'
-    ignoreCase: boolean
-  }>,
-]
+type Options = Partial<{
+  useConfigurationIf: {
+    allElementNamesMatchPattern?: string
```
azat-io, line 49: Maybe we should shorten it to just `pattern' or `matchPattern`?

azat-io:
Awesome idea!

rules/sort-objects.ts
hugop95: @azat-io I wanted to make it clear that **all** key **names** have to match for the configuration to be used (to make a difference if we decide to add a filter where **one** key has to match). I think we can simplify it to `allNamesMatchPattern`

PR #415, opened 2024-12-12
fix(sort-interfaces): prevents constructor declarations from being sorted

### LINKED ISSUE #414: Bug: sort-interfaces changes overload ordering for constructors
### Describe the bug

`sort-interfaces`  changes overload ordering of constructor in interfaces, which can cause existing code to break.

Method overloads do not change order, so constructors should not change too.

### Code example

```
interface Interface {
  new (value: number | string): number;
  new (value: number): unknown;
}
```

### ESLint version

latest

latest

### Additional comments

I have written the following test to verify the issue in this repo.
The case with `new` clause fails, while other cases pass.

```js
ruleTester.run(`${ruleName}(${type}): sorts interface overloads`, rule, {
  invalid: [
    {
      errors: [
        {
          data: {
            right: 'method',
            left: 'methodA',
          },
          messageId: 'unexpectedInterfacePropertiesOrder',
        },
      ],
      output: dedent`
        interface Interface {
          method(value: number | string): number;
          method(value: number): unknown;

          methodA(value: number | string): number;
          methodA(value: number): unknown;
        }
      `,
      code: dedent`
        interface Interface {
          methodA(value: number | string): number;
          methodA(value: number): unknown;

          method(value: number | string): number;
          method(value: number): unknown;
        }
      `,
      options: [options],
    },
  ],
  valid: [
    {
       code: dedent`
         interface Interface {
           method(value: number | string): number;
           method(value: number): unknown;
         }
       `,
       options: [options],
     },
     {
       code: dedent`
         interface Interface {
           new (value: number | string): number;
           new (value: number): unknown;
         }
       `,
       options: [options],
     },
   ],
})
```

### Validations

hugop95:
Fixes #414

### Description

Similarly to call signatures, this PR prevents constructor signatures from being sorted together. 

```ts
interface Interface {
  new (value: number | string): number;
  new (value: number): unknown;
}
```

### Changes

- Fixes the bug
- Adds tests for both `sort-interfaces` and `sort-object-types`.
- Renames a few tests.

- [x] Bug fix

### DIFF
--- a/rules/sort-object-types.ts
+++ b/rules/sort-object-types.ts
@@ -210,7 +210,10 @@ export let sortObjectTypeElements = <MessageIds extends string>({
 
   let formattedMembers: SortObjectTypesSortingNode[][] = elements.reduce(
     (accumulator: SortObjectTypesSortingNode[][], typeElement) => {
-      if (typeElement.type === 'TSCallSignatureDeclaration') {
+      if (
+        typeElement.type === 'TSCallSignatureDeclaration' ||
+        typeElement.type === 'TSConstructSignatureDeclaration'
+      ) {
         accumulator.push([])
         return accumulator
       }
--- a/rules/sort-object-types.ts
+++ b/rules/sort-object-types.ts
@@ -243,10 +243,8 @@ export let sortObjectTypeElements = <MessageIds extends string>({
         name = formatName(
           sourceCode.text.slice(typeElement.range.at(0), endIndex),
         )
-      } else if (
-        typeElement.type === 'TSMethodSignature' &&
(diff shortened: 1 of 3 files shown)


azat-io (approved):

azat-io:
Looks good.

Is it a critical bug?

hugop95:
@azat-io 

> Is it a critical bug?

- It can only break compilation and not runtime.
- It's been here for a long time but wasn't reported until recently.
- It can be temporarily fixed with `// eslint-disable`.

=> Not critical in my opinion.

azat-io:
Thanks!

PR #416, opened 2024-12-12
refactor(sort-classes): make group types readable

azat-io:
### Description

It seems the current types for groups are pretty hard to read. I wrote a generic type to make types more readable. And added tests for types.

### Additional context

N/A.

---

### What is the purpose of this pull request? 

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other

### Before submitting the PR, please make sure you do the following

- [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [x] Ideally, include relevant tests that fail without this PR but pass with it.
- [x] Read [contribution guidelines](https://github.com/azat-io/eslint-plugin-perfectionist/blob/main/contributing.md).

PR #418, opened 2024-12-13
refactor: uses `Join` for other rules

hugop95:
### Changes

- Extract the following types into their own files: `Join`, `SortingNode`.
- Transforms `Join` introduced in https://github.com/azat-io/eslint-plugin-perfectionist/pull/416  to `JoinWithDash` and uses it for other rules.

- [x] Other

### DIFF
--- a/typings/join-with-dash.ts
+++ b/typings/join-with-dash.ts
@@ -1,10 +1,3 @@
-export type JoinWithDash<T extends string[]> = T extends [
-  infer First extends string,
-  ...infer Rest extends string[],
-]
-  ? Rest extends []
-    ? `${First}`
-    : `${WithDashSuffixOrEmpty<First>}${JoinWithDash<Rest>}`
-  : never
+import type { Join } from './join'
 
-type WithDashSuffixOrEmpty<T extends string> = `${T}-` | ''
+export type JoinWithDash<T extends string[]> = Join<T, '-'>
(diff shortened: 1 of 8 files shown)


OlivierZal (approved):

typings/join-with-dash.ts
```diff
@@ -0,0 +1,10 @@
+export type JoinWithDash<T extends string[]> = T extends [
+  infer First extends string,
+  ...infer Rest extends string[],
+]
+  ? Rest extends []
+    ? `${First}`
+    : `${WithDashSuffixOrEmpty<First>}${JoinWithDash<Rest>}`
+  : never
+
+type WithDashSuffixOrEmpty<T extends string> = `${T}-` | ''
```
azat-io, line 10: Looks good. 

I have an idea. Maybe it makes sense to add a second `Separator` parameter?


```ts
export type Join<
  T extends string[],
  Separator extends string = '',
> = T extends [infer First extends string, ...infer Rest extends string[]]
  ? `${First}${Rest['length'] extends 0 ? '' : `${Separator}${Join<Rest, Separator>}`}`
  : ''

export type JoinWithDash<T extends string[]> = Join<T, '-'>
```

hugop95: @azat-io 👍 Improved in a25540611d7d34ca64ff06c6bfbc472b201a33c2.

azat-io (approved):

PR #420, opened 2024-12-13
feat(sort-objects): Adds ignore rule for objects passed as arguments to function calls

### LINKED ISSUE #413: Feature: sort-objects - add ignore rule for objects passed as arguments to function calls
### Describe the rule

There are times when passing objects directly to function calls that you would like the order of the objects to remain in a specific order e.g. with Redux slices, using the 'createSlice' function and passing in an object as first argument, following the documentation you would have 'name', 'initialState', 'reducers', etc, as the keys - there is an inherent logic/hierarchy to the ordering. 

You could add custom groupings for objects in the configuration, however this might conflict in different scenarios; you might not want the 'name' key to be first in one instance, but first in another. One could also use the ignore rule comment for the line in question, or the entire file, but then this introduces a lot of clutter to the files and you might not want to lose the rule for the entire file.

I would propose that there would be an option to specify an array of names of calling functions, to be excluded from the sort-objects rule e.g. 'createSlice'. Perhaps even linking it to the imported package it came from, e.g. '@reduxjs/toolkit/createSlice'.

### Code example

Without the sort-objects rule applied, you would have this arrangement as per the Redux documentation (as an example).

`export const authSlice = createSlice({
  name: 'auth',
  initialState,
  reducers: {
    setUser(state, action) {
      state.user = action.payload;
    }}`

You can't target the object from the configuration as the object is not named - naming the object first and passing it in introduces more clutter to the file, just to avoid the ordering issue.

### Additional comments

_No response_

### Validations

hugop95:
Resolves #413

### Changes

- Adds a `useConfigurationIf.callingFunctionNamePattern` option to `sort-objects`. Object that are not parameters of a function will not be matched by configurations using this option.
- Allows users to enter `unsorted` as `type`.

Example configuration:

```ts
{
    'error',
    {
      type: 'unsorted', // Don't sort objects passed to createSlice
      useConfigurationIf: {
        callingFunctionNamePattern: '^createSlice$',
      },
    },
    {
    }
  ],
}
```

- [x] New Feature


docs/content/rules/sort-array-includes.mdx
```diff
@@ -129,6 +129,7 @@ Specifies the sorting method.
 - `'natural'` — Sort items in a [natural](https://github.com/yobacca/natural-orderby) order (e.g., “item2” < “item10”).
 - `'line-length'` — Sort items by the length of the code line (shorter lines first).
 - `'custom'` — Sort items using the alphabet entered in the [`alphabet`](#alphabet) option.
+- `'unsorted'` — Do not sort items. To be used with the [`useConfigurationIf`](#useConfigurationIf) option.
```
azat-io, line 132: What do you think of the name `'no-sort'` or `'preserve-order'`? Or do you prefer `'unsorted'`?

hugop95: @azat-io 

- I prefer `unsorted` and `no-sort` to `preserve-order` (I feel the first two are more explicit).
- `unsorted` is already used in `customGroups`.
https://github.com/azat-io/eslint-plugin-perfectionist/blob/7fdadd7dcdf96fc259973fad79071d54ceac060f/rules/sort-classes.types.ts#L205-L216

I think it's better for consistency to keep the same name, so I would personally pick `unsorted`.

azat-io: Okay, thank you 👍

azat-io (approved):

azat-io:
Do I understand correctly that we should remove the `sort-objects` scope from commit?

hugop95:
> Do I understand correctly that we should remove the `sort-objects` scope from commit?

Sorry, I'm not sure to understand! What do you mean?

azat-io:
@hugop95 You added scope in commit message `feat(sort-objects)`

hugop95:
@azat-io you can remove it you want, the PR does affects other rules as well (by adding the `type: unsorted` to some other rules too)

PR #421, opened 2024-12-15
perf: improve rules performance

azat-io:
### Description

Some micro-optimizations of the code.

**sort-array-includes:**

- Remove an unnecessary call `getNodeName` function.
- Move the `generatePredefinedGroups` function call outside the loop. Currently, this function is called at each iteration of the loop.
- Pre-group nodes by `groupKind` using `reduce`, reducing sorting complexity from O(n × k) to O(n).
- Replace repeated `indexOf` calls with a `Map` for node indices, reducing order-checking complexity from O(n²) to O(n).

**sort-classes:**

- Move the `generatePredefinedGroups` function call outside the loop. Currently, this function is called at each iteration of the loop.

**sort-decorators:**

- Replace repeated `indexOf` calls with a `Map` for node indices, reducing order-checking complexity from O(n²) to O(n).

**sort-enums:**

- Replace recursion in `extractDependencies` with an iterative approach using an explicit stack. This improves performance by avoiding call stack overhead and prevents stack overflow errors for deeply nested AST nodes.
- Replace repeated `indexOf` calls with a `Map` for node indices, reducing order-checking complexity from O(n²) to O(n).
- Optimize `isNumericEnum` check, replace `Array.prototype.every` with an early-exit loop.

**sort-exports:**

- Replace repeated `indexOf` calls with a `Map` for node indices, reducing order-checking complexity from O(n²) to O(n).

**sort-heritage-clauses:**

- Replace repeated `indexOf` calls with a `Map` for node indices, reducing order-checking complexity from O(n²) to O(n).
- Remove unnecessary checking if node is `undefined`.

**sort-imports:**

- Replace repeated `indexOf` calls with a `Map` for node indices, reducing order-checking complexity from O(n²) to O(n).
- Reduce `formattedMembers.at(-1)` call.
- Replace the current array-based extension check in `isStyle` with a `Set` to enable faster membership testing

**sort-intersection-types:**

- Replace repetitive calls to `defineGroup` in the `switch-case` block with a precomputed mapping of `type` to group, reducing redundant operations and improving readability.
- Replace repeated `indexOf` calls with a `Map` for node indices, reducing order-checking complexity from O(n²) to O(n).
- Reduce `formattedMembers.at(-1)` call.

**sort-jsx-props:**

- Replace repeated `indexOf` calls with a `Map` for node indices, reducing order-checking complexity from O(n²) to O(n).
- Do not check if a prop is multiline if it is already an shorthand

**sort-maps:**

- Replace repeated `indexOf` calls with a `Map` for node indices, reducing order-checking complexity from O(n²) to O(n).

**sort-modules:**

- Replace repeated `indexOf` calls with a `Map` for node indices, reducing order-checking complexity from O(n²) to O(n).
- Move the validation of `selector` and `name` inside the `parseNode` function. This ensures that nodes without these required values are skipped immediately, avoiding unnecessary operations in the main loop.

**sort-named-exports:**

- Replace repeated `indexOf` calls with a `Map` for node indices, reducing order-checking complexity from O(n²) to O(n).

**sort-named-imports:**

- Replace repeated `indexOf` calls with a `Map` for node indices, reducing order-checking complexity from O(n²) to O(n).

**sort-object-types:**

- Move `formatName` function outside loop.
- Reduce `accumulator.at(-1)` call.
- Replace repeated `indexOf` calls with a `Map` for node indices, reducing order-checking complexity from O(n²) to O(n).
- Simplify selector checking.
- Move the `generatePredefinedGroups` function call outside the loop. Currently, this function is called at each iteration of the loop.

**sort-objects:**

- Replace repeated `indexOf` calls with a `Map` for node indices, reducing order-checking complexity from O(n²) to O(n).

**sort-switch-case:**

- Simplify `caseHasBreakOrReturn` function. Reduce its complexity.
- Replace repeated `indexOf` calls with a `Map` for node indices, reducing order-checking complexity from O(n²) to O(n).

**sort-union-types:**

- Replace repetitive calls to `defineGroup` in the `switch-case` block with a precomputed mapping of `type` to group, reducing redundant operations and improving readability.
- Replace repeated `indexOf` calls with a `Map` for node indices, reducing order-checking complexity from O(n²) to O(n).
- Reduce `formattedMembers.at(-1)` call.

**sort-variable-declararations:**

- Replace repeated `indexOf` calls with a `Map` for node indices, reducing order-checking complexity from O(n²) to O(n).

**Utils:**

- **pairwise:** Remove unnecessary variable check.
- **validate-groups-configuration:** Replace the `filter` approach with a `Set` to track already seen groups. By iterating through the array only once and checking membership in the `Set`, the complexity is reduced to linear time (O(n)).

### Additional context

N/A.

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other

### DIFF
--- a/rules/sort-array-includes.ts
+++ b/rules/sort-array-includes.ts
@@ -294,24 +294,18 @@ export let sortArray = <MessageIds extends string>({
   }
 
   for (let nodes of formattedMembers) {
-    let groupedNodesByKind = groupKindOrder.reduce<
-      Record<'literal' | 'spread' | 'any', SortArrayIncludesSortingNode[]>
-    >(
-      (accumulator, groupKind) => {
-        accumulator[groupKind] = nodes.filter(
-          currentNode =>
-            groupKind === 'any' || currentNode.groupKind === groupKind,
-        )
-        return accumulator
-      },
-      { literal: [], spread: [], any: [] },
+    let filteredGroupKindNodes = groupKindOrder.map(groupKind =>
+      nodes.filter(
+        currentNode =>
+          groupKind === 'any' || currentNode.groupKind === groupKind,
+      ),
     )
 
     let sortNodesIgnoringEslintDisabledNodes = (
       ignoreEslintDisabledNodes: boolean,
     ): SortArrayIncludesSortingNode[] =>
-      groupKindOrder.flatMap(groupKind =>
-        sortNodesByGroups(groupedNodesByKind[groupKind], options, {
+      filteredGroupKindNodes.flatMap(groupedNodes =>
+        sortNodesByGroups(groupedNodes, options, {
           getGroupCompareOptions: groupNumber =>
             getCustomGroupsCompareOptions(options, groupNumber),
           ignoreEslintDisabledNodes,
--- a/rules/sort-array-includes.ts
+++ b/rules/sort-array-includes.ts
@@ -168,12 +168,14 @@ export let sortArray = <MessageIds extends string>({
 
   let sourceCode = getSourceCode(context)
   let settings = getSettings(context.settings)
+
   let matchedContextOptions = getMatchingContextOptions({
     nodeNames: elements
       .filter(element => element !== null)
       .map(element => getNodeName({ sourceCode, element })),
     contextOptions: context.options,
   })
+
   let completeOptions = complete(
     matchedContextOptions[0],
     settings,
@@ -198,6 +200,7 @@ export let sortArray = <MessageIds extends string>({
     ruleName: context.id,
     sourceCode,
   })
+
   let formattedMembers: SortArrayIncludesSortingNode[][] = elements.reduce(
     (
       accumulator: SortArrayIncludesSortingNode[][],
@@ -218,11 +221,13 @@ export let sortArray = <MessageIds extends string>({
         selector = 'literal'
       }
 
-      for (let predefinedGroup of generatePredefinedGroups({
+      let predefinedGroups = generatePredefinedGroups({
         cache: cachedGroupsByModifiersAndSelectors,
         selectors: [selector],
         modifiers: [],
-      })) {
+      })
+
+      for (let predefinedGroup of predefinedGroups) {
         defineGroup(predefinedGroup)
       }
 
@@ -236,7 +241,10 @@ export let sortArray = <MessageIds extends string>({
           })
         ) {
           defineGroup(customGroup.groupName, true)
-          // If the custom group is not referenced in the `groups` option, it will be ignored
+          /**
+           * If the custom group is not referenced in the `groups` option, it
+           * will be ignored
+           */
           if (getGroup() === customGroup.groupName) {
             break
           }
@@ -245,11 +253,11 @@ export let sortArray = <MessageIds extends string>({
 
       let sortingNode: SortArrayIncludesSortingNode = {
         isEslintDisabled: isNodeEslintDisabled(element, eslintDisabledLines),
-        name: getNodeName({ sourceCode, element }),
         size: rangeToDiff(element, sourceCode),
         group: getGroup(),
         node: element,
         groupKind,
+        name,
       }
 
       let lastSortingNode = accumulator.at(-1)?.at(-1)
@@ -286,38 +294,51 @@ export let sortArray = <MessageIds extends string>({
   }
 
   for (let nodes of formattedMembers) {
-    let filteredGroupKindNodes = groupKindOrder.map(groupKind =>
-      nodes.filter(
-        currentNode =>
-          groupKind === 'any' || currentNode.groupKind === groupKind,
-      ),
+    let groupedNodesByKind = groupKindOrder.reduce<
+      Record<'literal' | 'spread' | 'any', SortArrayIncludesSortingNode[]>
+    >(
+      (accumulator, groupKind) => {
+        accumulator[groupKind] = nodes.filter(
+          currentNode =>
+            groupKind === 'any' || currentNode.groupKind === groupKind,
+        )
+        return accumulator
+      },
+      { literal: [], spread: [], any: [] },
     )
 
     let sortNodesIgnoringEslintDisabledNodes = (
       ignoreEslintDisabledNodes: boolean,
     ): SortArrayIncludesSortingNode[] =>
-      filteredGroupKindNodes.flatMap(groupedNodes =>
-        sortNodesByGroups(groupedNodes, options, {
+      groupKindOrder.flatMap(groupKind =>
+        sortNodesByGroups(groupedNodesByKind[groupKind], options, {
           getGroupCompareOptions: groupNumber =>
             getCustomGroupsCompareOptions(options, groupNumber),
           ignoreEslintDisabledNodes,
         }),
       )
+
     let sortedNodes = sortNodesIgnoringEslintDisabledNodes(false)
     let sortedNodesExcludingEslintDisabled =
       sortNodesIgnoringEslintDisabledNodes(true)
 
+    let nodeIndexMap = new Map<SortArrayIncludesSortingNode, number>()
+    for (let [index, node] of sortedNodes.entries()) {
+      nodeIndexMap.set(node, index)
+    }
+
     pairwise(nodes, (left, right) => {
-      let indexOfLeft = sortedNodes.indexOf(left)
-      let indexOfRight = sortedNodes.indexOf(right)
+      let leftIndex = nodeIndexMap.get(left)!
+      let rightIndex = nodeIndexMap.get(right)!
+
       let leftNumber = getGroupNumber(options.groups, left)
       let rightNumber = getGroupNumber(options.groups, right)
 
       let indexOfRightExcludingEslintDisabled =
         sortedNodesExcludingEslintDisabled.indexOf(right)
       if (
-        indexOfLeft < indexOfRight &&
-        indexOfLeft < indexOfRightExcludingEslintDisabled
+        leftIndex < rightIndex &&
+        leftIndex < indexOfRightExcludingEslintDisabled
       ) {
         return
       }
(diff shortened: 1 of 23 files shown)


rules/sort-array-includes.ts
```diff
@@ -218,11 +221,13 @@ export let sortArray = <MessageIds extends string>({
         selector = 'literal'
       }
 
-      for (let predefinedGroup of generatePredefinedGroups({
+      let predefinedGroups = generatePredefinedGroups({
```
hugop95, line 225: [nothing to change] 

> Move the `generatePredefinedGroups` function call outside the loop. Currently, this function is called at each iteration of the loop.

Performance wise, the previous and new code is the same:

```ts
for (let predefinedGroup of generatePredefinedGroups({ 
``` 

will not call `generatePredefinedGroups` at each iteration. 🙂

I agree this is cleaner to write it your way though!

rules/sort-classes.ts
```diff
@@ -494,11 +494,13 @@ export default createEslintRule<SortClassesOptions, MESSAGE_ID>({
...
 
-          for (let officialGroup of generatePredefinedGroups({
+          let predefinedGroups = generatePredefinedGroups({
             cache: cachedGroupsByModifiersAndSelectors,
             selectors,
             modifiers,
-          })) {
+          })
+
+          for (let officialGroup of predefinedGroups) {
```
hugop95, line 503: [nit] I would rename `officialGroup` to `predefinedGroup` to align with `predefinedGroups`. 🙂

utils/create-node-index-map.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export let createNodeIndexMap = (
+  nodes: SortingNode[],
+): Map<SortingNode, number> => {
+  let nodeIndexMap = new Map<SortingNode, number>()
+  for (let [index, node] of nodes.entries()) {
+    nodeIndexMap.set(node, index)
+  }
+  return nodeIndexMap
+}
```
hugop95, line 11: [suggestion] Adding a generic extending `SortingNode`:

```suggestion
export let createNodeIndexMap = <T extends SortingNode>(
  nodes: T[],
): Map<T, number> => {
  let nodeIndexMap = new Map<T, number>()
  for (let [index, node] of nodes.entries()) {
    nodeIndexMap.set(node, index)
  }
  return nodeIndexMap
}
```

rules/sort-enums.ts
```diff
@@ -175,11 +174,18 @@ export default createEslintRule<Options, MESSAGE_ID>({
       )
 
       let sortingNodes = formattedMembers.flat()
-      let isNumericEnum = sortingNodes.every(
-        sortingNode =>
-          sortingNode.numericValue !== null &&
-          !Number.isNaN(sortingNode.numericValue),
-      )
+
+      let isNumericEnum = true
```
hugop95, line 178: [perf] Complexity is not changed here: `every` will early exit as soon as it encounters a condition that is `false`:

```ts
let func = (x) => {
    console.log("calling")
    return x === 0
}

[0, 0, 1, 0, 0, 0].every(func)
```
Result:
```ts
calling
calling
calling

Process finished with exit code 0
```

I personally think that the `every` statement is more concise and readable, but that's subjective.

rules/sort-imports.ts
```diff
@@ -214,10 +217,19 @@ export default createEslintRule<Options<string[]>, MESSAGE_ID>({
...
+      '.styl',
+      '.pcss',
+      '.css',
+      '.sss',
+    ])
     let isStyle = (value: string): boolean => {
       let [cleanedValue] = value.split('?')
-      return ['.less', '.scss', '.sass', '.styl', '.pcss', '.css', '.sss'].some(
-        extension => cleanedValue.endsWith(extension),
+      return [...styleExtensions].some(extension =>
```
hugop95, line 231: [perf] Complexity is not changed here: This is still `O(n)` complexity (`n` is the size of `styleExtensions`):

`styleExtensions` being a `Set` has no benefit here, as the initial array has no duplicate, and the set is converted back into an `Array` with `[...styleExtensions]`.

I think that simply doing

```ts
    let styleExtensions = [
      '.less',
      '.scss',
      '.sass',
      '.styl',
      '.pcss',
      '.css',
      '.sss',
    ]
```

will be the optimal.

rules/sort-intersection-types.ts
```diff
@@ -117,69 +118,48 @@ export default createEslintRule<Options, MESSAGE_ID>({
         sourceCode,
       })
 
+      let typeToGroupMap: Record<string, Group> = {
```
hugop95, line 121: [suggestion]

For stronger typing:

```ts
let typeToGroupMap: {[key in TSESTree.TypeNode['type']]?: Group} = {
```

Performance wise, I think that `switch-case` statements and object literals are similar (`O(1)`). 

Readability-wise, I prefer the `switch-case` version because it shows which cases are grouped. The object literal version is not sorted by values here, so while it's theoretically easier to read when searching by type, I think it's handier to regroup by `group` instead. It's just my opinion though.

rules/sort-union-types.ts
```diff
@@ -117,69 +118,48 @@ export default createEslintRule<Options, MESSAGE_ID>({
         sourceCode,
       })
 
+      let typeToGroupMap: Record<string, Group> = {
```
hugop95, line 121: Same comment here as in `sort-intersection-types`.

hugop95 (commented):
@azat-io Great improvements overall 🚀

I've added some comments, I think that having a second opinion on them would be great.

@OlivierZal 👋

rules/sort-modules.ts
```diff
@@ -296,11 +297,12 @@ let analyzeModule = ({
...
         default:
       }
+
+      return !!selector && !!name
     }
     /* eslint-enable @typescript-eslint/no-loop-func */
-    parseNode(node)
 
-    if (!selector || !name) {
+    if (!parseNode(node)) {
```
hugop95, line 305: [readability] I like it the other way
- We have to read `parseNode` to see that the function checks that `selector` and `name` are defined to return `true`.
- Typescript doesn't know that, so we need to add `!` assertions to `name` and `selector`.

rules/sort-object-types-utils.ts
```diff
@@ -51,9 +51,7 @@ export let customGroupMatches = (props: CustomGroupMatchesProps): boolean => {
       props.elementName,
       props.customGroup.elementNamePattern,
     )
-    if (!matchesElementNamePattern) {
-      return false
-    }
+    return matchesElementNamePattern
```
hugop95, line 54: [maintainability] I personally prefer the old way: if we add another condition after this one later, we will need to change

```ts
return matchesElementNamePattern
```

to 

```ts
if (!matchesElementNamePattern) {
  return false
}
```

again so it doesn't exit early, and it's easy to forget.

rules/sort-object-types.ts
```diff
@@ -273,8 +275,9 @@ export let sortObjectTypeElements = <MessageIds extends string>({
       }
 
       if (
-        !selectors.includes('index-signature') &&
-        !selectors.includes('method')
+        !(['index-signature', 'method'] as Selector[]).some(selector =>
```
hugop95, line 278: [readability] 

1. If we keep this:

`const` is better: no type casting!

```suggestion
        !(['index-signature', 'method'] as const).some(selector =>
```

2. However, my brain finds it quite harder to understand

```ts
if (
  !(['index-signature', 'method'] as const).some(selector =>
    selectors.includes(selector),
  )
)
```

than two simple negations.

```ts
if (
  !selectors.includes('index-signature') &&
  !selectors.includes('method')
) {
```

hugop95 (commented):
Some additional things I spotted after a second review.

rules/sort-array-includes.ts
```diff
@@ -286,38 +294,51 @@ export let sortArray = <MessageIds extends string>({
   }
 
   for (let nodes of formattedMembers) {
-    let filteredGroupKindNodes = groupKindOrder.map(groupKind =>
-      nodes.filter(
-        currentNode =>
-          groupKind === 'any' || currentNode.groupKind === groupKind,
-      ),
+    let groupedNodesByKind = groupKindOrder.reduce<
```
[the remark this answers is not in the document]
OlivierZal: Not necessarily speaking about performances but for sure about readability, I like to use the new `Object.prototype.groupBy()` (link)[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/groupBy]

Can be supported thanks to:
```ts
import 'core-js/actual/object/group-by.js'
```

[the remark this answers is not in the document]
azat-io: Back to the previous version.

I don't like the option of adding core-js as a dependency, as that dependency is too big for such a small feature.

rules/sort-classes.ts
azat-io: Renamed.

utils/create-node-index-map.ts
azat-io: Added generic.

rules/sort-imports.ts
azat-io: My bad. Fixed. Thanks!

rules/sort-intersection-types.ts
azat-io: Back to the previous version.

rules/sort-union-types.ts
azat-io: Back to the previous version.

rules/sort-object-types.ts
azat-io: Improved.

rules/sort-object-types-utils.ts
azat-io: Back to the previous version.

rules/sort-modules.ts
azat-io: Back to the previous version.

azat-io:
Sorry for long response. I have fixed all of your comments as requested.

hugop95 (approved):

Issue #401, opened 2024-11-27
Feature: sort-imports - add option to sort within sub groups or remove new line between some groups


RARgames:
### What rule do you want to change?

sort-imports

### Describe the problem

I'm not sure if anything like this is possible right now. I couldn't find it in the docs.
Probably a new option is needed e.g. sort only within sub groups, or remove new line between defined groups

I want to achieve something like this:

```
import React, { useCallback, useState } from 'react';
import { useTranslation } from 'react-i18next';
import classNames from 'classnames';
import { dequal } from 'dequal';
import omit from 'lodash/omit';
import PropTypes from 'prop-types';
```

### Code example

I have `customGroup` for react-related imports called `react`,
I want to position it with first group:
```
 groups: [['react', 'builtin', 'external', 'internal'], ['parent', 'sibling', 'index'], 'unknown'],
 type: 'natural',
```

now my imports are sorted like this:

```
import classNames from 'classnames';
import { dequal } from 'dequal';
import omit from 'lodash/omit';
import PropTypes from 'prop-types';
import React, { useCallback, useState } from 'react';
import { useTranslation } from 'react-i18next';
```

I could change groups to:
```
 groups: ['react', ['builtin', 'external', 'internal'], ['parent', 'sibling', 'index'], 'unknown'],
```

to get:
```
import React, { useCallback, useState } from 'react';
import { useTranslation } from 'react-i18next';

import classNames from 'classnames';
import { dequal } from 'dequal';
import omit from 'lodash/omit';
import PropTypes from 'prop-types';
```

But I cannot get what I want.

The issue is that ['react', 'builtin', 'external', 'internal'] is treated as 1 group for sorting, and I'd like to get sorting within `react` separately from the rest. Maybe some syntax like this could help:
```
groups: [['react', ['builtin', 'external', 'internal']], ['parent', 'sibling', 'index'], 'unknown'],
``` 
Entries in the main array are separated by new line as defined in `newlinesBetween`, but further divisions decide about what is sorted together - just an idea.

### Validations

- [X] Read the [docs](https://perfectionist.dev).
- [X] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

hugop95:
Hi @RARgames 

If I understand correctly, this issue is solely related to newlines and not to sorting itself.
[...]

PR #425, opened 2024-12-20
refactor: improve folder structure

azat-io:
### Description

Updated the project file structure.

### Additional context

#417.

---

### What is the purpose of this pull request? 

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other

### Before submitting the PR, please make sure you do the following

- [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [x] Ideally, include relevant tests that fail without this PR but pass with it.
- [x] Read [contribution guidelines](https://github.com/azat-io/eslint-plugin-perfectionist/blob/main/contributing.md).

PR #428, opened 2024-12-21
feat: adds `customGroup.newlinesInside ` option

hugop95:
First PR aiming to partially resolve the issue https://github.com/azat-io/eslint-plugin-perfectionist/issues/401.
- ⚙️ **PR 1 (this)**: Add `customGroup.newlinesInside` option. We will take the opportunity to create a `getNewlinesBetweenOption` method that will help us later to precisely customize newline behavior between two specific groups. 
- ❌ **PR 2**: Add a way to choose a specific newline behavior between two groups. See the proposals [here](https://github.com/azat-io/eslint-plugin-perfectionist/issues/401#issuecomment-2558581331).

# Description

We currently have the `newlinesBetween` option, which applies the same newlines-related rules between all groups.
This PR adds a sub-option for the new array-based `customGroups`:
- `newlinesInside?: 'always' | 'never'`

Here is the list of rules with this new addition:
- `sort-classes`
- `sort-modules`
- `sort-object-types`
- `sort-interfaces`

# New option

## `customGroup.newlinesInside?: 'ignore' | 'always' | 'never'`

Newlines rule that should be applied between two consecutive nodes of the **same** custom group.

Example: Enforce a newline between all class methods.
```ts
{
  customGroups: [
    {
      groupName: "methodsWithNewlinesBetween",
      selector: "method",
      newlinesInside: "always"
    }
  ]
}
```

# Other changes

- Unifies the way spacing errors are reported:
  - The `leftNum < rightNum` check is currently only made if `numberOfEmptyLinesBetween === 0` (as confirmed by passing tests).
  - With `newlinesBetween: always` and if `numberOfEmptyLinesBetween > 1`, we were not checking if `leftNum < rightNum`.
  - The `leftNum < rightNum` check is not made when `newlinesBetween: never` either (as confirmed by passing tests).
  - => Always check if `leftNum < rightNum`.
  - The few tests impacted can be found in those commits: 04132ee4bd1bbbebf2b0c27f802cf77e8aa829ea, 252873c1a821ce92ae7307396d499f6b8df29452.
    - Only some spacing errors reported were removed, outputs are unchanged.

- [x] New Feature

### DIFF
--- a/test/get-newlines-between-option.test.ts
+++ b/test/get-newlines-between-option.test.ts
@@ -1,0 +1,205 @@
+import { describe, expect, it } from 'vitest'
+
+import type { GetNewlinesBetweenOptionParameters } from '../utils/get-newlines-between-option'
+import type { SortingNode } from '../types/sorting-node'
+
+import { getNewlinesBetweenOption } from '../utils/get-newlines-between-option'
+
+describe('get-newlines-between-option', () => {
+  describe('global "newlinesBetween" option', () => {
+    it('should return the global option if "customGroups" is not defined', () => {
+      expect(
+        getNewlinesBetweenOption(
+          buildParameters({
+            newlinesBetween: 'ignore',
+          }),
+        ),
+      ).toBe('ignore')
+    })
+
+    it('should return the global option if "customGroups" is not an array', () => {
+      expect(
+        getNewlinesBetweenOption(
+          buildParameters({
+            newlinesBetween: 'ignore',
+            customGroups: {},
+          }),
+        ),
+      ).toBe('ignore')
+    })
+
+    it('should return "ignore" if "newlinesBetween" is "ignore"', () => {
+      expect(
+        getNewlinesBetweenOption(
+          buildParameters({
+            newlinesBetween: 'ignore',
+          }),
+        ),
+      ).toBe('ignore')
+    })
+
+    it('should return "never" if "newlinesBetween" is "never"', () => {
+      expect(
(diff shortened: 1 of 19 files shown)


hugop95:
@OlivierZal 👋 Feel free to review this if you want 🙂

test/get-newlines-between-option.test.ts
```diff
@@ -0,0 +1,205 @@
+import { describe, expect, it } from 'vitest'
```
azat-io, line 1: Can you move this file to `test/utils/`?

hugop95: @azat-io Oops, rebase from https://github.com/azat-io/eslint-plugin-perfectionist/pull/425 didn't move it 😅. Moved in f4b4ad5060d0c87da7621928c4d73e5ac8c53644.

OlivierZal (approved):

test/get-newlines-between-option.test.ts
OlivierZal: Looks good to me!

test/rules/sort-imports.test.ts
```diff
@@ -290,13 +290,6 @@ describe(ruleName, () => {
               },
               messageId: 'unexpectedImportsGroupOrder',
             },
-            {
```
azat-io, line 293: Can you tell me why you removed these tests?

azat-io (approved):

PR #433, opened 2024-12-24
feat: makes `partitionByComment` support `line` and `block` filters

### LINKED ISSUE #432: Feature: `partitionByComment` support `line` and `block`

any rule support `partitionByComment`

### Describe the problem

Mostly is issue mixed with `jsdoc`.

Comments are used to explain code but also causes sorting behavior unexpected.

### Code example

With `sort-objects` config to `{ type: "alphabetical", order: "asc", partitionByComment: true, }`

```ts
export const foobar = {
  foo: 'foo',

  /**
   * This is bar
   */
  bar: 'bar',
}
```

If `partitionByComment: 'line'` is supported. It can be fixed automaticly.

### Additional comments

_No response_

### Validations

hugop95:
Resolves #432

Most of the lines added are tests and documentation.

### Description

Updates the `partitionByComment` option to support `block` and `line` filters:

```ts
type T = {
        block?: string[] | boolean | string;
        line?: string[] | boolean | string;
    }
    | string[]
    | boolean
    | string
```

### Affected rules

- `sort-enums`
- `sort-array-includes`
- `sort-sets`
- `sort-decorators`
- `sort-object-types`
- `sort-interfaces`
- `sort-modules`
- `sort-classes`
- `sort-variable-declarations`
- `sort-union-types`
- `sort-intersection-types`
- `sort-objects`
- `sort-named-imports`
- `sort-named-exports`
- `sort-maps`
- `sort-imports`
- `sort-exports`

- [x] New Feature

### DIFF
--- a/docs/content/rules/sort-array-includes.mdx
+++ b/docs/content/rules/sort-array-includes.mdx
@@ -192,11 +192,6 @@ Allows you to group array elements by their kind, determining whether spread val
 
 ### partitionByComment
 
-<sub>
-  - type:
-    - `boolean | string | string[]` (`BaseType`)
-    - `{ block: BaseType; line: BaseType }`
-</sub>
 <sub>default: `false`</sub>
 
 Allows you to use comments to separate the members of arrays into logical groups. This can help in organizing and maintaining large arrays by creating partitions based on comments.
@@ -205,7 +200,7 @@ Allows you to use comments to separate the members of arrays into logical groups
 -	`false` — Comments will not be used as delimiters.
 - `string` — A regexp pattern to specify which comments should act as delimiters.
 - `string[]` — A list of regexp patterns to specify which comments should act as delimiters.
-- `{ block: BaseType; line: BaseType }` — Specify which block and line comments should act as delimiters.
+- `{ block: boolean | string | string[]; line: boolean | string | string[] }` — Specify which block and line comments should act as delimiters.
 
 ### partitionByNewLine
 
--- a/docs/content/rules/sort-array-includes.mdx
+++ b/docs/content/rules/sort-array-includes.mdx
@@ -192,6 +192,11 @@ Allows you to group array elements by their kind, determining whether spread val
 
 ### partitionByComment
 
+<sub>
+  - type:
+    - `boolean | string | string[]` (`BaseType`)
+    - `{ block: BaseType; line: BaseType }`
+</sub>
 <sub>default: `false`</sub>
 
 Allows you to use comments to separate the members of arrays into logical groups. This can help in organizing and maintaining large arrays by creating partitions based on comments.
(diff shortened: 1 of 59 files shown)


OlivierZal (approved):

docs/content/rules/sort-array-includes.mdx
```diff
@@ -192,6 +192,11 @@ Allows you to group array elements by their kind, determining whether spread val
 
 ### partitionByComment
 
+<sub>
+  - type:
+    - `boolean | string | string[]` (`BaseType`)
```
azat-io, line 197: Are you sure we need this block? We just have default values in other places.

docs/content/rules/sort-array-includes.mdx
```diff
@@ -200,6 +205,7 @@ Allows you to use comments to separate the members of arrays into logical groups
 -	`false` — Comments will not be used as delimiters.
 - `string` — A regexp pattern to specify which comments should act as delimiters.
 - `string[]` — A list of regexp patterns to specify which comments should act as delimiters.
+- `{ block: BaseType; line: BaseType }` — Specify which block and line comments should act as delimiters.
```
azat-io, line 208: I'm not sure it's clear what BaseType means here.

hugop95: @azat-io The idea is to tell that `block` and `line` can be `string | string[] | boolean`.

I felt that 
```ts
`{ block: string | string[] | boolean; line: string | string[] | boolean }`
```

was a bit hard to read but we can do that if you feel that it's better.

docs/content/rules/sort-array-includes.mdx
azat-io: @hugop95 What do you think?

hugop95: @azat-io We can keep `boolean | string | string[]` but, we also need to say that users can also write `{ block: string | string[] | boolean; line: string | string[] | boolean }`.

We can remove that block and go with

```ts
### partitionByComment

<sub>default: `false`</sub>

Allows you to use comments to separate the exports into logical groups. This can help in organizing and maintaining large export blocks by creating partitions based on comments.

- `true` — All comments will be treated as delimiters, creating partitions.
-	`false` — Comments will not be used as delimiters.
- `string` — A regexp pattern to specify which comments should act as delimiters.
- `string[]` — A list of regexp patterns to specify which comments should act as delimiters.
- `{ block: string | string[] | boolean; line: string | string[] | boolean }` — Specify which block and line comments should act as delimiters.
```

even though the last line is a bit long. What do you think?

azat-io: I think it makes more sense that way.

hugop95: @azat-io Updated in 566872ab88a737733a76b15ac7d6f6144216ba35!

Issue #11, opened 2023-06-09
Bug: Fix objects must not be overlapped in a report.


Stringsaeed:
### Describe the bug

When running this command
```sh
eslint . --ext .js,.jsx,.ts,.tsx --fix --cache
```

I get this error 

```sh

Oops! Something went wrong! :(

ESLint: 8.41.0


AssertionError [ERR_ASSERTION]: Fix objects must not be overlapped in a report.
    at mergeFixes (PROJECT_PATH/node_modules/eslint/lib/linter/report-translator.js:152:9)
    at normalizeFixes (PROJECT_PATH/node_modules/eslint/lib/linter/report-translator.js:182:16)
    at PROJECT_PATH/node_modules/eslint/lib/linter/report-translator.js:349:49
    at Object.report (PROJECT_PATH/node_modules/eslint/lib/linter/linter.js:1021:41)
    at PROJECT_PATH/node_modules/eslint-plugin-perfectionist/dist/index.js:3:217
    at I (PROJECT_PATH/node_modules/eslint-plugin-perfectionist/dist/index.js:1:2481)
    at PROJECT_PATH/node_modules/eslint-plugin-perfectionist/dist/index.js:3:143
    at Array.forEach (<anonymous>)
    at Program:exit (PROJECT_PATH/node_modules/eslint-plugin-perfectionist/dist/index.js:3:131)
    at ruleErrorHandler (PROJECT_PATH/node_modules/eslint/lib/linter/linter.js:1050:28)
    at PROJECT_PATH/node_modules/eslint/lib/linter/safe-emitter.js:45:58
    at Array.forEach (<anonymous>)
    at Object.emit (PROJECT_PATH/node_modules/eslint/lib/linter/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (PROJECT_PATH/node_modules/eslint/lib/linter/node-event-generator.js:297:26)
    at NodeEventGenerator.applySelectors (PROJECT_PATH/node_modules/eslint/lib/linter/node-event-generator.js:326:22)
    at NodeEventGenerator.leaveNode (PROJECT_PATH/node_modules/eslint/lib/linter/node-event-generator.js:349:14)
    at CodePathAnalyzer.leaveNode (PROJECT_PATH/node_modules/eslint/lib/linter/code-path-analysis/code-path-analyzer.js:816:23)
    at PROJECT_PATH/node_modules/eslint/lib/linter/linter.js:1087:32
[...]

PR #441, opened 2025-01-10
fix: fixes overlapping errors between `newlinesBetween` and `comment-after` fixes

hugop95:
Related issue: https://github.com/azat-io/eslint-plugin-perfectionist/issues/11
Related comment: https://github.com/azat-io/eslint-plugin-perfectionist/issues/11#issuecomment-2581673008

## Description

Note: good news! I've managed to find a way to fix this without impacting any existing test 🙂

In some specific cases, comments-related fixes and newline fixes may overlap.

### Example

```ts
type Test = {
  onAdd: () => void;
  title: string; // The fix related to this comment will overlap with newline fixes

  onClick: () => void; 
};
```

```ts
'perfectionist/sort-object-types': [
  'error',
  {
    type: 'natural',
    groups: [
      'property',
      'required-method',
    ],
    newlinesBetween: 'always',
  },
]
```

Today, both comment/order fixes and newlines fixes are done in the same cycle. The easiest fix is to separate those fixes into a different cycle (see below).

## Proposal

The PR includes a refactor part and a simple fix that aims to separate the **"comments after"** fixes and the **newlines** fixes. **"comments after"** fixes are fixes related to comments appearing at the end of lines, after nodes.

### State today

- `makeFixes` handles:
  - **Order**-related fixes.
  - **Comments after** fixes.
- `makeNewlinesFixes` handles only newline fixes.

### Commits split

The refactor commits (first 5 commits) rename and split existing functions into new functions in order to separate each type of fixes independently.

This allows us in the end to do

```ts
  let orderFixes = makeOrderFixes({...})

  let commentAfterFixes = makeCommentAfterFixes({...})
  if (commentAfterFixes.length) {
    return [...orderFixes, ...commentAfterFixes]
  }

  let newlinesFixes = makeNewlinesFixes({...})
  return [...orderFixes, ...newlinesFixes]
```

- [x] Bug fix


azat-io:
Looks good. I am a bit confused by the long function names. And I want to take some time to think if it can be made a bit simpler.

hugop95:
@azat-io I've simplified things in the last two commits:
- https://github.com/azat-io/eslint-plugin-perfectionist/pull/441/commits/8732a401a0a86da5956b3c8d1ee8bf6c5db43d0f
- https://github.com/azat-io/eslint-plugin-perfectionist/pull/441/commits/ce18e182552e57b0986be1fe3d5929724f3e80e7

The remaining **fixies** functions are
- `makeAllFixes`
- `makeNewlinesFixes`
- `makeCommentAfterFixes`
- `makeOrderFixes`

rules/sort-array-includes.ts
```diff
@@ -39,6 +38,7 @@ import { createEslintRule } from '../utils/create-eslint-rule'
 import { getLinesBetween } from '../utils/get-lines-between'
 import { getGroupNumber } from '../utils/get-group-number'
 import { getSourceCode } from '../utils/get-source-code'
+import { makeAllFixes } from '../utils/make-all-fixes'
```
azat-io, line 41: Why not `makeFixes`?

hugop95: @azat-io Renamed in ee1f5eb1ec5163aed0a0f82b43dff4c216a4b3d5.

PR #455, opened 2025-01-30
feat(sort-named-exports): adds `ignoreAlias` option

### LINKED ISSUE #454: Feature: sort-named-exports needs ignoreAlias

sort-named-exports

### Describe the problem

[sort-named-exports](https://perfectionist.dev/rules/sort-named-exports) needs an `ignoreAlias` option (as [sort-named-imports](https://perfectionist.dev/rules/sort-named-imports#ignorealias) has)

### Code example

This export statement
```typescript
export {
  throttle as a,
  calculateAge as b,
  formatDate as c,
  generateUUID as d,
  debounce as e,
  parseQueryString as f
} from "./utils";
```
will always be sorted as it is ☝️ (by aliases). But the perfectionist inside me says that there should be an option to sort it as:
```typescript
export {
  calculateAge as b,
  debounce as e,
  formatDate as c,
  generateUUID as d,
  parseQueryString as f,
  throttle as a
} from "./utils";
```

### Additional comments

Thank you for this awesome stuff!

### Validations

hugop95:
Resolves #454

# Description

This PR adds the `ignoreAlias` option (similar to `sort-named-imports`) to `sort-named-exports`.

# ℹ️ Note

According to the AST, the following syntax is allowed

```ts
export { 'literal' as local } from './import'
```

I don't think this is valid in any Typescript version, but I've added a handler for that case anyway since it costs nothing 🤷
229fe1f581e28c60f85bf42708e8d4c844355ba2

- [x] New Feature


azat-io (approved):

azat-io:
Nice, thank you!

PR #461, opened 2025-02-05
fix: fixes complex decorators not being handled

### LINKED ISSUE #460: Bug: Unexpected decorator expression's callee type
### Describe the bug

Hey,

I think the code doesn't accept decorators that come from a namespace.

This doesn't work

```
import * as TGql from "type-graphql";

export const Resolver = (deps: Dependencies) => {
  @TGql.Resolver(Model)
  class InnerResolver {
    // Issue is below
    @TGql.Query(() => Query)
    // Issue is above
    @TGql.Authorized()
    methodName() {}
  }
}
```

because [this line](https://github.com/azat-io/eslint-plugin-perfectionist/blob/a490cce9565d66cffbc79c709bc018bb9f315118/rules/sort-decorators/get-decorator-name.ts#L6) expects the `callee` for the `CallExpression` to be an `Identifier` even though it can be a `MemberExpression`.

Thanks for all the hard work.

### Code example

```
import * as TGql from "type-graphql";

export const Resolver = (deps: Dependencies) => {
  @TGql.Resolver(Model)
  class InnerResolver {
    // Issue is below
    @TGql.Query(() => Query)
    // Issue is above
    @TGql.Authorized()
    methodName() {}
  }
}
```

### ESLint version

v9.19.0

v4.8.0

### Additional comments

_No response_

### Validations

hugop95:
Fixes #460

## Description

Complex decorators such as `@A.B(() => C)` are currently not handled. This PR fixes this.

## Proposed solution

We used to get the decorator name by switch-casing on the type of decorator. It turns out to be simpler to use `sourceCode.getText(decorator)`, remove the first character `@`, and split at the first `(` instead.

- [x] Bug fix


azat-io (approved):

azat-io:
Thank you! Is it critical bug?

OlivierZal:
I would rather say it's not critical: decorator sorting is a nice to have since the order matters, and needs custom config to avoid sorting to be broken.

hugop95:
@azat-io cc @OlivierZal @FredericEspiau

I would say it's not critical as well:
- `sort-classes` and `sort-modules` will not throw an error without the fix.
- `sort-decorators` will throw an error if the codebase uses decorators in namespaces (`@A.B`)
  - But the rule is not one of the major ones, and the conditions to reproduce the bug make me think that its impact on actual users is limited.

FredericEspiau:
Understandable and thanks for the hard work and the prompt fix

Issue #465, opened 2025-02-10
Bug: Getting 'Expected dependency "b" to come before "a"' for function properties in classes


OmniSliver:
### Describe the bug

I'm getting an `Expected dependency "b" to come before "a"` error since v4.7.0 when a function property depends on another.
This didn't happen on v4.6.0 and still happens on v4.8.0.
This is using the plain `recommended-alphabetical` config.

### Code example
```javascript
class MyClass {
  a = () => this.b()

  b = () => null
}
```

### ESLint version

v9.20.0

### ESLint Plugin Perfectionist version

v4.8.0

### Validations

- [x] Read the [docs](https://perfectionist.dev).
- [x] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

PR #466, opened 2025-02-10
fix(sort-classes): fixes dependency detection regression

hugop95:
Fixes https://github.com/azat-io/eslint-plugin-perfectionist/issues/465

### Description

Commit https://github.com/azat-io/eslint-plugin-perfectionist/commit/85c614a87bfc77047d298de259d1644548c1f880 introduced a regression in dependency detection in some cases of `sort-classes`.

The fix is in [this commit](https://github.com/azat-io/eslint-plugin-perfectionist/pull/466/commits/4b48cb16393fbb393840f2f27cbd1683ae499906). It applies what existed before the breaking commit:
https://github.com/azat-io/eslint-plugin-perfectionist/commit/85c614a87bfc77047d298de259d1644548c1f880#diff-468c11407b72f2bd0648f49662a8f912e9b6875205f5428a2ed862c406ba21d1L470-L481

### Changes

- Fixes the bug and adds a test.
- Introduces [`UnreachableCaseError`](https://github.com/ts-essentials/ts-essentials/tree/master/lib/functions/unreachable-case-error) to increase compilation safety in switch clauses.

- [x] Bug fix


rules/sort-classes.ts
```diff
@@ -332,8 +333,74 @@ export default createEslintRule<SortClassesOptions, MESSAGE_ID>({
           let selectors: Selector[] = []
           let addSafetySemicolonWhenInline: boolean = true
           switch (member.type) {
-            case 'TSAbstractMethodDefinition':
-            case 'MethodDefinition': {
+            case TSESTree.AST_NODE_TYPES.TSAbstractPropertyDefinition:
```
azat-io, line 336: Why?

hugop95: @azat-io I find it cleaner, but can revert it if needed

hugop95: @azat-io 

> I think it's better to use strings rather than enum values. It looks more clean.

Updated in `sort-maps`: cd87039ac3f817eb17aa5ad1aaaaf750f6770d48.

For `sort-classes`, it's actually a bit more problematic: I've just done tests and there are compilation differences between using the string and enum versions of the AST types.

For example, in `sort-classes`, removing `TSESTree.AST_NODE_TYPES` and using strings will prevent typescript from knowing that the `default` case is not reachable.

image

I've still changed enums to strings in 8834afb3c9add3c0408971ee8aa915cc8e5e3aef, but I had to remove `UnreachableCaseError`.

If you think that using `UnreachableCaseError` outweighs the cons of using enums rather than string, I can revert that last commit. Otherwise, feel free to merge if you think that it's good.

hugop95: > I'm also thinking of including the erasableSyntaxOnly option in tsconfig.json in the future:

That's exciting news, but I've never used that feature, so I can't say what impact it will have on this codebase 🙂

azat-io (approved):

Issue #68, opened 2023-09-07
Sort keys by type


Grawl:
### Describe the problem

I want to sort object properties by its type

### Code example

I have the following type object:

```ts
type MyObject = {
  updatedAt: Date
  count: number
  createdAt: Date
}
```

I want to colocate properties by type, so `Date` will be next to each other:

```ts
type MyObject = {
  count: number
  createdAt: Date
  updatedAt: Date
}
```

### Validations

- [X] Read the [docs](https://eslint-plugin-perfectionist.azat.io).
- [X] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

hugop95:
Commenting because I like the idea, but some subjects need to be addressed:
- How to deal with union/intersection types?
- What takes precedence between alphabetical sorting and type name sorting?

azat-io:
Added in 79d044161c414838eff7110c712fd82c69ba1ac7.
Released in [v4.10.0](https://github.com/azat-io/eslint-plugin-perfectionist/releases/tag/v4.10.0).

PR #469, opened 2025-02-13
feat: add `sortBy: "value"` related-options

hugop95:
Should resolve https://github.com/azat-io/eslint-plugin-perfectionist/issues/68

# Description

This PR adds **3** `sort by value` related-options to `sort-object-types` and `sort-interfaces`:
- `customGroups.elementValuePattern` to match property patterns.
- `sortBy: 'name' | 'value'` (by default `name`).
- `customGroups.sortBy: 'name' | 'value'` to override the `sortBy` option in a given custom group.

# Example

Sort properties by:
- `ObjectId`.
- `Date`.
- Primitive types.
- All the rest, sorted by name alphabetically.

Users have two ways to write this: use `sortBy: 'value'` by default, and for the last group, use `sortBy: 'name'`, or the opposite.

<details>
<summary>Details</summary>

```json
{
"perfectionist/sort-object-types": [
    "error",
    {
      "groups": [
        "objectIds",
        "dates",
        "primitives"
      ],
      "customGroups": [
        {
          "groupName": "objectIds",
          "elementValuePattern": "^ObjectId$",
        },
        {
          "groupName": "dates",
          "elementValuePattern": "^Date$",
        },
        {
          "groupName": "primitives",
          "elementValuePattern": "string|number|boolean",
          "sortBy": "value"
        }
      ]
    }
  ]
}	
```

Or

```json
"perfectionist/sort-object-types": [
    "error",
    {
      "sortBy": "value",
      "groups": [
        "objectIds",
        "dates",
        "primitives",
        "allTheRestSortedNormally"
      ],
      "customGroups": [
        {
          "groupName": "objectIds",
          "elementValuePattern": "^ObjectId$"
        },
        {
          "groupName": "dates",
          "elementValuePattern": "^Date$"
        },
        {
          "groupName": "primitives",
          "elementValuePattern": "string|number|boolean"
        },
        {
          "groupName": "allTheRestSortedNormally",
          "sortBy": "name"
        }
      ]
    }
  ],
```

Result:

```ts
type Type = {
  _id: ObjectId;
  groupId: ObjectId;
  createdAt: Date;
  lastLoginAt: Date;
  b: boolean;
  c: number;
  a: string;
  // We are sorted alphabetically!
  anotherObject: AnotherObject;
  someValue: AnObject;
};
```

</details>

# Important notes

## ❓Can users sort by value alphabetically, then by name alphabetically if the values are the same ?

Currently, no. This means that 

```json
{
  "perfectionist/sort-object-types": [
    "error",
    {
      "sortBy": "value",
    }
  ],
}
```

will tolerate both

```ts
type Types = {
  a: string
  b: string
}
```

and

```ts
type Types = {
  b: string
  a: string
}
```

We would need to allow users to write

```ts
"perfectionist/sort-object-types": [
    "error",
    {
      "sortBy": "value",
	  "fallbackSort": { "sortBy": "name"}
    }
  ],
```

This can come in a later PR.

## ❓How does `sortBy: 'value'` work with **non-properties** ?

It depends if it's the base `sortBy` option, or the `customGroups` one.

### Base `sortBy` option

Non-properties nodes will **not** have a sort order enforced.

```ts
type Types = {
  z(): void
  a: string
}
```
and 
```ts
type Types = {
  b: string
  z(): void
}
```

are both allowed.

### `customGroups.sortBy` option

**Grouping** is still enabled, but within the group, the sort order is not enforced. This essentially means that we are enforcing the non-property member (such as functions) matching the group to be in that group, but the user can choose to move that member inside that group wherever they want.

## Commit split

- Commits **1-5** are refactors.
- Commit **6+** are the actual implementation.

- [x] New Feature


azat-io:
LGTM.

In the beginning, I thought about using Typescript to solve this problem.

To have TypeScript ESLint detect the type of the value and sort them according to the type. For example, if it is a JSX props object. And the types are specified somewhere else.

```ts
interface Props {
  isOpen: boolean
  isCompact: boolean
  otherProp: string
  value: string
  onChange: () => {}
}

// For example if booleans should be last

export let Component: FC<Props> = ({
  onChange,
  otherProp
  value,
  isCompact,
  isOpen,
}) => {}
```

P.S. At least I want to try to do it in another plugin:
https://github.com/azat-io/eslint-plugin-de-morgan/pull/3

hugop95:
@azat-io I think that we would need to dive deeper into how `typescript-eslint` infers typings.

I haven't had the time to l look into that, so I don't know if it's going to end up being hard to implement or not. The benefits can be quite big for a lot of rules from `perfectionist`, and for your new `de-morgan` plugin as well. It's something I can start looking into as well once the whole `sortBy: value` milestone is done. 🙂

azat-io:
The benefits can be quite big, but I'm afraid it could affect the performance.

hugop95:
@azat-io 

> The benefits can be quite big, but I'm afraid it could affect the performance.

It's always something we can put behind an option, similarly to `sort-imports` and `tsconfig` parsing!

azat-io:
Yes 👍

Issue #477, opened 2025-03-05
Bug: sort-modules stuck in loop with interdependent modules


Floriferous:
### Describe the bug

If you write this:

```js
class MyClass2 {
  static sisterClass = { sister: () => MyClass1 };
}

class MyClass1 {
  static sisterClass = { sister: () => MyClass2 };
}
```

You will get an eslint error, no matter how you order these 2 classes - impossible to fix: `Expected dependency "MyClass1" to come before "MyClass2".eslint[perfectionist/sort-modules](https://perfectionist.dev/rules/sort-modules)`

However, if you write this:

```js
class MyClass1 {
  static sisterClass = () => MyClass2;
}

class MyClass2 {
  static sisterClass = () => MyClass1;
}
```

There is no error. This seems like a strange inconsistency.

### Code example

See above.

### ESLint version

v8.57.0

### ESLint Plugin Perfectionist version

v4.9.0

### Validations

- [x] Read the [docs](https://perfectionist.dev).
- [x] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

hugop95:
Hi @Floriferous 

There is one confirmed bug and one where I'm a bit more puzzled:

> You will get an eslint error, no matter how you order these 2 classes - impossible to fix

This is a bug indeed: circular dependencies seem to cause this loop. It should let the user choose which element they want to put first.

> This seems like a strange inconsistency.

```ts
class MyClass1 {
  static sisterClass = () => MyClass2;
}

class MyClass2 {
  static sisterClass = () => MyClass1;
}
```

will not raise an error because we explicitly disable dependency searching in static methods and arrow function properties if:
- There is no `static` block in the class.
- AND there is no static property. 

You will see that 

```ts
class MyClass1 {
  static property = MyClass2.sisterClass(); // Or simply static property;
  static sisterClass = () => MyClass2;
}

class MyClass2 {
  static sisterClass = () => MyClass1;
}
```

raises the ESLint error (typescript does as well, but users might not use typescript).

However,

```ts
class MyClass2 {
  static sisterClass = { sister: () => MyClass1 };
}

class MyClass1 {
  static sisterClass = { sister: () => MyClass2 };
}
```

will not, because `sisterClass` is not a static arrow function property.

We could theoretically ignore `functions/arrow functions included in objects from static properties`, but:
- This becomes convoluted/quite complex.
[...]

PR #478, opened 2025-03-05
fix: prevent circular dependency loops

hugop95:
Fixes partially #477.

### Description

This PR prevents circular dependencies from raising unfixable errors:

```ts
class MyClass2 {
  static property = MyClass1.property;
}

class MyClass1 {
  static property = MyClass2.property;
}
```

Dependency order will not be enforced for elements being dependent in a circular way, meaning that the standard order will take over (it will expect `MyClass1` to come before `MyClass2` due to the alphabetical sort).

- [x] Bug fix


azat-io (approved):

azat-io:
Look good. Thank you!

Issue #479, opened 2025-03-05
Feature: Enforce partition group comments


nwidynski:
### What rule do you want to change?

sort-*

### Describe the problem

As showcased by the docs, groups are often partitioned by comments. Currently there is no way to enforce the presence of partition comments. 

Adding partition comments automatically via the fixer should be fairly trivial but I'm unsure whether this is out of scope for this plugin. Since groups are defined here, I would argue the plugin is uniquely positioned to solve it.

### Code example

Options `perfectionist/sort-imports`
```jsonc
{
  // Only applicable when partitionByComment is `true`
  groups: [
    'builtin',
    { newlineBetween: true, partitionComment: 'external', block: true },
    'external',
  ]
}
```

Bad:
```tsx
import { resolve } from "node:path";
import { foo } from "bar";
```

Good:
```tsx
import { resolve } from "node:path";

/* external */
import { foo } from "bar";
```

### Validations

- [x] Read the [docs](https://perfectionist.dev).
- [x] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

hugop95:
Hi @nwidynski ,

I understand the principle, although just to make sure that we understand things correctly: I would rename this option to `commentBetween`. `partitionByComment` separate blocks that both apply `groups` rules separately. What you are suggesting is a system similar to `newlinesBetween`.

I find the idea interesting, here are some points that should be explored a bit more:

## Comment removal

Could/should this system remove added comments if the configuration changes? How can we ensure that we are not removing a comment that the user meant to keep (read: a comment not added by the autofix for example).

## Comment matching

There should probably be a difference between comment matching (ensuring that a comment is there when the configuration says it should be here) and the comment to write with the autofix.

Example: Allow any comment matching `external.*` but autofix with `external` if not.

## Use cases

This is where I'm a bit skeptical at the moment: at first sight, this sounds like something that would be mostly used in `sort-imports` (which is what your example uses).

But is it really common for users to want to add a comment between some groups for **all** files? In your example, is adding `// external` above the first `external` import in all files really useful?

nwidynski:
@hugop95 Thanks for getting back so quickly!

> But is it really common for users to want to add a comment between some groups for all files? In your example, is adding // external above the first external import in all files really useful?

[...]

PR #482, opened 2025-03-08
feat(sort-exports): add `groups`-related options

hugop95:
- Request from https://github.com/azat-io/eslint-plugin-perfectionist/issues/479#issue-2898499262

> PS: We would love to see groups extended to sort-exports as well, especially once this feature is supported 👍

## Description

This PR adds the following options to `sort-exports`:
- `groups`
- `customGroups` (new API)
- `newlinesBetween`

It also deprecates `groupKind` for `groups`.

### `groups`

- Allowed selectors: `export`.
- Allowed modifiers: `value`, `type`.

Example:

- `value-export`.
- `type-export`.

### `customGroups`

The `customGroups` option supports `elementNamePattern`.

- [x] New Feature

### DIFF
--- a/rules/sort-exports.ts
+++ b/rules/sort-exports.ts
@@ -7,9 +7,17 @@ import {
   buildCustomGroupsArrayJsonSchema,
   partitionByCommentJsonSchema,
   partitionByNewLineJsonSchema,
+  newlinesBetweenJsonSchema,
   commonJsonSchemas,
   groupsJsonSchema,
 } from '../utils/common-json-schemas'
+import {
+  MISSED_SPACING_ERROR,
+  EXTRA_SPACING_ERROR,
+  GROUP_ORDER_ERROR,
+  ORDER_ERROR,
+} from '../utils/report-errors'
+import { validateNewlinesAndPartitionConfiguration } from '../utils/validate-newlines-and-partition-configuration'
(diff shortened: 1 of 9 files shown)


nwidynski:
@hugop95 Thank you so much for working on this 🚀 It looks great so far!

rules/sort-exports.ts
```diff
@@ -149,30 +215,40 @@ export default createEslintRule<Options, MESSAGE_ID>({
...
   meta: {
-    schema: [
-      {
+    schema: {
+      items: {
         properties: {
           ...commonJsonSchemas,
           groupKind: {
             enum: ['mixed', 'values-first', 'types-first'],
             description: 'Specifies top-level groups.',
```
azat-io, line 224: TypeScript ESLint adds information about deprecated options to the option description. Do you think we should do the same?

https://github.com/typescript-eslint/typescript-eslint/blob/84744742c79cf3c1a81cbb08fd2cbaf75b45afc7/packages/eslint-plugin/src/rules/max-params.ts#L57

hugop95: @azat-io Good idea, added in [`47231d9` (#482)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/482/commits/47231d9536b91ec1f201e3ec26f5e701884d5544) 🙂

azat-io (approved):

Issue #483, opened 2025-03-09
Bug: `fallbackSort` causes `sort-imports` failed to fix


ntnyq:
### Repro steps

- Run `pnpm run lint` will report errors.

- Run `pnpm run lint:fix` error still exists.

- Comment `fallbackSort` option in `eslint.config.mjs`, no error reported.

Besides, if there is no `// eslint-disable-next-line perfectionist/sort-imports`, it still works fine.

### Reproduction repo

https://github.com/ntnyq-dev/repro-perfectionist-sort-imports

### Config

```ts
// @ts-check

import pluginPerfectionist from 'eslint-plugin-perfectionist'
import { config, configs } from 'typescript-eslint'

export default config(
  {
    extends: configs.recommended,
  },
  {
    plugins: {
      perfectionist: pluginPerfectionist,
    },
    rules: {
      '@typescript-eslint/ban-ts-comment': 'off',
      'perfectionist/sort-imports': [
        'error',
        {
          fallbackSort: { order: 'asc', type: 'alphabetical' },
          order: 'asc',
          type: 'alphabetical',
          groups: [
            // Side effect style imports (e.g. 'normalize.css')
            'side-effect-style',

            // Styles (e.g. *.{css,scss,less})
            'style',
            /**
             * Imports that don’t fit into any other group
             */
            'unknown',
          ],
          internalPattern: ['^~/.+', '^@/.+', '^#.+'],
        },
      ],
    },
  },
)
```

### Code

```ts
import '@unocss/reset/tailwind.css'
import 'element-plus/theme-chalk/dark/css-vars.css'
import 'element-plus/theme-chalk/index.css'
import 'uno.css'
// COMMENT BELLOW is IMPORTANT
// eslint-disable-next-line perfectionist/sort-imports
import '@/styles/app.scss'

```

### ESLint version

v9.22.0

### ESLint Plugin Perfectionist version

v4.10.0

### Additional comments

Since option `order` and `type` is the same as  option `fallbackSort`.

I think there is no need to attempt to apply fallback sort.

Maybe some deep equal check.

### Validations

- [x] Read the [docs](https://perfectionist.dev).
- [x] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

hugop95:
@ntnyq Thank you for the details, a fix will come soon. It's actually unrelated to `fallbackSort`.

[...]

PR #485, opened 2025-03-09
fix(sort-imports): fix explicit `fallbackSort` overriding side-effects

hugop95:
Fixes https://github.com/azat-io/eslint-plugin-perfectionist/issues/483

### Description

Explicitely writing `fallbackSort` in `sort-imports` will override `sideEffects` which should stay unsorted.

### Bug severity

Medium-high:
- Related to a relatively new feature (fallback sorting)
- Requires writing an explicit `fallbackSort` option, meaning that `type` should be `line-length` logically.
- Side-effects import order can break runtime.

=> I think that a hotfix is justified.

- [x] Bug fix

### DIFF
--- a/test/rules/sort-imports.test.ts
+++ b/test/rules/sort-imports.test.ts
@@ -1829,6 +1829,31 @@ describe(ruleName, () => {
           valid: [],
         },
       )
+
+      ruleTester.run(
+        `${ruleName}(${type}): does not sort side-effects and side-effect-style even with fallback sort`,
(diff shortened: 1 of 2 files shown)


hugop95:
Hi @azat-io 👋

For this bug, I believe that deploying a hotfix is appropriate.

azat-io:
Okay, I'll create release. Thank you!

azat-io (approved):

PR #487, opened 2025-03-11
feat: export plugin meta data

ntnyq:
### Description

This PR adds meta data to plugin.

### Additional context

~Root `name` field is changed to `packageName` as ESLint recommended use npm package name as the meta name.~

Root `name` field is removed as it's deprecated and replaced by `meta.name`.

### Refs

- https://eslint.org/docs/latest/extend/plugins#meta-data-in-plugins

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -59,7 +59,6 @@ interface PluginConfig {
     version: string
     name: string
   }
-  name: string
 }
 
 interface BaseOptions {
@@ -96,7 +95,6 @@ let plugin = {
     version: packageVersion,
     name: packageName,
   },
-  name: packageName,
 } as unknown as ESLint.Plugin
 
 let getRules = (options: BaseOptions): Linter.RulesRecord =>
--- a/index.ts
+++ b/index.ts
@@ -1,5 +1,6 @@
 import type { ESLint, Linter, Rule } from 'eslint'
 
+import { version as packageVersion, name as packageName } from './package.json'
 import sortVariableDeclarations from './rules/sort-variable-declarations'
 import sortIntersectionTypes from './rules/sort-intersection-types'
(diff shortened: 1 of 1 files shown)


azat-io (approved):

azat-io:
Hi, I'm trying to figure out which version of ESLint `name` was deprecated in favor of `meta.name`. And apparently it happened a long time ago.

Perhaps we should remove the deprecated field?

ntnyq:
> Hi, I'm trying to figure out which version of ESLint `name` was deprecated in favor of `meta.name`. And apparently it happened a long time ago.
> 
> Perhaps we should remove the deprecated field?

ESLint introduced the plugin metadata from https://github.com/eslint/eslint/pull/16992 which has been released since https://github.com/eslint/eslint/releases/tag/v8.37.0.

As it says in https://eslint.org/docs/latest/extend/plugins#meta-data-in-plugins:

> While the meta object is the preferred way to provide the plugin name and version, this format is also acceptable and is provided for backward compatibility.

> As an alternative, you can also expose `name` and `version` properties at the root of your plugin.

Meta is just the preferred way.

azat-io:
We are now supporting ESLint with version 8.45.0:
https://github.com/azat-io/eslint-plugin-perfectionist/blob/main/package.json#L121

It seems that `meta.name` is not just a recommended method, but a replacement for the deprecated `name` property:
https://github.com/eslint/eslint/blob/main/lib/types/index.d.ts#L1468

I think the old style can safely be removed.

Perhaps a similar thing should be done here:
https://github.com/azat-io/eslint-plugin-de-morgan

Issue #488, opened 2025-03-12
Bug: Docs seem to be unclear


maartentibau:
### Describe the bug

Hi,

I have been trying for hours now, but I just can get it to work, so I hope I can reach out here, because really don't understand what I'm doing wrong, or maybe my idea of these rule-set is completely wrong.

So I'm trying to come up with a rule-set specifically for Angular. And I'm only using `perfectionist/sort-classes`

### Code example

In an ideal world my code would be sorted like this.

```ts
import { input, output } from '@angular/core';

type Foo<T> = {
  bar: T;
};

class FormatComponent<T> {
  a = input();
  b = input<T>();
  c = input<Foo<string>>({} as Foo<string>);

  d = output<void>();
  e = output<boolean>();
}
```

So in my config I try it like this:

```
{
  'perfectionist/sort-classes': [
    'error',
    {
      type: 'alphabetical',
      alphabet: `abcdefghijklmnopqrstuvwxyz`,
      newlinesBetween: 'always',
      groups: [
        'angular-inputs',
        'angular-outputs'
      ],
      customGroups: [
        {
          groupName: 'angular-inputs',
          selector: 'property',
          elementValuePattern: /input(?:<[^<>]+(?:|[^<>]+)*>)?(.*?)/g
        },
        {
          groupName: 'angular-outputs',
          selector: 'property',
          elementValuePattern: /output(?:<[^<>]+(?:|[^<>]+)*>)?(.*?)/g
        }
      ]
    }
  ],
  'perfectionist/sort-variable-declarations': 'off',
  'perfectionist/sort-intersection-types': 'off',
  'perfectionist/sort-heritage-clauses': 'off',
  'perfectionist/sort-array-includes': 'off',
  'perfectionist/sort-named-imports': 'off',
  'perfectionist/sort-named-exports': 'off',
  'perfectionist/sort-object-types': 'off',
  'perfectionist/sort-union-types': 'off',
  'perfectionist/sort-switch-case': 'off',
  'perfectionist/sort-interfaces': 'off',
  'perfectionist/sort-decorators': 'off',
  'perfectionist/sort-jsx-props': 'off',
  'perfectionist/sort-modules': 'off',
  'perfectionist/sort-imports': 'off',
  'perfectionist/sort-exports': 'off',
  'perfectionist/sort-objects': 'off',
  'perfectionist/sort-enums': 'off',
  'perfectionist/sort-sets': 'off',
[...]

PR #489, opened 2025-03-12
feat: handle `regex` objects until full deprecation

hugop95:
- Related issue: https://github.com/azat-io/eslint-plugin-perfectionist/issues/488
- Alternative PR: https://github.com/azat-io/eslint-plugin-perfectionist/pull/490

### Description

This PR adds a handler for actual RegExp objects. They should not be tolerated but there currently is an issue with the JSON schema.

On the next major version, an error can be thrown to prevent breaking some user configurations.

- [x] New Feature

### DIFF
--- a/utils/matches.ts
+++ b/utils/matches.ts
@@ -9,5 +9,10 @@ export let matches = (value: string, regexOption: RegexOption): boolean => {
     return new RegExp(regexOption).test(value)
   }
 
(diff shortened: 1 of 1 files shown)


hugop95:
Closed for https://github.com/azat-io/eslint-plugin-perfectionist/pull/490.

PR #490, opened 2025-03-12
fix: throw an error if RegExp expressions are entered

hugop95:
- Issue: https://github.com/azat-io/eslint-plugin-perfectionist/issues/488
- Alternative PR: https://github.com/azat-io/eslint-plugin-perfectionist/pull/489

## Description

This PR throws a custom error when a RegExp option is entered instead of a string.

Screenshot 2025-03-13 at 12 55 17

In the next major version, the error can be added on the JSON schema level.

- [x] Bug fix

### DIFF
--- a/test/utils/matches.test.ts
+++ b/test/utils/matches.test.ts
@@ -31,4 +31,11 @@ describe('matches', () => {
       ]),
     ).toBeTruthy()
   })
+
+  it('should throw an error if the regex option is not a string', () => {
+    expect(() => matches('foo', /foo/u as unknown as string)).toThrow(
+      'Invalid configuration: please enter your RegExp expressions as strings.\n' +
+        'For example, write ".*foo" instead of /.*foo/',
+    )
+  })
 })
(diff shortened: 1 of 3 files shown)


hugop95:
@azat-io Let me know if you prefer this fix over https://github.com/azat-io/eslint-plugin-perfectionist/pull/489 🙂

azat-io:
I don't have a clear opinion on that. What do you think is the best way to do it?

hugop95:
@azat-io I think that we can throw an error today, but customized (not through JSON schema yet):
- Configurations using RegExp expressions (not strings) simply do not work (as seen in https://github.com/azat-io/eslint-plugin-perfectionist/issues/488) but don't throw: I don't think that many users would keep them in their configs without realizing that this simply does nothing.
- Throwing "Please enter your RegExp expression as a string" will make it clear for users that still have invalid configurations.
- Allowing `RegExp` expressions would allow a bad habit that eventually would have to be fixed in the next major version by users.

I'll update this PR by:
- Commenting out the JSON schema check: this should be activated in next major version.
- Throw a custom error in the meantime.

azat-io (approved):

PR #500, opened 2025-03-29
refactor: add `computeGroup` function

hugop95:
### Description

TODO

- [x] Other


utils/use-groups.ts
```diff
@@ -9,46 +9,44 @@ interface UseGroupsValue {
...
 }
 
+/**
+ * Functions to set and define groups
+ * @deprecated for computeGroups function
+ * @param {object} params - The parameters for the operation
+ * @param {string[]} params.groups - groups array
+ * @returns {UseGroupsValue} functions to set and define groups
+ */
 export let useGroups = ({ groups }: UseGroupParameters): UseGroupsValue => {
```
azat-io, line 28: Can we remove it?

hugop95: @azat-io It's unfortunately not as easy as for other rules: `sort-imports` uses `useGroups` in a special way: it calls `setCustomGroups` twice with different groups (`options.customGroups.value` and `options.customGroups.type`)

https://github.com/azat-io/eslint-plugin-perfectionist/blob/48b45663e3ef6ed0ccfdc4b326821fe2769a147a/rules/sort-imports.ts#L357

https://github.com/azat-io/eslint-plugin-perfectionist/blob/48b45663e3ef6ed0ccfdc4b326821fe2769a147a/rules/sort-imports.ts#L410

I would like to refactor this rule a bit (https://github.com/azat-io/eslint-plugin-perfectionist/pull/504) by extracting and testing internal functions a bit more to simplify the main logic before finally getting rid of `useGroups` here.

azat-io (approved):

PR #502, opened 2025-03-30
fix: fix invalid `newlinesBetween` result enforced

### LINKED ISSUE #501: Bug: sort-imports: Using `newlinesBetween` more than once in `groups` unexpected behaviour
### Describe the bug

```js
'perfectionist/sort-imports: [
  'warn',
  {
    groups: [
      ['internal'],
      {newlinesBetween: 'never'},
      ['parent'],
      {newlinesBetween: 'never'},
      ['index', 'sibling'],
    ],
    internalPattern: ['^#'],
    newlinesBetween: 'always',
  },
],
```

With the config above, I would expect this to pass, but it doesn't:

```js
import something from '#src/something';
import another from './another'; // Warning: "Missed spacing between ..."
```

It appears that when no import occupies the `['parent']` group, the `['internal']` and `['index', 'sibling']` groups get split. See this example, which passes:

```js
import something from '#src/something';
import parent from '../parent';
import another from './another'; // No warnings!
```

I guess the correct behaviour would be: `{newlinesBetween: string}` should influence the gap between _all_ groups above and _all_ groups below, instead of just the adjacent two.

### Code example

See description

### ESLint version

9.23.0

4.10.1

### Additional comments

_No response_

### Validations

hugop95:
Fixes #501

## Description

When two elements belong to different groups, we check all the `newlinesBetween` options in-between those groups to deduce if we should enforce a newline behavior or not.

A case was missing: when all in-between elements have `newlinesBetween: never`, we should enforce no newline.

## Bug severity

Low:
- Does not affect build/runtime.
- Appears in quite specific cases.

- [x] Bug fix

### DIFF
--- a/test/utils/get-newlines-between-option.test.ts
+++ b/test/utils/get-newlines-between-option.test.ts
@@ -186,8 +186,9 @@ describe('get-newlines-between-option', () => {
       })
 
       describe('non-adjacent groups', () => {
-        it('should return `always` if the global option is `always`', () => {
-          for (let newlinesBetween of ['always', 'ignore', 'never'] as const) {
+        it.each(['always', 'ignore', 'never'] as const)(
+          'should return `always` if the global option is `always`',
+          newlinesBetween => {
             expect(
               getNewlinesBetweenOption(
                 buildParameters({
@@ -201,8 +202,8 @@ describe('get-newlines-between-option', () => {
                 }),
               ),
             ).toBe('always')
-          }
-        })
+          },
+        )
 
         it('should return `always` if `always` exists between the groups', () => {
           expect(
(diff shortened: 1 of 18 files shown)


test/utils/get-newlines-between-option.test.ts
```diff
@@ -7,46 +7,45 @@ import { getNewlinesBetweenOption } from '../../utils/get-newlines-between-optio
...
-      expect(
-        getNewlinesBetweenOption(
-          buildParameters({
-            newlinesBetween: 'ignore',
-            customGroups: {},
-          }),
-        ),
-      ).toBe('ignore')
-    })
+    it.each(['always', 'ignore', 'never'] as const)(
```
hugop95, line 10: ℹ️ Hide whitespace changes for this commit.

test/utils/get-newlines-between-option.test.ts
```diff
@@ -7,46 +7,45 @@ import { getNewlinesBetweenOption } from '../../utils/get-newlines-between-optio
...
+          getNewlinesBetweenOption(
+            buildParameters({
+              newlinesBetween,
+            }),
+          ),
+        ).toBe(newlinesBetween)
+      },
+    )
 
-    it('should return "ignore" if "newlinesBetween" is "ignore"', () => {
```
hugop95, line 31: This test was merged with the one below:

`should return "%s" if "newlinesBetween" is "%s"`

azat-io: 👍

azat-io (approved):

azat-io:
Thanks!

Issue #494, opened 2025-03-19
Question: Sorting named and default imports within the group


tedlatentai:
### Describe the bug

Hi there.

Is there any way to configure this library for more granular sorting of imports? I'd like to achieve separation between default and named imports within groups. Right now, these imports are just sorted all mixed together, and honestly, it doesn't look great in my opinion.

I tried looking for something here, in the docs, and online, but came up empty.

### Code example

plugin config, simplified

```json
'perfectionist/sort-imports': [
        'error',
        {
          groups: [
            'builtin',
            'external',
            'internal',
          ],
          type: 'natural',
        },
```

internal group, current sorting behaviour

```js
import { svelteTesting } from '@testing-library/svelte/vite';
import Icons from 'unplugin-icons/vite';
import { defineConfig } from 'vite';
```

desired order

```js
import Icons from 'unplugin-icons/vite';
import { svelteTesting } from '@testing-library/svelte/vite';
import { defineConfig } from 'vite';
```

### ESLint version

v9.22.0

### ESLint Plugin Perfectionist version

v4.10.1

### Validations

- [x] Read the [docs](https://perfectionist.dev).
- [x] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

tedlatentai:
@hugop95 

Thanks for considering this! Looking forward to being able to use it as soon as it's available.

tedlatentai:
@hugop95 

Thanks for considering this!

PR #505, opened 2025-04-06
feat(sort-imports) add support for new `groups` and `customGroups` API

hugop95:
- Resolves https://github.com/azat-io/eslint-plugin-perfectionist/issues/509.
- Related to the following issues:
  - https://github.com/azat-io/eslint-plugin-perfectionist/issues/494.
  - https://github.com/azat-io/eslint-plugin-perfectionist/issues/510.

# Description

We are finally migrating `sort-imports` to the new `customGroups` API with selectors and modifiers. 
This PR only focuses on the migration to the new API. The old API is still supported.

Here is the list of selectors and modifiers:

## Modifiers

- `type`: matches `type` imports.

## Selectors

Ordered from most important to least important (it is the same order as the predefined groups today).

- `index-type` (ℹ️ deprecated for the `type` modifier with `index` selector)
- `sibling-type` (ℹ️ deprecated, same as above)
- `parent-type` (ℹ️ deprecated, same as above)
- `internal-type` (ℹ️ deprecated, same as above)
- `builtin-type` (ℹ️ deprecated, same as above)
- `external-type` (ℹ️ deprecated, same as above)
- `type` (**not** marked as deprecated, as it can be useful to quickly aggregate all types at the same place)
- `side-effect-style`
- `side-effect`
- `style`
- `index`
- `sibling`
- `parent`
- `internal`
- `builtin` 
- `external`

> [!IMPORTANT]
> The `object` group was never matched, so it has been deprecated (but is still allowed in `groups` for backward compatibility).
> [See the code](https://github.com/azat-io/eslint-plugin-perfectionist/blob/8108a6eae2b731cfdd0e1abfc8773908f4c89847/rules/sort-imports.ts) before the refactor to check that the `object` group was never defined. It can be re-added and implemented later if needed. 

## `customGroups` option

Supports `elementNamePattern`.

# Follow-up improvements

- Update the default options to use the new generated groups (https://github.com/azat-io/eslint-plugin-perfectionist/pull/511).
- Add additional selectors and modifiers (https://github.com/azat-io/eslint-plugin-perfectionist/pull/512)

- [x] New Feature


utils/validate-generated-groups-configuration.ts
```diff
@@ -54,11 +54,30 @@ let isPredefinedGroup = (
     return false
   }
   let twoWordsSelector = input.split('-').slice(-2).join('-')
-  let isTwoWordSelectorValid = allSelectors.includes(twoWordsSelector)
-  if (!allSelectors.includes(singleWordSelector) && !isTwoWordSelectorValid) {
+  let threeWordsSelector = input.split('-').slice(-3).join('-')
```
azat-io, line 57: Perhaps we can simplify the code of this function and make it more declarative?

```ts
let isPredefinedGroup = (
  allSelectors: string[],
  allModifiers: string[],
  input: string,
): boolean => {
  if (input === 'unknown') {
    return true
  }
  
  let parts = input.split('-')
  if (parts.length === 0) {
    return false
  }
  
  let possibleSelectors = [
    { selector: parts.slice(-3).join('-'), wordCount: 3 },
    { selector: parts.slice(-2).join('-'), wordCount: 2 },
    { selector: parts.slice(-1)[0], wordCount: 1 }
  ].filter(({ selector, wordCount }) => 
    allSelectors.includes(selector) && parts.length >= wordCount
  )
  
  if (possibleSelectors.length === 0) {
    return false
  }
  
  let { wordCount } = possibleSelectors[0]
  
  let modifiers = parts.slice(0, -wordCount)
  
  return (
    new Set(modifiers).size === modifiers.length &&
    modifiers.every(modifier => allModifiers.includes(modifier))
  )
}
```

azat-io (approved):
Hi @hugop95! Thanks for this PR that adds a new API for `sort-imports`.

This PR looks good. 👍

utils/validate-generated-groups-configuration.ts
hugop95: @azat-io Thank you, it's more readable indeed!

[`9fea396` (#505)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/505/commits/9fea39632cf187224dea7341c19783b164c15a42)

Issue #510, opened 2025-04-11
Feature: subpath imports


remcohaszing:
### What rule do you want to change?

`perfectionist/sort-imports`

### Describe the problem

Node.js supports [subpath imports](https://nodejs.org/api/packages.html#subpath-imports). There is no group for this in the `perfectionist/sort-imports` rule by default.

### Code example

```jsonc
// package.json
{
  "imports": {
    "#internal": {
      "browser": "./browser-internal.js"
      "default": "./default-internal.js"
    }
  }
}
```

```js
// Some JavaScript code
import {} from '#internal'
```

### Additional comments

Currently this can be solved by leveraging internal patterns or custom groups, but this seems like an oversight.

### Validations

- [x] Read the [docs](https://perfectionist.dev).
- [x] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

hugop95:
# Dev notes

Can be investigated and eventually added with the `groups` and `customGroups` migration (https://github.com/azat-io/eslint-plugin-perfectionist/pull/505)

PR #511, opened 2025-04-12
feat(sort-imports): update default groups to new API

hugop95:
- Follow-up of https://github.com/azat-io/eslint-plugin-perfectionist/pull/505.
- Related to the following issues:
  - https://github.com/azat-io/eslint-plugin-perfectionist/issues/494.
  - https://github.com/azat-io/eslint-plugin-perfectionist/issues/510.

# Description

This PR adds a `value` modifier alongside an `import` selector in order for the `groups` option not to rely on deprecated selectors.

## `import` selector

This selector has the lowest priority of all selectors and will match every import.

## `value` modifier

Similar to the `type` modifier. Any non-`type` import will have the `value` modifier, even `side-effect` imports.

## Tests impacted

No input/output impact. Only error messages had to be updated.

- [x] New Feature

### DIFF
--- a/docs/content/rules/sort-imports.mdx
+++ b/docs/content/rules/sort-imports.mdx
@@ -323,6 +323,7 @@ The list of selectors is sorted from most to least important:
 #### Modifiers
 
 - `'type'` — Typescript time imports.
+- `'value'` — Value imports.
 
 #### Important notes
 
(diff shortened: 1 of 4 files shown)


azat-io (approved):

azat-io:
Awesome!

PR #519, opened 2025-04-20
feat(sort-imports): add back `ts-equals` (`object` group) modifier

hugop95:
# Description

In [`62c5c62` (#505)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/505/commits/62c5c621162bc2102c9890d5f450594259eed33d), the `object` group/selector was deprecated as it was never matched.

The previous documentation had this description:

https://github.com/azat-io/eslint-plugin-perfectionist/blob/f7907794ea899e1deae3f75de77b965bb3d09d5a/docs/content/rules/sort-imports.mdx?plain=1#L309

This PR does two things:
- It adds a `ts-equals` modifier that matches those objects.
- It prevents non-external reference ts import-equals imports from matching `external`:
  - `import log = console.log` currently matches `external` (it was matching `unknown` before https://github.com/azat-io/eslint-plugin-perfectionist/pull/508), but this syntax can also match internal imports (see the second example below).

## Example

```ts
// `value-ts-equals-import`
import log = console.log
```

```ts
// `type-ts-equals-import`
import type ForbiddenError = ErrorsNamespace.ForbiddenError
```

```ts
// `value-ts-equals-external`
import someImport = require('some-external-module')
```

- [x] Bug fix
- [x] New Feature

### DIFF
--- a/rules/sort-imports.ts
+++ b/rules/sort-imports.ts
@@ -220,20 +220,22 @@ export default createEslintRule<Options, MESSAGE_ID>({
         })
       }
 
-      if (isStyleSideEffect) {
-        selectors.push('side-effect-style')
-      }
+      if (!isNonExternalReferenceTsImportEquals(node)) {
+        if (isStyleSideEffect) {
+          selectors.push('side-effect-style')
+        }
 
-      if (isSideEffect) {
-        selectors.push('side-effect')
-      }
+        if (isSideEffect) {
+          selectors.push('side-effect')
+        }
 
-      if (isStyleValue) {
-        selectors.push('style')
-      }
+        if (isStyleValue) {
+          selectors.push('style')
+        }
 
-      for (let selector of commonSelectors) {
-        selectors.push(selector)
+        for (let selector of commonSelectors) {
+          selectors.push(selector)
+        }
       }
       selectors.push('import')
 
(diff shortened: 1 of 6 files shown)


rules/sort-imports.ts
```diff
@@ -69,6 +69,7 @@ let defaultGroups = [
   'value-internal',
   ['type-parent', 'type-sibling', 'type-index'],
   ['value-parent', 'value-sibling', 'value-index'],
+  'ts-equals-import',
```
hugop95, line 72: Reviewers:

Adds it as the same place as the `object` group before:

https://github.com/azat-io/eslint-plugin-perfectionist/blob/f7907794ea899e1deae3f75de77b965bb3d09d5a/rules/sort-imports.ts#L65-L66

test/rules/sort-imports.test.ts
```diff
@@ -368,7 +368,7 @@ describe(ruleName, () => {
     })
 
     ruleTester.run(
-      `${ruleName}(${type}): supports typescript object-imports`,
+      `${ruleName}(${type}): supports typescript ts import-equals`,
```
hugop95, line 371: Reviewers:

Here is the equivalent test when the `group` object was present:

https://github.com/azat-io/eslint-plugin-perfectionist/blob/b07949b82326a756cbfc7c1db34bd72f9cfca84c/test/rules/sort-imports.test.ts#L370-L437

`console.log` is now placed back again at the bottom as before.

test/rules/sort-imports.test.ts
```diff
@@ -3650,6 +3650,7 @@ describe(ruleName, () => {
               ],
               output: dedent`
                 import aImport from "b";
+
```
hugop95, line 3648: Reviewers:

`import a = aImport.ai.a2` is not `external` anymore, so the two imports are not in the same group anymore.

azat-io (approved):

Issue #527, opened 2025-04-24
Bug: `sort-imports` performance degraded in `4.12.0`


TimShilov:
### Describe the bug

After upgrading to `4.12.0` (with no changes to the config) we have noticed a significant slow down of the `sort-imports` rule, to the point of being unreasonable.
The following patch releases up until `4.12.2` didn't help, so we had to downgrade to `4.11.0`, which solves the issue.

Following are the outputs of `TIMING=1 pnpm eslint --fix --quiet .` on different versions:
**On `4.11.0`**
```
Rule                                | Time (ms) | Relative
:-----------------------------------|----------:|--------:
n/no-extraneous-import              |  1499.392 |    28.5%
unicorn/no-unnecessary-polyfills    |   719.804 |    13.7%
perfectionist/sort-objects          |   493.767 |     9.4%
@typescript-eslint/no-unused-vars   |   231.960 |     4.4%
n/no-unsupported-features/es-syntax |   131.468 |     2.5%
perfectionist/sort-imports          |    87.728 |     1.7%
unicorn/no-thenable                 |    79.073 |     1.5%
n/no-extraneous-require             |    62.046 |     1.2%
n/no-unpublished-require            |    61.484 |     1.2%
n/no-missing-require                |    53.696 |     1.0%
```

**On `4.12.2`**
```
:-----------------------------------|-----------:|--------:
perfectionist/sort-imports          | 163354.154 |    96.4%
n/no-extraneous-import              |   1777.778 |     1.0%
unicorn/no-unnecessary-polyfills    |    855.536 |     0.5%
perfectionist/sort-objects          |    509.117 |     0.3%
@typescript-eslint/no-unused-vars   |    233.514 |     0.1%
n/no-unsupported-features/es-syntax |    125.231 |     0.1%
unicorn/no-thenable                 |     89.087 |     0.1%
unicorn/import-style                |     87.392 |     0.1%
n/no-extraneous-require             |     80.501 |     0.0%
perfectionist/sort-named-imports    |     72.124 |     0.0%
```

### Code example

Here's the config we use (it's not the full config, but should be enough):
```ts
import perfectionist from "eslint-plugin-perfectionist";

/** @type {import("@typescript-eslint/utils").TSESLint.FlatConfig.ConfigFile} */
export const perfectionistPlugin = [
    perfectionist.configs["recommended-natural"],
    {
        rules: {
            "perfectionist/sort-imports": [
                "error",
                {
[...]

PR #528, opened 2025-04-24
perf(sort-imports): use `validateGeneratedGroupsConfiguration`

hugop95:
- Fixes https://github.com/azat-io/eslint-plugin-perfectionist/issues/527.

### Description

Only affects configurations using the deprecated `customGroups` API.

There are 2 ways to validate groups currently:
- [`validateGeneratedGroupsConfiguration`](https://github.com/azat-io/eslint-plugin-perfectionist/blob/main/utils/validate-generated-groups-configuration.ts) for the new API.
- [`validateGroupsConfiguration`](https://github.com/azat-io/eslint-plugin-perfectionist/blob/main/utils/validate-groups-configuration.ts) for the old API.

`validateGroupsConfiguration` relies on an array of all allowed predefined groups to work and is not optimized for large allowed groups arrays (which is the case with many selectors and modifiers, as you can permute modifiers as you wish).

This PR makes `sort-imports` use `validateGeneratedGroupsConfiguration` in both cases to ensure optimal performance all the time.

### Tests

Tests made in local, linting is fast as before with this version.

#### Before

Screenshot 2025-04-24 at 14 45 16

#### After

Screenshot 2025-04-24 at 14 33 12

- [x] Bug fix

### DIFF
--- a/rules/sort-imports.ts
+++ b/rules/sort-imports.ts
@@ -35,7 +35,6 @@ import { validateSideEffectsConfiguration } from './sort-imports/validate-side-e
 import { validateCustomSortConfiguration } from '../utils/validate-custom-sort-configuration'
 import { getCustomGroupOverriddenOptions } from '../utils/get-custom-groups-compare-options'
 import { readClosestTsConfigByPath } from './sort-imports/read-closest-ts-config-by-path'
-import { validateGroupsConfiguration } from '../utils/validate-groups-configuration'
 import { getOptionsWithCleanGroups } from '../utils/get-options-with-clean-groups'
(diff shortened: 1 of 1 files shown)


hugop95:
@azat-io

This issue is serious enough to require another deployment I'm afraid. 

There are other performance improvements we can make, such as validating configurations only once if it has been validated before (when linting another file for example). Those other improvements can be worked on in a separate PR.

ShGKme:
Fixes the performance issue, doesn't break the order with custom groups, thanks!

azat-io (approved):

azat-io:
Thank you! I think this is a serious problem and the release needs to be made immediately.

PR #533, opened 2025-04-26
test: add performance benchmark tests

hugop95:
- Related to https://github.com/azat-io/eslint-plugin-perfectionist/issues/527.

### Description

This PR adds basic performance benchmark tests to help detect and prevent heavy performance regressions. 

- It adds a `RuleTesterWithPerformanceBenchmark` class extending `typescript-eslint`'s `RuleTester`to monitor performance and throw if needed.
  - It won't throw in the standard `unit` tests checks.
- It adds a new `npm` script: `test:performance` as well as a CI job `performance` (see below).
  - It essentially runs unit tests again without coverage. 
  - The benchmark values to reach are based on **CI runs**. Local tests will likely end up running much faster. 
  - This test is **NOT** mandatory: it can fail while still allowing PRs to be merged.
    - Some tests can take more time than others (for example, `tsconfig` tests in `sort-imports`). I've empirically set expected benchmarks to reach based on various CI runs, but those values may still need to be tweaked over time. It's preferable for unexpected benchmark failures (in situations where performance shouldn't be impacted) to not block PRs merging. 
    - The default maximum value to reach for a test is `250ms` in the CI for most rules, with some exceptions for `sort-classes` (`300ms`) and `sort-imports` tests using `tsconfig`. 

image

[Example of failed `unit` CI check (coverage failed) and successful `performance` CI check.](https://github.com/hugop95/eslint-plugin-perfectionist/actions/runs/14685281175/job/41213081673?pr=12)

[Here is what the performance benchmark tests would have thrown before the recent `sort-imports` performance fix](https://github.com/azat-io/eslint-plugin-perfectionist/actions/runs/14685475074/job/41213508152?pr=533)

- [x] New Feature


test/rules/sort-jsx-props.test.ts
```diff
@@ -19,19 +19,24 @@ describe(ruleName, () => {
...
   RuleTester.it = it
 
-  let ruleTester = new RuleTesterWithPerformanceBenchmark({
-    languageOptions: {
-      parserOptions: {
-        tsconfigRootDir: path.join(__dirname, '../fixtures'),
-        extraFileExtensions: ['.svelte', '.astro', '.vue'],
-        ecmaFeatures: {
-          jsx: true,
+  let ruleTester = new RuleTesterWithPerformanceBenchmark(
```
hugop95, line 22: ℹ️ Reviewers:

Hide whitespace changes for this commit.

azat-io:
Thank you for your PR!
I'd like to do a little research on the topic of performance testing. I'll try to respond in the next few days.

azat-io:
Sorry for the long response.

I am currently doing some research on this issue. So far, I'm leaning towards the following:

- I don't really want to mix unit tests with performance tests
- I'd like to see more information about performance (number of operations per second, max and min values, etc).
- I think performance tests should be run a large number of times in order to get more accurate results
- I would like it to be a bot in Github Actions that would report performance information directly in the pull request as a comment
- I think it would be useful for us to know the exact numbers and what changes happened, not just information about whether we made it in any number of seconds, this would be useful for refactoring purposes

I've currently started developing a tool to run such tests. Development is still in progress:
https://github.com/azat-io/eslint-rule-benchmark

What do you think about it?

PR #540, opened 2025-05-03
fix: ignore Shebang comments

### LINKED ISSUE #539: Bug: Shebang comments are not ignored
### Describe the bug

Issue detected here: https://github.com/azat-io/eslint-plugin-perfectionist/issues/479#issuecomment-2841808874

Shebang comments are not ignored and will be moved around, resulting in code that does not compile. See code example.

### Code example

Code input (alphabetical sort):

```ts
#!/usr/bin/node
import { b } from "b";
import { a } from "a";
```

💥 Result:

```ts
import { b } from "b";
#!/usr/bin/node
import { a } from "a";
```

### ESLint version

8.57.0

4.12.3

### Additional comments

_No response_

### Validations

hugop95:
Fixes #539.

### Description

This PR makes comment fixes ignore shebang comments.

- [x] Bug fix

### DIFF
--- a/utils/get-comments-before.ts
+++ b/utils/get-comments-before.ts
@@ -9,7 +9,7 @@ interface GetCommentsBeforeParameters {
 
 /**
  * Returns a list of comments before a given node, excluding ones that are
- * right after code. Includes comment blocks.
+ * right after code. Includes comment blocks, ignore shebang comments.
  * @param {object} params - Parameters object.
  * @param {TSESTree.Node} params.node - The node to get comments before.
  * @param {TSESLint.SourceCode} params.sourceCode - The source code object.
@@ -22,7 +22,7 @@ export let getCommentsBefore = ({
   sourceCode,
   node,
 }: GetCommentsBeforeParameters): TSESTree.Comment[] => {
-  let commentsBefore = getCommentsBeforeNodeOrToken(sourceCode, node)
+  let commentsBefore = getRelevantCommentsBeforeNodeOrToken(sourceCode, node)
   let tokenBeforeNode = sourceCode.getTokenBefore(node)
   if (
     commentsBefore.length > 0 ||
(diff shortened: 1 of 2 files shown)


utils/get-comments-before.ts
```diff
@@ -31,18 +31,24 @@ export let getCommentsBefore = ({
...
+      /**
+       * `getCommentsBefore` also returns comments that are right after code,
+       * filter those out
+       */
+      let tokenBeforeComment = source.getTokenBefore(comment)
+      return tokenBeforeComment?.loc.end.line !== comment.loc.end.line
+    })
+
+let isShebangComment = (comment: TSESTree.Comment): boolean =>
+  comment.type === ('Shebang' as unknown as 'Block')
```
hugop95, line 54: Typings don't accept the `Shebang` type.

azat-io (approved):

azat-io (changes requested):
I'm not sure it will work if we use this with the standard espress parser.

For some reason TypeScript ESLint parser and Espress parser have a difference:

TypeScript ESLint:

Screenshot 2025-05-04 at 1 00 59 PM

Espree:

Screenshot 2025-05-04 at 1 00 50 PM

Perhaps we should support both options (Shebang and Hashbang)? Or make some custom function to define Hashbang? Like this:

https://github.com/eslint-community/eslint-plugin-n/blob/master/lib/rules/hashbang.js#L35-L47

hugop95:
@azat-io 

Thanks for pointing this out, I've added support for `Hashbang` in [`7fe24c1` (#540)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/540/commits/7fe24c13c0110ba9833866840eb04e01aad65395).

I've tried adding tests to check this but was unable to: after installing `"espree": "^10.3.0"`, I tried adding a failing test, but that test passes: the comment gets detected as a `Shebang` comment instead of `Hashbang`.

```ts
{
  languageOptions: {
    parserOptions: {
      sourceType: 'module',
      ecmaVersion: 2023,
      comment: true,
    },
    parser: espree,
  },
  errors: [
    {
      data: {
        right: 'a',
        left: 'b',
      },
      messageId: 'unexpectedImportsOrder',
    },
  ],
  output: dedent`
    #!/usr/bin/node
    import { a } from 'a'
    import { b } from 'b'
  `,
  code: dedent`
    #!/usr/bin/node // <= Should be detected as a `Hashbang` comment but is detected as a `Shebang` comment
    import { b } from 'b'
    import { a } from 'a'
  `,
  options: [{}],
},
```

Surpriginsly

```ts
    let code = dedent`
      #!/usr/bin/node
      import {a} from 'a'
    `
    espree.parse(code, {
      sourceType: 'module',
      ecmaVersion: 2023,
      comment: true,
    })
```

will parser the comment as `Hashbang` though, so I'm wondering if `eslint` doesn't do something internally regarding `Hashbang` comments.

PR #542, opened 2025-05-10
feat(sort-imports): allow to find alternate `tsconfig` file

### LINKED ISSUE #536: Feature: allow to find alternate tsconfig file (i.e.: tsconfig.base.json when using nx workspace)

sort-imports

### Describe the problem

In the `sort-imports` rule, you can specify the `tsconfigRootDir` option but the rule actually looks only for the file `tsconfig.json`. 
It would be nice to also be able to search for alternate file name like `tsconfig.base.json`. 
Maybe this could be an additional property to the `sort-imports` rule.

### Code example

```js
export default [
  // ...
  {
    // ...
    'perfectionist/sort-imports': [
        'error',
        {
          groups: [ /* ... */ ],
          tsconfigRootDir: '.',
          tsconfigFileName: 'tsconfig.base.json', // optinal: default to tsconfig.json
          order: 'asc',
          type: 'natural',
          newlinesBetween: 'never',
        },
      ],
    // ...
  },
  // ...
];
```
or
```js
export default [
  // ...
  {
    // ...
    'perfectionist/sort-imports': [
        'error',
        {
          groups: [ /* ... */ ],
          tsconfigRootPath: './tsconfig.base.json', // replace tsconfigRootDir entirely
          order: 'asc',
          type: 'natural',
          newlinesBetween: 'never',
        },
      ],
    // ...
  },
  // ...
];
```

### Additional comments

_No response_

### Validations

hugop95:
- Resolves #536

### Description

This PR:
- Adds an option `tsconfig: { rootDir: string; filename?: string}`.
  - If `filename` is not entered, will search for the closest `tsconfig.json` file.
- Deprecates `tsconfigRootDir`.

- [x] New Feature

### DIFF
--- a/rules/sort-imports/read-closest-ts-config-by-path.ts
+++ b/rules/sort-imports/read-closest-ts-config-by-path.ts
@@ -59,7 +59,7 @@ export let readClosestTsConfigByPath = ({
   } while (directory.length > 1 && directory.length >= tsconfigRootDir.length)
 
   throw new Error(
-    `Couldn't find any tsconfig.json relative to '${filePath}' within '${tsconfigRootDir}'.`,
+    `Couldn't find any ${tsconfigFilename} relative to '${filePath}' within '${tsconfigRootDir}'.`,
   )
 }
 
--- a/rules/sort-imports/read-closest-ts-config-by-path.ts
+++ b/rules/sort-imports/read-closest-ts-config-by-path.ts
@@ -22,10 +22,12 @@ export let contentCacheByPath = new Map<
 >()
 
 export let readClosestTsConfigByPath = ({
+  tsconfigFilename,
   tsconfigRootDir,
   contextCwd,
   filePath,
 }: {
+  tsconfigFilename: string
   tsconfigRootDir: string
   contextCwd: string
   filePath: string
@@ -39,7 +41,7 @@ export let readClosestTsConfigByPath = ({
   let checkedDirectories = [directory]
 
   do {
-    let tsconfigPath = path.join(directory, 'tsconfig.json')
+    let tsconfigPath = path.join(directory, tsconfigFilename)
     let cachedDirectory = directoryCacheByPath.get(directory)
     if (!cachedDirectory && fs.existsSync(tsconfigPath)) {
       cachedDirectory = tsconfigPath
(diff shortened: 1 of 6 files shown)


test/rules/sort-imports.test.ts
```diff
@@ -7464,119 +7464,125 @@ describe(ruleName, () => {
...
-        `${ruleName}: marks internal imports as 'internal'`,
-        rule,
-        {
-          valid: [
-            {
-              options: [
-                {
-                  groups: ['internal', 'unknown'],
-                  tsconfigRootDir: '.',
+      describe('tsconfigRootDir option', () => {
```
hugop95, line 7467: Reviewers:

ℹ️ Hide whitespace changes for this commit.

rules/sort-imports/read-closest-ts-config-by-path.ts
```diff
@@ -49,22 +51,15 @@ export let readClosestTsConfigByPath = (
...
     checkedDirectories.push(directory)
-  } while (
-    directory.length > 1 &&
-    directory.length >= input.tsconfigRootDir.length
-  )
+  } while (directory.length > 1 && directory.length >= tsconfigRootDir.length)
 
   throw new Error(
-    `Couldn't find any tsconfig.json relative to '${input.filePath}' within '${input.tsconfigRootDir}'.`,
+    `Couldn't find any tsconfig.json relative to '${filePath}' within '${tsconfigRootDir}'.`,
```
azat-io, line 62: Perhaps we should remove the mention of the `tsconfig.json` file? The name of a TypeScript config file can be anything.

An example of a repository where different variants are used in the project root:
https://github.com/withastro/astro

image

hugop95: @azat-io Good point.

The entered filename will now be returned in the error: [`b5fa869` (#542)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/542/commits/b5fa869ce9ab2d46cd1849d31913035c62a1af56)

azat-io (approved):

azat-io:
Added in 54b8a42.
Released in [v4.14.0](https://github.com/azat-io/eslint-plugin-perfectionist/releases/tag/v4.14.0).

Issue #543, opened 2025-05-12
Feature: Add @/* default alias pattern to sort-imports


caribou-code:
### What rule do you want to change?

sort-imports

### Describe the problem

By default the sort-imports rule will treat alias paths with the pattern `['^~/.*']` as internal paths, however it seems the more common approach when creating alias paths is with the pattern `['^@/.*']`. If one is supported already, it would make sense to support both.

### Code example

Currently requires manually overriding like so:
```
'perfectionist/sort-imports': [
  'error',
  {
    internalPattern: ['^@/.*'],
    type: 'natural',
  },
],
```

### Validations

- [x] Read the [docs](https://perfectionist.dev).
- [x] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

azat-io:
Added in cbc91e0c61e0963111706b3f51b984c1682789d5.
Released in [v4.13.0](https://github.com/azat-io/eslint-plugin-perfectionist/releases/tag/v4.13.0).

PR #544, opened 2025-05-12
feat(sort-imports): use @/ as internal import pattern by default

azat-io:
### Description

#543

### Additional context

N/A.

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/docs/content/rules/sort-imports.mdx
+++ b/docs/content/rules/sort-imports.mdx
@@ -203,7 +203,7 @@ Specifies the sorting locales. Refer To [String.prototype.localeCompare() - loca
 
 ### internalPattern
 
-<sub>default: `['^~/.*']`</sub>
+<sub>default: `['^~/.+', '^@/.+']`</sub>
 
 Specifies a pattern for identifying internal imports. This is useful for distinguishing your own modules from external dependencies.
 
--- a/rules/sort-imports.ts
+++ b/rules/sort-imports.ts
@@ -87,8 +87,8 @@ export default createEslintRule<Options, MESSAGE_ID>({
     let userOptions = context.options.at(0)
     let options = getOptionsWithCleanGroups(
       complete(userOptions, settings, {
+        internalPattern: ['^~/.+', '^@/.+'],
         fallbackSort: { type: 'unsorted' },
-        internalPattern: ['^~/.+'],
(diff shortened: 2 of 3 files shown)


hugop95:
@azat-io I wonder if this would not cause issues for existing configs. Shouldn't this be kept for a major version upgrade?

azat-io:
@hugop95 I think using `@/*` for internal imports is pretty much standard practice. External imports usually look like `@some-word/*`.

docs/content/rules/sort-imports.mdx
```diff
@@ -203,7 +203,7 @@ Specifies the sorting locales. Refer To [String.prototype.localeCompare() - loca
 
 ### internalPattern
 
-<sub>default: `['^~/.*']`</sub>
+<sub>default: `['^~/.+', '^@/.+']`</sub>
```
hugop95, line 206: @azat-io [nit] [docs]

We could also update the `## Usage` section.

https://github.com/azat-io/eslint-plugin-perfectionist/blob/5b21319ab5a67929947c22af39eab9e748e6439b/docs/content/rules/sort-imports.mdx?plain=1#L614

https://github.com/azat-io/eslint-plugin-perfectionist/blob/5b21319ab5a67929947c22af39eab9e748e6439b/docs/content/rules/sort-imports.mdx?plain=1#L656

azat-io: Thanks! Done ✅️️

NMinhNguyen:
@azat-io Apologies for commenting on a merged PR, but would you be open to adding `#` imports to `internalPattern` by default? I can raise a PR if you give the go ahead. Or do you think it's not safe to assume that `#` is for internal?

azat-io:
@NMinhNguyen Thanks for the suggestion! `#` imports (Node.js subpath imports) are already handled by a dedicated `subpath` selector, so you can place them in any group you want. 

Adding them to `internalPattern` by default would mix two different concepts - internalPattern is for regex-based alias matching (`~/...`, `@/...`), while `subpath` is a first-class selector. 

If you want `#` imports grouped with internal ones, you can either add `'^#'` to `internalPattern` or use the `subpath` selector in your groups configuration.

PR #547, opened 2025-05-25
ci: integrate eslint rule benchmark

azat-io:
### Description

This is an experiment, we are implementing [ESLint Rule Benchmark](https://github.com/azat-io/eslint-rule-benchmark). We plan to study how it will work in practice, whether it will help in finding performance problems.

First of all, the goal is to understand what problems the benchmarking tool has.

### Additional context

The same PR in ESLint Plugin De Morgan: https://github.com/azat-io/eslint-plugin-de-morgan/pull/5

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other


azat-io:
Wow, what a huge report.

hugop95:
Hi @azat-io 👋

I'll think about the subject a bit more this week when I find time, but the main point of my initial PR was to detect big outlier cases (a bad optimization in a validation function for example).

It would be interesting to see what the report says for cases like:
- Using the `tsconfig`/`tsconfigRootDir` option in `sort-imports` as it's quite expensive performance wise.
- Using `customGroups: {}` before the optimization in https://github.com/azat-io/eslint-plugin-perfectionist/pull/528.

I think that we ideally should:
- Store the current benchmarks to reach on `main`, or set maximum thresholds not to go above.
- In the report, hide/collapse results that have no difference with `main` (with a `<details>` tag for example), and only highlight stronger variations (positive or negative).

Obviously, it's not easy, and I don't know if it's worth going this far, but that's what think would be perfect 😁

hugop95:
@azat-io 

I wonder if it would be possible to easily run performance checks for all unit test cases. The advantage is that:
- We wouldn't have to write performance-dedicated tests: writing a test means both checking functional validity and acceptable performance.
- We ensure maximum use-case coverage as well.

PR #552, opened 2025-06-04
fix(sort-imports): allow single imports to be linted for `commentAbove`

hugop95:
- Original issue: https://github.com/azat-io/eslint-plugin-perfectionist/issues/479
- Original comment: https://github.com/azat-io/eslint-plugin-perfectionist/issues/479#issuecomment-2941422278

## Description

For small optimization purposes, node arrays that don't have a size of 2 are not considered, which makes sense sorting-wise.

The `commentAbove` option should work for files containing a single import, however.

### Changes

- Allows `pairwise` to iterate on single-element arrays.

## Bug severity

Very low.

- [x] Bug fix

### DIFF
--- a/utils/pairwise.ts
+++ b/utils/pairwise.ts
@@ -1,10 +1,8 @@
-import { isSortable } from './is-sortable'
-
 export let pairwise = <T>(
   nodes: T[],
   callback: (left: null | T, right: T) => void,
 ): void => {
-  if (!isSortable(nodes)) {
+  if (nodes.length === 0) {
     return
   }
 
--- a/test/rules/sort-exports.test.ts
+++ b/test/rules/sort-exports.test.ts
@@ -1697,6 +1697,40 @@ describe(ruleName, () => {
       })
 
       ruleTester.run(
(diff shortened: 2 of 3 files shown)


utils/pairwise.ts
```diff
@@ -1,10 +1,8 @@
-import { isSortable } from './is-sortable'
-
 export let pairwise = <T>(
   nodes: T[],
   callback: (left: null | T, right: T) => void,
 ): void => {
-  if (!isSortable(nodes)) {
+  if (nodes.length === 0) {
```
azat-io, line 5: What is the difference between `!isSortable(nodes)` and `nodes.length === 0`?

hugop95: @azat-io 

https://github.com/azat-io/eslint-plugin-perfectionist/blob/4583226f275073de83e298c9db4c92720e631d82/utils/is-sortable.ts#L1-L2

`isSortable` return true only if the array has at least 2 elements.

azat-io (approved):

PR #553, opened 2025-06-06
feat: allow `newlinesBetween` to support `number`

hugop95:
- Related comment: https://github.com/azat-io/eslint-plugin-perfectionist/issues/479#issuecomment-2941650924

> [!IMPORTANT]
> Most line changes are **whitespace changes** in tests in commit [`9fcdf67` (#553)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/553/commits/9fcdf67dc257a8fe7234d9ad986d63b6ffcb39a7). Hide whitespace changes in this commit to view relevant changes.

## Description

It can be useful to enforce more than 1 newline in order to place comments in-between imports, separated by a newline.

*Example*:
```ts
import { something } from "somewhere"

// ---

import { anotherThing } from "somewhere-else" 
```

`newlinesBetween` is naturally fit to support `number`:
- `ignore` is unchanged.
- `never` ➡️ 0.
- `always` ➡️ 1.
- Any other number ➡️ Enforce that number of newlines between group.

The same applies for `newlinesInside`.

## Changes

- Allows `newlinesBetween` to support `number`. It still supports `always` and `never` as before.
- Allows `newlinesInside` to support `number`. It still supports `always` and `never` as before.

The documentation types `newlinesBetween` and `newlinesInside` as `number | 'ignore'` as `always` and `never` are redundant with `1` and `0`.

Support for `always` and `never` can be dropped in the next major release.

- [x] New Feature


test/utils/get-newlines-between-option.test.ts
```diff
@@ -6,9 +6,12 @@ import type { NewlinesBetweenOption } from '../../types/common-options'
 import { getNewlinesBetweenOption } from '../../utils/get-newlines-between-option'
 import { UnreachableCaseError } from '../../utils/unreachable-case-error'
 
+const NEVER_OPTIONS = [0, 'never'] as const
```
hugop95, line 9: ℹ️ Reviewers: hide whitespace changes for this commit.

test/rules/sort-array-includes.test.ts
```diff
@@ -1524,184 +1524,188 @@ describe(ruleName, () => {
...
-                    right: 'y',
-                    left: 'a',
-                  },
-                  messageId: 'extraSpacingBetweenArrayIncludesMembers',
-                },
-                {
-                  data: {
-                    right: 'b',
-                    left: 'z',
+      for (let newlinesBetween of ['never', 0] as const) {
```
hugop95, line 1527: ℹ️ Reviewers: Hide whitespace changes for this commit !

azat-io (approved):

PR #556, opened 2025-06-10
fix: prevent newlines fixes between different partitions

hugop95:
- Related comment: https://github.com/azat-io/eslint-plugin-perfectionist/issues/479#issuecomment-2953160238
- Follow-up of https://github.com/azat-io/eslint-plugin-perfectionist/pull/554

## Description

Users can separate partitions with comments using `partitionByComment`. However, newlines errors and fixes will be raised/applied between two consecutive elements of different partitions.
https://github.com/azat-io/eslint-plugin-perfectionist/pull/554 reduced the occurrences of this happening, but it may still happen in some configurations, such as when partition A depends on the first groups of the `groups` option, and partition B depends on last groups of the `groups` options.

## Code example

```ts
    groups: [
      "external",
      'sibling'
    ],
    newlinesBetween: "never",
    partitionByComment: true,
    type: "alphabetical"
}]
```

```ts
import a from 'a';

// Siblings

import { c } from './c';
import { b } from './b';
```

Gets linted into ➡️

```ts
import a from 'a';
// Siblings
import { b } from './b';
import { c } from './c';
```
❌ *Newline enforced between partitions above and before // Siblings*

- [x] Bug fix

### DIFF
--- a/rules/sort-array-includes.ts
+++ b/rules/sort-array-includes.ts
@@ -229,7 +229,7 @@ export let sortArray = <MessageIds extends string>({
         options,
       })
 
-      let sortingNode: SortArrayIncludesSortingNode = {
+      let sortingNode: Omit<SortArrayIncludesSortingNode, 'partitionId'> = {
         isEslintDisabled: isNodeEslintDisabled(element, eslintDisabledLines),
         size: rangeToDiff(element, sourceCode),
         node: element,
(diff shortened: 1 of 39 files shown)


azat-io (approved):

azat-io:
Nice, thank you!

PR #565, opened 2025-08-09
refactor: improve jsdoc

azat-io:
### Description

Did some refactoring to add JSDoc.

### Additional context

N/A.

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other


hugop95:
@azat-io 

I have no strong opinion on extensive internal JSDoc (I'm more in the [code as documentation](https://martinfowler.com/bliki/CodeAsDocumentation.html) school personally), but could we merge this on `main` ? As it's impactless for users, it will make it more easily maintainable if we have to later make a change on `main` and `next` at the same time.

azat-io:
I think that in 2025, this will make sense and will be useful not only for developers but also for LLM.

PR #559, opened 2025-06-29
feat: annotation-based config support

hugop95:
- Related comment https://github.com/azat-io/eslint-plugin-perfectionist/issues/557#issuecomment-2967372297
- Resolves #400

> [!IMPORTANT]
> Available from `v5.0.0`.

## Description

Sorry, it's a bit of a heavy PR, but it's been open for too long (before my holidays).

Users have requested multiple times the possibility to disable sorting if a comment is present.

This PR adds a `useConfigurationIf.declarationCommentMatchesPattern` option to the following rules:
- `sort-objects`.
- `sort-interfaces`.
- `sort-object-types`.

### Example usage

```ts
// Ignore me
const obj = {
  b,
  c,
  a,
}
```

```ts
{
  useConfigurationIf: {
    declarationCommentMatchesPattern: '^Ignore me$',
  },
  type: 'unsorted',
},
```

### Note

`declarationCommentMatchesPattern` currently only affects comment above the immediate's object/type parent.

Nested matching is currently not supported, but there is plan to improve that. See https://github.com/azat-io/eslint-plugin-perfectionist/issues/571#issuecomment-3196206473.

## Commit split

- Commits **1-4**: Side refactoring.
- Commits **5-6**: `sort-object-types` and `sort-interfaces` implementation.
- Commits **7-8**: `sort-objects` implementation.
- Commits **9-11**: Readability refactor.

## What is the purpose of this pull request?

- [x] New Feature

PR #569, opened 2025-08-10
release v5.0.0

### LINKED ISSUE #566: Release: v5.0.0
### Describe the rule

...

### Code example

...

### Additional comments

_No response_

### Validations

azat-io:
Close #566.


package.json
```diff
@@ -57,16 +57,16 @@
     "test:unit": "vitest --run --coverage"
   },
   "dependencies": {
-    "@typescript-eslint/types": "^8.38.0",
-    "@typescript-eslint/utils": "^8.38.0",
+    "@typescript-eslint/types": "^8.39.0",
+    "@typescript-eslint/utils": "^8.39.0",
```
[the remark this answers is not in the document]
azat-io: Fixed.

index.ts
```diff
@@ -68,33 +72,37 @@ interface BaseOptions {
...
+  'sort-jsx-props': sortJsxProps,
+  'sort-modules': sortModules,
+  'sort-classes': sortClasses,
+  'sort-imports': sortImports,
+  'sort-exports': sortExports,
+  'sort-objects': sortObjects,
+  'sort-enums': sortEnums,
+  'sort-sets': sortSets,
+  'sort-maps': sortMaps,
+}
```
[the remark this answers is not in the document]
azat-io: Don't worry about const vs. let.

azat-io:
@coderabbitai pause

package.json
```diff
@@ -11,7 +11,7 @@
   "repository": "azat-io/eslint-plugin-perfectionist",
   "license": "MIT",
   "author": "Azat S. <to@azat.io>",
-  "type": "commonjs",
+  "type": "module",
```
hugop95, line 14: Hi @azat-io 

Don't you think we should revert this commit for full ESLint v8 support ?

hugop95:
@azat-io 

For info, I still would like to improve the `sort-objects` and `sort-object-types`/`sort-interfaces`'s `useConfigurationIf.declarationMatchesPattern` ([replacement of `ignorePattern`](https://github.com/azat-io/eslint-plugin-perfectionist/pull/569/commits/61cabe05db666037536ee3c880686a5b81e8f202)), as it currently doesn't fulfill all use cases (cf https://github.com/azat-io/eslint-plugin-perfectionist/issues/571).

If you still want to release early, I think it's best to re-introduce `ignorePattern` until the improvements are made.

azat-io:
@hugop95

Thanks for flagging this! I'm happy to wait if you think the improvements are close, or we can go with your suggestion to re-introduce `ignorePattern` temporarily - whatever works best for you. Do you have a rough idea of when you might get to it?

package.json
azat-io: It seems that we have no problems supporting the old version, taking into account the definition of the supported Node.js version.

hugop95: @azat-io 

There is something I must be missing then, when I install `next`, I get

```ts
import { name, version } from './package.json.js'
^^^^^^

SyntaxError: Cannot use import statement outside a module
```

with a legacy config and node v22.19.0 (it works if I revert https://github.com/azat-io/eslint-plugin-perfectionist/pull/569/commits/7f2436083c30831767392c49fbf03d8c12c43d01)

---

Regardless of the actual issue though, I think it's important because it means that users that go from V4 to V5 will potentially have a big breaking change that is similar to the ESlint V8 -> V9 migration.

hugop95:
@azat-io 

> Do you have a rough idea of when you might get to it?

I would say that by the end of next week, I should either have a clear plan and implementation, or scrap the idea for now.

azat-io:
@hugop95 Sounds good, that works for me! Take your time - I'll hold off on the release until then. Let me know how it goes 👍

hugop95:
@azat-io 

I've found a fix for `sort-object`'s `ignorePattern` depreciation (see https://github.com/azat-io/eslint-plugin-perfectionist/pull/632#issuecomment-3623149521). Another one for `sort-object-types`/`sort-interfaces` should come soon.

There is one last thing I would like to fix/push for a later feature: the new `declarationCommentMatchesPattern` feature.
- PR: https://github.com/azat-io/eslint-plugin-perfectionist/pull/559

It's a V5 feature, so if really needed, we can always remove documentation for it, release the V5, then fix it as needed before officially releasing it in a minor version. I would just like to avoid releasing it in its current state in case we have to make breaking changes there.

The main issue is that it works for simple cases, but not complex ones. Example:

```ts
// ignore me
type T = {
...
}
```

With 

```ts
{
  useConfigurationIf: {
    declarationCommentMatchesPattern: "ignore me"
  }
}
```

Will work.

But not for

```ts
// ignore me
type T = A & {
...
}
```

I might have a fix by the end of the week hopefully. If not, I would like to remove the documentation for that part for the release.

azat-io:
@hugop95 

It seems that the problem has now been solved.

Please let me know if you have any further plans regarding this release. Or can I proceed to global review and testing and focus on merging and releasing v5?

hugop95:
@azat-io 

There is no fix for the [`declarationCommentMatchesPattern`](https://github.com/azat-io/eslint-plugin-perfectionist/pull/569#issuecomment-3623167228) part yet I'm afraid. I also don't have a clear fix in mind at the moment, I'll remove the documentation for this feature in the meantime so it doesn't affect the release.

---

There is one more feature that should at least be done before the release: https://github.com/azat-io/eslint-plugin-perfectionist/pull/634 but for `sort-object-types`. It's very similar to what was done in that PR, so it should be quick.

Optionally, there is [this ticket](https://github.com/azat-io/eslint-plugin-perfectionist/issues/631) that I will also investigate more afterward, but I have no estimate on how long it will take or even if it will be achievable, so I suggest not waiting for this for the next release.

azat-io:
@hugop95 Do you have any other plans for the release? I'm thinking about releasing it on Tuesday or Wednesday.

hugop95:
@azat-io I don't have anything urgent to add to the release 👍

PR #573, opened 2025-08-16
feat: improve test typings

hugop95:
### Description

TODO

- [x] Other


test/rules/create-typed-rule-tester.ts
```diff
@@ -6,13 +6,15 @@ import type { RuleModule } from '@typescript-eslint/utils/ts-eslint'
 
 import { createRuleTester } from 'eslint-vitest-rule-tester'
 
+import type { ReadonlyDeep } from '../../types/readonly-deep'
```
hugop95, line 9: Reviewers:

This is needed in order to support `as const` assertions in next commits. `as const` transforms the type to a nested `readonly` one.

hugop95:
Hi @azat-io 👋

Can you let me know what you think of this proposal regarding test typings?

Thanks! 🙏

azat-io:
Thanks for the PR! Technically, everything looks good, and the typed wrapper approach seems well thought out.

I have doubts about the proportionality of the solution. The changes add an extra layer of abstraction to solve a problem that manifests itself in tests. And I'm not sure it's worth the effort.

As a developer, I will now want to study another file when writing tests.

hugop95:
@azat-io Thanks, I understand. Let's close this here in that case. 👍

PR #584, opened 2025-09-13
feat: add sort-import-attributes rule

azat-io:
### Description

Adds new rule `sort-import-attributes`.

---

- [ ] Bug fix
- [x] New Feature
- [x] Documentation update
- [ ] Other

### DIFF
--- a/rules/sort-import-attributes.ts
+++ b/rules/sort-import-attributes.ts
@@ -1,4 +1,5 @@
 import type { TSESTree } from '@typescript-eslint/types'
+import type { TSESLint } from '@typescript-eslint/utils'
 
 import type { Options } from './sort-import-attributes/types'
 import type { SortingNode } from '../types/sorting-node'
@@ -17,6 +18,7 @@ import {
   GROUP_ORDER_ERROR,
   ORDER_ERROR,
 } from '../utils/report-errors'
+import { validateNewlinesAndPartitionConfiguration } from '../utils/validate-newlines-and-partition-configuration'
 import { buildGetCustomGroupOverriddenOptionsFunction } from '../utils/get-custom-groups-compare-options'
 import { validateGeneratedGroupsConfiguration } from '../utils/validate-generated-groups-configuration'
 import { validateCustomSortConfiguration } from '../utils/validate-custom-sort-configuration'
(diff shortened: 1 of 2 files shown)


docs/content/rules/sort-import-attributes.mdx
```diff
@@ -0,0 +1,275 @@
...
+groups: ['type-attrs', 'unknown']
+customGroups: [
+  { groupName: 'type-attrs', elementNamePattern: '^type' },
+]
+newlinesBetween: 'always'
+```
+
+### customGroups
+
+Define custom attribute groups with optional per-group sort overrides.
```
hugop95, line 178: @azat-io 

I think that we should tell what is the default value for this option (`[]`).

rules/sort-import-attributes.ts
```diff
@@ -0,0 +1,206 @@
...
+
+      let { sourceCode, id } = context
+      let settings = getSettings(context.settings)
+      let options = complete(context.options.at(0), settings, defaultOptions)
+      validateCustomSortConfiguration(options)
+      validateGeneratedGroupsConfiguration({
+        selectors: [],
+        modifiers: [],
+        options,
+      })
```
hugop95, line 76: @azat-io 

We can also `validateNewlinesAndPartitionConfiguration` (checks that `partitionByNewline` and `newlinesBetween`-options don't conflict together).

rules/sort-import-attributes.ts
```diff
@@ -0,0 +1,206 @@
...
+  defaultOptions: [defaultOptions],
+  name: 'sort-import-attributes',
+})
+
+function getAttributeName(attribute: TSESTree.ImportAttribute): string {
+  let { key } = attribute
+  if (key.type === 'Identifier') {
+    return key.name
+  }
+  return key.value!.toString()
```
hugop95: @azat-io 

I couldn't find a case where `key.value` can actually be `null` (like the AST says), but I think it would be safer to write

```ts
key.value?.toString() ?? "attribute"
```

so the rule doesn't crash in case someone finds a case where it happens.

hugop95 (approved):
@azat-io 💯

rules/sort-import-attributes.ts
azat-io: Done

rules/sort-import-attributes.ts
azat-io: Done

hugop95: @azat-io Thanks for the changes !

Issue #564, opened 2025-08-05
Feature: sorting enums should take into account if it's a numeric based enum


niemyjski:
### What rule do you want to change?

https://perfectionist.dev/rules/sort-enums

### Describe the problem

Enums get extended over time and they are typically sorted by numeric if they are not string based. For string based, I always want to sort by name.

### Code example

```
export enum SuspensionCode {
    Billing = 0,
    Overage = 1,
    Abuse = 2,
}
```

```
export enum SuspensionCode {
    Billing = 0,
    Overage,
    Abuse,
}
```

### Additional comments

I think this might be solved like this:

```
 'error',
      {
        type: 'unsorted', // disables primary sorting
        sortByValue: true, // sorts using values (will be numeric if possible)
        forceNumericSort: false, // allows fallback when values aren't numeric
        fallbackSort: {
          type: 'alphabetical',
          order: 'asc',
        },
      },
```

Just kinda feels like this should be the default. I've never seen a numeric enum sorted alpha by default in any language or code project I've used.

### Validations

- [x] Read the [docs](https://perfectionist.dev).
- [x] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

hugop95:
Hi @niemyjski 

If I understand correctly, you suggest that:
- Numeric only enums are sorted by numeric value.
- Non-numeric only enums are sorted in the alphabetical order.

So something like this ?

```ts
export enum SuspensionCode {
  Billing = 0,
  Overage = 1,
  Abuse = 2,
}

export enum SuspensionCode {
  Abuse = "Abuse",
  Billing = "Billing",
  Overage = "Overage",
}
```

If so, `forceNumericSort: true` is enough to get this order.
Having it enabled by default is interesting, but that should come within a major version update in order not to have breaking chances.

---

N.B:

```ts
export enum SuspensionCode {
    Billing = 0,
    Overage,
    Abuse,
}
```

will not be sorted (this is the expected behavior), as the non-defined enum values are given by the their order in the enum.

niemyjski:
That is correct! I’ll try that setting and will keep an eye out for any major

azat-io:
We are now planning a major release:

https://github.com/azat-io/eslint-plugin-perfectionist/issues/566

azat-io:
[...]

PR #586, opened 2025-09-16
feat(sort-enums)!: replace `forceNumericSort` and update default `sortByValue` option

hugop95:
- Resolves https://github.com/azat-io/eslint-plugin-perfectionist/issues/564

# Description

This PR addresses 3 points of the `sort-enums` rule.

## `forceNumericSort` replacement

There are 2 options used to control sorting by value today:
- `sortByValue: boolean`.
  - If `true`, will always sort by value.
- `forceNumericSort: boolean`.
  - If `true`, will sort by value if the enum is numeric.   

Those two options can be simply merged into:
`sortByValue: boolean | "ifNumericEnum"`.

## Incorrect `NaN` detection

When detecting if an enum value is a number, we ignore the fact that expressions such as `NaN | 42` result in `42`.

The case happens in enums such as

```ts
enum Enum {
  A = 1,
  B = 2,
  C = A << 2 // A is considered as NaN, so the result will be considered as 0, but it should be 4 in reality if we could parse that A is 1
}
```

## Sort numeric enums with `natural` sort by default

- Request from https://github.com/azat-io/eslint-plugin-perfectionist/issues/564

- [x] New Feature

### DIFF
--- a/docs/content/rules/sort-enums.mdx
+++ b/docs/content/rules/sort-enums.mdx
@@ -176,7 +176,7 @@ Controls whether sorting should be done using the enum's values or names.
 
 - `true` — Use enum values.
 - `false` — Use enum names.
-- `ifNumericEnum` — Use enum values only if the enum is numeric.
+- `'ifNumericEnum'` — Use enum values only if the enum is numeric.
 
 When this setting is `true`, numeric enums will have their values sorted numerically regardless of the `type` setting.
 
--- a/docs/content/rules/sort-enums.mdx
+++ b/docs/content/rules/sort-enums.mdx
@@ -176,17 +176,9 @@ Controls whether sorting should be done using the enum's values or names.
 
 - `true` — Use enum values.
 - `false` — Use enum names.
+- `ifNumericEnum` — Use enum values only if the enum is numeric.
 
 When this setting is `true`, numeric enums will have their values sorted numerically regardless of the `type` setting.
-
-### forceNumericSort
-
-<sub>default: `false`</sub>
-
-Controls whether numeric enums should always be sorted numerically, regardless of the `type` and `sortByValue` settings.
-
-- `true` — Use enum values.
-- `false` — Use enum names.
 
 ### partitionByComment
 
@@ -361,7 +353,6 @@ Custom groups have a higher priority than any predefined group.
                   partitionByNewLine: false,
                   newlinesBetween: 'ignore',
                   sortByValue: false,
-                  forceNumericSort: false,
                   groups: [],
                   customGroups: [],
                 },
@@ -393,7 +384,6 @@ Custom groups have a higher priority than any predefined group.
                 partitionByNewLine: false,
                 newlinesBetween: 'ignore',
                 sortByValue: false,
-                forceNumericSort: false,
                 groups: [],
                 customGroups: [],
               },
--- a/docs/content/rules/sort-enums.mdx
+++ b/docs/content/rules/sort-enums.mdx
@@ -170,7 +170,7 @@ Specifies the sorting locales. Refer To [String.prototype.localeCompare() - loca
 
 ### sortByValue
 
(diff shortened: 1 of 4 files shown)


rules/sort-enums/types.ts
```diff
@@ -37,6 +37,14 @@ export type Options = Partial<
      */
     newlinesBetween: NewlinesBetweenOption
 
+    /**
+     * Whether to sort enum members by their values instead of names. When true,
+     * compares enum values; when false, compares enum member names.
+     *
+     * @default false
```
azat-io: Is this a mistake? It seems that now the default value is ifNumericEnum.

docs/content/rules/sort-enums.mdx
```diff
@@ -170,24 +170,16 @@ Specifies the sorting locales. Refer To [String.prototype.localeCompare() - loca
...
 ### sortByValue
 
-<sub>default: `false`</sub>
+<sub>default: `'ifNumericEnum'`</sub>
 
 Controls whether sorting should be done using the enum's values or names.
 
 - `true` — Use enum values.
 - `false` — Use enum names.
+- `ifNumericEnum` — Use enum values only if the enum is numeric.
```
azat-io: What do you think about specifying the value as `'ifNumericEnum'`? So that the user doesn't accidentally think that true and false are strings.

docs/content/rules/sort-enums.mdx
```diff
@@ -170,24 +170,16 @@ Specifies the sorting locales. Refer To [String.prototype.localeCompare() - loca
 
 ### sortByValue
 
-<sub>default: `false`</sub>
+<sub>default: `'ifNumericEnum'`</sub>
```
azat-io, line 173: I'm also a little confused that the field values ​​are different data types: boolean and string.

hugop95: @azat-io We could go with

`sortByValue: "always" | "never" | "ifNumericEnum"` (similarly to how `newlinesBetween` worked).

However, this is a bigger breaking change:
- With `sortByValue: boolean | "ifNumericEnum"`, it would only be a breaking change for:
  - Users who had `forceNumericEnum: true` in their config.
- With `sortByValue: "always" | "never" | "ifNumericEnum"`, it would be a breaking change for:
  - Users who had `forceNumericEnum: true` in their config.
  - Users who had `sortByValue: true` in their config.

We can assume that users wouldn't set up the default value explicitely (`sortByValue: false` and `forceNumericEnum: false`).

I'll change to `sortByValue: "always" | "never" | "ifNumericEnum"` if that's OK for you.

docs/content/rules/sort-enums.mdx
hugop95: @azat-io Good point !

[`482ab96` (#586)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/586/commits/482ab96f7a05944ac7e03e37ccd91a9304f3c677)

rules/sort-enums/types.ts
hugop95: @azat-io I missed that 👍

[`19910e4` (#586)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/586/commits/19910e49cf175da6d5a5e3b3b020e59b791d1fcf)

docs/content/rules/sort-enums.mdx
azat-io: Yes, looks good 👍

azat-io:
LGTM. But let's get rid of mixed types `boolean | 'ifNumericEnum'` -> `'always' | 'never' | 'ifNumericEnum'`.

hugop95:
@azat-io I'll update it this afternoon and ping you for a new review 👍

hugop95:
@azat-io Done.

azat-io (approved):

PR #590, opened 2025-09-22
feat!: drop legacy configs

azat-io:
### Description

Drop legacy configs support.

### Additional context

#566.

---

- [ ] Bug fix
- [x] New Feature
- [x] Documentation update
- [ ] Other

### DIFF
--- a/docs/content/configs/recommended-alphabetical.mdx
+++ b/docs/content/configs/recommended-alphabetical.mdx
@@ -13,9 +13,6 @@ keywords:
   - eslint-plugin-perfectionist
 ---
 
-import CodeTabs from '../../components/CodeTabs.svelte'
-import dedent from 'dedent'
-
 Configuration for the `eslint-plugin-perfectionist` plugin, which provides all plugin rules with predefined options: alphabetical sorting in ascending order.
 
 Uses [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) to sort elements.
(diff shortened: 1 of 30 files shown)


hugop95 (commented):
@azat-io

I think that this change would not be beneficial:
- Legacy config support costs little maintenance on our side AFAIK.
- A lot of (especially old/large) projects are still running on the old format.
  - The new format is not to the liking of everyone.

I'm afraid that a large portion of users will never want to run v5 because of this until a long time.

azat-io:
@hugop95 

I have mixed feelings about this.

On the one hand, it's a good way to encourage users to migrate to ESLint v9. On the other hand, we are not obligated to do so.

It's quite possible that most of our users are on outdated projects that do not update dependencies and do not extend ESLint configurations in any way, since migration and changes are difficult for large and old projects.

Also, we are not completely removing support for the old version. There is a workaround, which is documented.

But this may alienate users.

I understand that this pull request looks aggressive, and maybe it really should be rejected. Do you think it's premature?

hugop95:
@azat-io 

> Do you think it's premature?

In my opinion, yes.

> On the one hand, it's a good way to encourage users to migrate to ESLint v9

Unfortunately, ESLint v9's launch was pretty bad, and a lot of users/projects don't want to spend hours trying to understand how to migrate their projects.

Enforcing it (or sounding like we don't support the old version) will definitely alienate a good chunk of users.

---

I see official legacy ESLint support as a low effort but high reward choice today.

Maybe things will change in a year, but even by then I still have doubts.

azat-io:
Okay, thank you for your opinion. I'll close this PR

PR #600, opened 2025-10-18
feat: add sort-regexp rule

azat-io:
### Description

A new rule for sorting regular expressions is added. It works with:

- capture groups (`/(error|warning|info|debug)/`)
- character classes (`/[zxa-f0-9A-Z]/`)
- flags (`/pattern/igmus`)

### Additional context

N/A.

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/test/rules/sort-regexp.test.ts
+++ b/test/rules/sort-regexp.test.ts
@@ -826,6 +826,15 @@ describe('sort-regexp', () => {
       })
     })
 
+    it('keeps escape alternatives before wildcard fallback', async () => {
+      await valid({
+        code: dedent(String.raw`
+          const UNESCAPE = /\\([0-9A-Fa-f]{1,6}[ \f\n\r\t]?|[\s\S])/g;
+        `),
+        options: [options],
+      })
+    })
+
     it('skips sorting when alternatives are unnamed capturing groups', async () => {
       await valid({
         code: dedent`
@@ -2197,6 +2206,15 @@ describe('sort-regexp', () => {
         code: dedent`
           /(specific|.+|.*|.?)/
         `,
+        options: [options],
+      })
+    })
+
+    it('keeps escape alternatives before wildcard fallback', async () => {
+      await valid({
+        code: dedent(String.raw`
+          const UNESCAPE = /\\([0-9A-Fa-f]{1,6}[ \f\n\r\t]?|[\s\S])/g;
+        `),
         options: [options],
       })
     })
--- a/rules/sort-regexp/get-first-character-paths.ts
+++ b/rules/sort-regexp/get-first-character-paths.ts
@@ -17,6 +17,12 @@ export interface FirstCharacterPath {
 
   /** Indicates whether additional characters must follow to complete the match. */
   requiresMore: boolean
+
+  /**
+   * Indicates whether the alternative can consume more characters after the
+   * prefix.
+   */
+  canMatchMore: boolean
 }
 
 /** Matcher produced from a character class AST node. */
@@ -60,8 +66,14 @@ interface AnalysisContext {
   /** Cache storing computed minimum lengths for AST nodes. */
   minLengthCache: WeakMap<object, LengthResult>
 
+  /** Cache storing computed maximum lengths for AST nodes. */
+  maxLengthCache: WeakMap<object, LengthResult>
+
   /** Alternatives currently on the recursion stack. */
-  activeAlternatives: Set<Alternative>
+  minLengthActiveAlternatives: Set<Alternative>
+
+  /** Alternatives on the recursion stack for maximum-length calculation. */
+  maxLengthActiveAlternatives: Set<Alternative>
 
   /** Indicates whether collection exceeded the maximum allowed paths. */
   limitExceeded: boolean
@@ -72,8 +84,9 @@ interface AnalysisContext {
 
 /** Internal extension that includes metadata needed during traversal. */
 interface FirstCharacterPathInternal extends FirstCharacterPath {
-  /** Mirrors the public flag for convenience when mutating paths. */
+  /** Mirrors the public flags for convenience when mutating paths. */
   requiresMore: boolean
+  canMatchMore: boolean
 }
 
 type LengthResult = LengthInfo | null
@@ -91,8 +104,10 @@ export function getFirstCharacterPaths(
   alternative: Alternative,
 ): FirstCharacterPath[] {
   let context: AnalysisContext = {
+    minLengthActiveAlternatives: new Set(),
+    maxLengthActiveAlternatives: new Set(),
     minLengthCache: new WeakMap(),
-    activeAlternatives: new Set(),
+    maxLengthCache: new WeakMap(),
     limitExceeded: false,
     pathCount: 0,
   }
@@ -104,6 +119,86 @@ export function getFirstCharacterPaths(
   }
 
   return paths
+}
+
+/**
+ * Computes the maximum possible length for an element.
+ *
+ * @param element - AST element to analyze.
+ * @param context - Shared traversal context.
+ * @returns Maximum length in characters, `2` for "two or more", or `null` if
+ *   unknown.
+ */
+function getElementMaxLength(
+  element: Element,
+  context: AnalysisContext,
+): LengthResult {
+  // Defensive guard triggers only when traversal exceeded path limit earlier.
+  /* c8 ignore next 3 */
+  if (context.limitExceeded) {
+    return null
+  }
+
+  let cached = context.maxLengthCache.get(element)
+
+  if (cached !== undefined) {
+    return cached
+  }
+
+  let result: LengthResult = null
+
+  switch (element.type) {
+    case 'CharacterClass':
+    case 'CharacterSet':
+    case 'Character': {
+      result = 1
+      break
+    }
+    case 'CapturingGroup':
+    case 'Group': {
+      result = getGroupMaxLength(element, context)
+      break
+    }
+    case 'Backreference': {
+      result = null
+      break
+    }
+    case 'Quantifier': {
+      let innerLength = getElementMaxLength(element.element, context)
+
+      if (innerLength === null) {
+        result = null
+        break
+      }
+
+      // Numerical sentinels are unreachable with current AST inputs.
+      /* c8 ignore start */
+      if (innerLength === 0 || element.max === 0) {
+        result = 0
+        break
+      }
+
+      if (element.max === Infinity) {
+        result = 2
+        break
+      }
+      /* c8 ignore stop */
+
+      result = multiplyLength(innerLength, element.max)
+      break
+    }
+    case 'Assertion': {
+      result = 0
+      break
+    }
+    default: {
+      result = null
+    }
+  }
+
+  context.maxLengthCache.set(element, result)
+
+  return result
 }
 
 /**
@@ -128,6 +223,7 @@ function collectFirstCharacterPathsFromElement(
         {
           matcher: { type: 'character-class', value: element },
           requiresMore: false,
+          canMatchMore: false,
         },
       ]
     }
@@ -146,6 +242,7 @@ function collectFirstCharacterPathsFromElement(
         {
           matcher: { type: 'character-set', value: element },
           requiresMore: false,
+          canMatchMore: false,
         },
       ]
     }
@@ -157,6 +254,7 @@ function collectFirstCharacterPathsFromElement(
         {
           matcher: { value: element.value, type: 'character' },
           requiresMore: false,
+          canMatchMore: false,
         },
       ]
     }
@@ -195,10 +293,14 @@ function collectFirstCharacterPathsFromAlternative(
 
     if (elementPaths.length > 0) {
       let restLength = getElementsMinLength(elements, index + 1, context)
+      let restMaxLength = getElementsMaxLength(elements, index + 1, context)
 
       if (restLength !== null) {
+        let restCanMatchMore = restMaxLength !== 0
+
         for (let path of elementPaths) {
           addPath(results, context, {
+            canMatchMore: path.canMatchMore || restCanMatchMore,
             requiresMore: path.requiresMore || restLength > 0,
             matcher: path.matcher,
           })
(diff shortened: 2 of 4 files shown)


test/rules/sort-regexp.test.ts
```diff
@@ -0,0 +1,2835 @@
+/* Cspell:ignore gimsu igmus yusmig ysumgi Ωmega Δelta */
```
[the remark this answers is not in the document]
azat-io: 🤦

docs/content/rules/sort-regexp.mdx
```diff
@@ -0,0 +1,272 @@
...
+import CodeExample from '../../components/CodeExample.svelte'
+import CodeTabs from '../../components/CodeTabs.svelte'
+import dedent from 'dedent'
+
+Enforce consistent ordering in regular expressions: flags, character-class elements, and alternation branches inside capture groups or the top-level pattern.
+
+Large patterns become easier to maintain when related branches are grouped and repeatedly used flags or character classes follow the same order. This rule helps you spot missing alternatives, avoid duplicated work, and keeps refactors safer by applying one sorting strategy across your codebase.
+
+The rule is **safe** – it only reorders elements without changing their meaning.
+
```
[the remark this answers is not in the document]
hugop95: @azat-io 

There are indeed dangers with the current state:

image

but

image

---

I don't think we should include this rule in the `recommended`, they are very likely other edge cases we will be missing.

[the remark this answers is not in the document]
azat-io: @hugop95 I fixed it

test/rules/sort-regexp.test.ts
```diff
@@ -0,0 +1,2835 @@
...
+          /[a\x61]/
+        `),
+        options: [customOptions],
+      })
+    })
+
+    it('sorts character classes', async () => {
+      let customOptions = {
+        ...options,
+        type: 'natural',
```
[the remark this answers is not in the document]
azat-io: Fixed.

rules/sort-regexp.ts
```diff
@@ -617,6 +647,22 @@ function getAlternativeAlias(alternative: Alternative): string | null {
...
+  if (first.length === 0 || second.length === 0) {
+    return true
+  }
+
+  if (first.length === second.length) {
+    return first === second
+  }
+
+  if (first.length < second.length) {
+    return second.startsWith(first)
```
hugop95: @azat-io 

This won't work for cases such as: `/(a|.*ab)/.exec("ab")` and `/(.*ab|a)/.exec("ab")`.

```ts
/(a|.*ab)/.exec("ab")
(2) ['a', 'a', index: 0, input: 'ab', groups: undefined]

/(.*ab|a)/.exec("ab")
(2) ['ab', 'ab', index: 0, input: 'ab', groups: undefined]
```

I haven't looked deep into it, but we probably need to `includes` for maximal safety.

In fact, it simply might never be fully safe with `exec`.

Maybe we would have to restrain sorting to specific usages (`.test` for example).

docs/content/rules/sort-regexp.mdx
```diff
@@ -0,0 +1,272 @@
...
+import CodeExample from '../../components/CodeExample.svelte'
+import CodeTabs from '../../components/CodeTabs.svelte'
+import dedent from 'dedent'
+
+Enforce consistent ordering in regular expressions: flags, character-class elements, and alternation branches inside capture groups or the top-level pattern.
+
+Large patterns become easier to maintain when related branches are grouped and repeatedly used flags or character classes follow the same order. This rule helps you spot missing alternatives, avoid duplicated work, and keeps refactors safer by applying one sorting strategy across your codebase.
+
+The rule is **safe** – it only reorders elements without changing their meaning.
```
hugop95, line 22: @azat-io 

Honestly, I'm not confident that we can 100% ensure full safety here at this stage. I would even say that regex re-ordering probably has more weird edge cases that any other rules we have (which already have complex edge cases due to dependency ordering).

I expect bugs related to ordering that change the behavior when the rule is launched. It might be OK in 95% of the cases, but I have doubt about the remaining complex cases.

rules/sort-regexp/get-sorting-node-name.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @param parameters - Alternative metadata and rule options.
+ * @returns String representation used during sorting.
+ */
+export function getSortingNodeName({
+  alternativeAlias,
+  alternative,
+  options,
+}: SortingNodeNameParameters): string {
+  if (!options.ignoreAlias && alternativeAlias) {
+    return `${alternativeAlias}: ${alternative.raw}`
```
hugop95: @azat-io 

I would remove the space after `:`.

azat-io: @hugop95 Quick question about the space after `:` - you suggested removing it from `"pet: cat"` → `"pet:cat"`.

Is this a style preference, or is there a specific reason? I kept the space for readability in error messages, but happy to change if there's a convention I'm missing.

docs/content/rules/sort-regexp.mdx
azat-io: I've added additional safeguards in the latest commits:

  - **ES2024 `/v` flag guard** - skips character class sorting to avoid breaking set operations (`&&`, `--`)
  - **More backreference tests** - covers named groups and complex scenarios
  - **Existing shadowing detection** - already handles `/(ab|a)/` cases

The rule is conservative - it skips sorting when it detects potentially unsafe patterns. That said, I agree 100% safety is hard to guarantee with regex.

azat-io:
I’m currently testing the rule on large codebases.

So far I haven't managed to fix the issue where line breaks are split or checked.

- Before: `/\r\n|[\r\n\u2028\u2029]/`
- After: `/[\n\r\u2028\u2029]|\r\n/`

In JavaScript, alternatives are tried from left to right. When \r\n is moved behind the character class, the engine will match the single `\r` first, leaving the `\n` unmatched.

This matters in real code:

```js
const LINEBREAK_MATCHER_OLD = /\r\n|[\r\n\u2028\u2029]/;
const LINEBREAK_MATCHER_NEW = /[\n\r\u2028\u2029]|\r\n/;

let text = 'a\r\nb';

console.log(text.split(LINEBREAK_MATCHER_OLD)); // ['a', 'b']
console.log(text.split(LINEBREAK_MATCHER_NEW)); // ['a', '', 'b']
```

azat-io:
I fixed the issue mentioned above and added the necessary tests.

I also tested the new ESLint rule on several large projects:

- [Babel](https://github.com/babel/babel)
- [ESLint](https://github.com/eslint/eslint)
- [TypeScript ESLint](https://github.com/typescript-eslint/typescript-eslint)
- [Vue](https://github.com/vuejs/core)
- [Astro](https://github.com/withastro/astro)
- [Prettier](https://github.com/prettier/prettier)

All changes and auto-fixes are cosmetic in nature and do not create functional changes.

@hugop95 This PR is now ready for review.

rules/sort-regexp.ts
```diff
@@ -0,0 +1,388 @@
...
+
+          if (
+            hasShadowingAlternatives({
+              alternatives: alternative.parent.alternatives,
+              flags: literalNode.regex.flags,
+            })
+          ) {
+            literalsWithShadowing.add(literalNode)
+            return
+          }
```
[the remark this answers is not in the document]
azat-io: `hasShadowingAlternatives` already treats `/\r\n|[\r\n\u2028\u2029]/` as unsafe, so the rule marks that literal as shadowed and leaves both the disjunction and the character class untouched. This behavior is covered by `test/rules/sort-regexp-shadowing.test.ts:405` and `test/rules/sort-regexp.test.ts:588`

docs/content/rules/sort-regexp.mdx
```diff
@@ -0,0 +1,272 @@
...
+<sub>default: `false`</sub>
+
+Determines whether named capturing group aliases (e.g., `(?<role>...)`) are considered during comparison.
+
+- `false` — Sort by `alias: pattern` (default).
+- `true` — Ignore the alias and only compare the branch content.
+
+This is useful when you only care about the pattern and not how it is aliased.
+
+### customGroups
```
hugop95, line 137: @azat-io 

For other rules, this section is after `Groups`, so I would do the same here for consistency.

docs/content/rules/sort-regexp.mdx
```diff
@@ -0,0 +1,272 @@
...
+- `selector` — Filter by element type:
+  - `'alias'` — Named capturing group alternatives.
+  - `'pattern'` — Plain alternatives without aliases.
+- `elementNamePattern` — Regex applied to the alias (if present).
+- `elementValuePattern` — Regex applied to the branch content.
+- `type`, `order`, `fallbackSort` — Override the respective global options for the group.
+
+Custom groups are evaluated in order; the first match wins and overrides predefined groups.
+
+### groups
```
hugop95, line 182: @azat-io 

I would take the same `sub` sections as for other rules for consistency as well, even if this feature probably won't likely be used for this rule in particular:

https://github.com/azat-io/eslint-plugin-perfectionist/blob/f9919cb36d64c5acb7d41723e874bb316e84b452/docs/content/rules/sort-objects.mdx?plain=1#L460-L477

package.json
```diff
@@ -57,6 +57,7 @@
     "test:unit": "vitest --run --coverage"
   },
   "dependencies": {
+    "@eslint-community/regexpp": "^4.12.1",
```
hugop95, line 60: @azat-io 

For information:
- `@eslint-community/regexpp`'s unpacked size is **474kB** according to [NPM](https://www.npmjs.com/package/@eslint-community/regexpp):
- [Perfectionist V4](https://www.npmjs.com/package/eslint-plugin-perfectionist) is at **327kb** 
- [`@typescript-eslint/utils`](https://www.npmjs.com/package/@typescript-eslint/utils) is at **195kB**.
- [`natural-orderby`](https://www.npmjs.com/package/natural-orderby) is at **72kb**.

Just something to keep in mind based on the adoption of the rule.

rules/sort-regexp.ts
```diff
@@ -0,0 +1,416 @@
...
+import { reportAllErrors } from '../utils/report-all-errors'
+import { ORDER_ERROR } from '../utils/report-errors'
+import { getSettings } from '../utils/get-settings'
+import { isSortable } from '../utils/is-sortable'
+import { sortNodes } from '../utils/sort-nodes'
+import { pairwise } from '../utils/pairwise'
+import { complete } from '../utils/complete'
+import { compare } from '../utils/compare'
+
+type SortRegExpSortingNode = SortingNode<TSESTree.Literal>
```
hugop95, line 42: @azat-io 

We could put and export this type in `rules/sort-regexp/types.ts`, and use it in places doing `SortingNode<TSESTree.Literal>` again.
- In `createCharacterClassSortingNode` in this file.
- `rules/sort-regexp/create-flag-sorting-nodes.ts`.
- `rules/sort-regexp/create-character-class-sorting-node.ts`.

rules/sort-regexp.ts
```diff
@@ -0,0 +1,416 @@
...
+
+type MessageId = 'unexpectedRegExpOrder'
+
+/**
+ * Retrieves the original source text for a character class element.
+ *
+ * @param parameters - Character class element metadata.
+ * @returns Slice of the literal raw string covering the element.
+ */
+function getCharacterClassElementRawText({
```
hugop95, line 54: @azat-io 

[nit] It's a bit strange IMO to have this local function at the top. What about putting it at the very bottom ?

rules/sort-regexp.ts
```diff
@@ -0,0 +1,416 @@
...
+              sortingNode.node === sortedAlternatives.at(index)?.node,
+          )
+
+          if (isAlreadySorted) {
+            return
+          }
+
+          reportAllErrors<MessageId, SortRegExpSortingNode>({
+            availableMessageIds: {
+              unexpectedGroupOrder: 'unexpectedRegExpOrder',
```
hugop95, line 239: @azat-io 

Could we create a dedicated `unexpectedRegExpGroupOrder` messageID to stay consistent with other rules ?

test/rules/sort-regexp.test.ts
```diff
@@ -0,0 +1,3267 @@
...
+        output: dedent`
+          /(?<b>(?<a>a)|b)/
+        `,
+        code: dedent`
+          /(?<b>b|(?<a>a))/
+        `,
+      })
+    })
+
+    it('respects custom groups by alias name', async () => {
```
hugop95, line 1945: @azat-io 

I think that the `groups` feature doesn't work: if you change `groups` to `["unknown", "digits"]`, the result is still valid.

The only case that uses `sortNodesByGroups` is for `sortAlternatives`.

hugop95 (commented):
@azat-io 

This is a very ambitious initiative !

The PR is very large, so I haven't reviewed everything yet, but I have some comments already on a couple of things.

One problem that I think is blocking right now is that the `groups` feature doesn't work (or at the very least, tests do not test it correctly, I've left a comment for that).

My other feedback is regarding the current approach: the rule sorts 3 things as you mention in the description:
- capture groups `(/(error|warning|info|debug)/)`
- character classes `(/[zxa-f0-9A-Z]/)`
- flags `(/pattern/igmus)`

I wonder if it wouldn't have been better to go with 3 separate rules:
- They are all linked to RegExp, but represent very different things.
- We already have one rule that sorts multiple things at the same time: `sort-switch-case` (it can sort "groups" of `cases`, and what is inside a consecutive list of `cases`).
  - It's a bit of our ugly duckling rule: it's the only one that doesn't supports `groups`, because this notion mean different things depending on what we want to sort. 
  - We have a bit of the same case here:
    - What do `groups`/`customGroups` apply to ? `flags`, `capture groups`, or `character classes` ?
    - Can users `group` flags only, but not other things ?
    - I think that in all cases, this should be documented, because it's impossible to know otherwise.
- Code wise, it's harder to understand the sort of all 3 things in the same rule.
  - Maybe we could have common functions use by the 3 rules, but each rule is kept small and only focuses on sorting one thing. 
- `ignoreAlias` is only useful for `capture groups` sorting.
- This way, it allows users might to enable `flags` sorting, but not `capture groups`.
  - I think that it's important: flag sorting is **fully safe**, but capture groups isn't in theory.

Let me know what you think !

azat-io:
Closed with this reason: https://github.com/azat-io/eslint-plugin-perfectionist/issues/566#issuecomment-3568080700

Issue #605, opened 2025-11-05
Feature: (Sort imports - reverse multi-line and single-line import order)


travis-mod-io:
### What rule do you want to change?

sort-imports

### Describe the problem

I want to be able to swap the sorting position for multi-line imports vs single-line imports. e.g. for 'asc' have the multi-line imports on top, and for 'desc' have the multi-line imports on the bottom.

### Code example

```
'perfectionist/sort-imports': [
  'error',
  {
    type: 'line-length',
    order: 'desc',
    multilinePosition: 'bottom',
  },
],
```

```
import { toRefs, ref, computed } from 'vue'
import { useStuff } from '@composables'
import {
  OPTION_ONE_NAME,
  OPTION_TWO_NAME,
} from '@options/example_file.js'
```

Or alternatively

```
'perfectionist/sort-imports': [
  'error',
  {
    type: 'line-length',
    order: 'asc',
    multilinePosition: 'top',
  },
],
```

```
import {
  OPTION_ONE_NAME,
  OPTION_TWO_NAME,
} from '@options/example_file.js'
import { useStuff } from '@composables'
import { toRefs, ref, computed } from 'vue'
```

### Validations

- [x] Read the [docs](https://perfectionist.dev).
- [x] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

hugop95:
# Dev notes

A simple `multiline`, and optionally `singleline` modifier usable in `groups` is enough.

azat-io:
Added in 2c954d48f18714af81cb33b8344edf986761a20d. Released in [v5.0.0](https://github.com/azat-io/eslint-plugin-perfectionist/releases/tag/v5.0.0).

If you want to support our release, you can repost:

https://bsky.app/profile/azat.io/post/3ma6evf26xc2p

https://x.com/azat_io_en/status/2001228025644425659

travis-mod-io:
> A simple `multiline`, and optionally `singleline` modifier usable in `groups` is enough.

Just made use of this from the new release, works perfectly 😄 thanks heaps!

PR #606, opened 2025-11-05
feat(sort-imports): add `multiline` and `singleline` modifiers

hugop95:
- Resolves https://github.com/azat-io/eslint-plugin-perfectionist/issues/605

### Description

This PR adds two modifiers to `sort-import`, with the lowest priority:
- `multiline`.
- `singleline`.

- [x] New Feature

### DIFF
--- a/utils/is-node-on-single-line.ts
+++ b/utils/is-node-on-single-line.ts
@@ -1,5 +1,11 @@
 import type { TSESTree } from '@typescript-eslint/types'
 
+/**
+ * Returns if a node is on a single line.
+ *
+ * @param node - The node to check.
+ * @returns True if the node is on a single line, false otherwise.
+ */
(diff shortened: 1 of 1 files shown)


utils/is-node-on-single-line.ts
```diff
@@ -0,0 +1,5 @@
+import type { TSESTree } from '@typescript-eslint/types'
+
+export function isNodeOnSingleLine(node: TSESTree.Node): boolean {
```
azat-io, line 9: What do you think about adding simple JSDoc to make the utilities look consistent?

hugop95: @azat-io 👍

[`9fa7df7` (#606)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/606/commits/9fa7df73f29ed735031d8a2027e02ac5c6d30136)

azat-io (approved):

PR #610, opened 2025-11-14
feat(sort-imports)!: drop deprecated selectors

hugop95:
## Description

Some selectors are marked as deprecated and to be deleted for v5.0.0.

### List of removed selectors

- `internal-type` ➡️ To be replaced with `type-internal` (`type` modifier, `internal` selector).
- `external-type` ➡️ To be replaced with `type-external`.
- `sibling-type` ➡️ To be replaced with `type-sibling`.
- `builtin-type` ➡️ To be replaced with `type-builtin`.
- `index-type` ➡️ To be replaced with `type-index`.
- `object`.

- [x] Other


hugop95:
Hi @azat-io 👋

The [release v5.0.0](https://github.com/azat-io/eslint-plugin-perfectionist/pull/569) PR has a ton of comments due to automatic reviews, and this makes the page sometimes fail to load. As more PRs get merged in `next`, more comments will be added, making it even harder to access it.

image

What do you think about closing and re-opening it, so we start back with a fresh state ?

azat-io:
@hugop95 I cleaned up the pull request from strange and outdated comments. It should work better now.

azat-io (approved):
Nice PR! 👍

PR #616, opened 2025-11-22
refactor: reduce message IDs duplication in tests

hugop95:
### Description

This PR reduces code duplication related to message IDs.
Changes are ISO functional.

- [x] Other

### DIFF
--- a/rules/sort-array-includes.ts
+++ b/rules/sort-array-includes.ts
@@ -47,11 +47,10 @@ import { complete } from '../utils/complete'
 /** Cache computed groups by modifiers and selectors for performance. */
 let cachedGroupsByModifiersAndSelectors = new Map<string, string[]>()
 
-export const ORDER_ERROR_ID = 'unexpectedArrayIncludesOrder'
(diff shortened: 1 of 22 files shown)


hugop95:
@azat-io 

I understand what you mean. IMO:
- The error messages coupling is acceptable and helps a little bit in case of refactors.
- The coupling will still exist (but indirectly) through the helper function (but that's one helper function to maintain).

For now, let's undo the changes then, and not export the constants 👍 
The duplicated strings are not that problematic anyway. I've reverted changes so tests don't use the constant anymore. Now, the only changes are:
- Put `data` elements on the same line - [`f341e4c` (#616)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/616/commits/f341e4c73b37db676f6f1807c8005ffece00fb42)
- Rules don't export message IDs anymore - [`81a1b73` (#616)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/616/commits/81a1b7317453b33b97ad93f603ef7711998242ef)

Issue #593, opened 2025-10-03
Feature: sort imports from the same module by "import type"


IlyaSemenov:
### What rule do you want to change?

sort-imports

### Describe the problem

With alphabetic imports sorting, I would like to be able to enforce strict order between type- and value- imports from the same module.

### Code example

For example, with this setup:

```ts
"perfectionist/sort-imports": ["error", {
  groups: [
    ["builtin", "builtin-type"],
    ["external", "external-type"],
    ["internal", "internal-type"],
    ["parent", "parent-type"],
    ["sibling", "sibling-type", "index", "index-type"],
    "unknown",
  ],
  newlinesBetween: "always",
  order: "asc",
  type: "alphabetical",
}],
```

this is valid code:

```ts
import { antfu } from "@antfu/eslint-config"
import type { Awaitable } from "@antfu/eslint-config"
import annotation from "eslint-plugin-annotation"
```

and this is too:

```ts
import type { Awaitable } from "@antfu/eslint-config"
import { antfu } from "@antfu/eslint-config"
import annotation from "eslint-plugin-annotation"
```

I would like to be able to enforce either of the above (configurable — either put type imports first, or second, or do not sort — by default, as it works now).

### Validations

- [x] Read the [docs](https://perfectionist.dev).
- [x] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

hugop95:
Thinking out loud: 

- It's a custom extension of [`fallbackSort`](https://perfectionist.dev/rules/sort-imports#fallbacksort).
- Something like `fallbackSort: { type: "type-import-first", order: "asc/desc" }` could potentially work.

To see if it complexifies things too much or not.

cusxio:
I believe this feature would be particularly useful during code review.

IlyaSemenov:
Perhaps the `fallbackSort` should be an array then (or `MaybeArray` for backwards compatibility)? Because I imagine someone would want to use `type-import-first` alongside one of the existing options.

hugop95:
I'll start investigating this more after https://github.com/azat-io/eslint-plugin-perfectionist/pull/615.

hugop95:
@IlyaSemenov 

> Perhaps the fallbackSort should be an array then (or MaybeArray for backwards compatibility)? Because I imagine someone would want to use type-import-first alongside one of the existing options.

[...]

PR #617, opened 2025-11-23
refactor: improve sorting override framework 

hugop95:
- Pre-work for https://github.com/azat-io/eslint-plugin-perfectionist/issues/593

## Description

Our current sorting system relies on the following:
- 4 hardcoded sorting types (`alphabetical`, `natural`, `line-length`, `custom`).
- By default, the value used for sorting is a [`SortingNode`](https://github.com/azat-io/eslint-plugin-perfectionist/blob/dda2e6e4e2b16274d182a2a2a13829bc54071921/types/sorting-node.ts)'s [`name`](https://github.com/azat-io/eslint-plugin-perfectionist/blob/dda2e6e4e2b16274d182a2a2a13829bc54071921/types/sorting-node.ts#L81).
  - But we can choose a different value through [`nodeValueGetter`](https://github.com/azat-io/eslint-plugin-perfectionist/blob/dda2e6e4e2b16274d182a2a2a13829bc54071921/utils/sort-nodes.ts#L15) and [`fallbackSortNodeValueGetter`](https://github.com/azat-io/eslint-plugin-perfectionist/blob/dda2e6e4e2b16274d182a2a2a13829bc54071921/utils/sort-nodes.ts#L13). 

## The problem

`nodeValueGetter` and `fallbackSortNodeValueGetter` are now lacking for future use cases.

With https://github.com/azat-io/eslint-plugin-perfectionist/issues/593, we would like to be able to add a fallback sort which logic is not convenient to represent with just those two properties:

The two following imports are equal with the `alphabetical` and `natural` sorts in this case:
- `import { a } from "a"`.
- `import type { Type } from "a"`. 

In order to be able to place the `type` import first with the current options we have, we would need to make sure that the first import's `name` comes before the second one.

This can be achieved with a `nodeValueGetter` that transforms the import's name to `aa` and the second to `ab`, but this isn't really clean.

## Solution

Rather than allowing rules to pass `nodeValueGetter` and `fallbackSortNodeValueGetter`, allow them to directly pass **comparison functions**. Those comparison functions can be computed from options.

For most rules, the comparison functions to pass are the default ones (alphabetical sorting, natural sorting, etc).

For rules that allow sorting by value (`sort-enums` for example), we simply compute an alphabetical/natural sorting function that takes the node's **value** field instead of name, and pass those to `sort-nodes`/`sort-nodes-by-groups`.   

For https://github.com/azat-io/eslint-plugin-perfectionist/issues/593, we simply need to create and pass a comparison function that takes two imports, and gives priority to the one with `type`. 

## How it looks

<details>
<summary>🧑‍💻 The default comparison functions passed</summary>

```ts
export type ComparatorByOptionsComputer<
  S extends CommonOptions,
  T extends SortingNode,
> = (options: S) => Comparator<T>

export type Comparator<T extends SortingNode> = (a: T, b: T) => number

type Options = Pick<
  CommonOptions,
  | 'specialCharacters'
  | 'fallbackSort'
  | 'ignoreCase'
  | 'alphabet'
  | 'locales'
  | 'order'
  | 'type'
>

export let defaultComparatorByOptionsComputer: ComparatorByOptionsComputer<
  Options,
  SortingNode
> = options => {
  switch (options.type) {
    case 'alphabetical':
      return (a, b) => compareAlphabetically(a.name, b.name, options)
    case 'line-length':
      return (a, b) => compareByLineLength(a, b, options)
    case 'unsorted':
      return unsortedComparator
    case 'natural':
      return (a, b) => computeNaturalSort(a.name, b.name, options)
    case 'custom':
      return (a, b) => compareByCustomSort(a.name, b.name, options)
    default:
      throw new UnreachableCaseError(options.type)
  }
}
```

</details>

<details>
<summary>🧑‍💻 The comparison functions for `sort-enums` (which allows sorting by value)</summary>

```ts
export function buildComparatorByOptionsComputer(
  isNumericEnum: boolean,
): ComparatorByOptionsComputer<
  Required<Options[number]>,
  SortEnumsSortingNode
> {
  return options => {
    switch (options.sortByValue) {
      case 'ifNumericEnum':
        if (isNumericEnum) {
          return byNumericValueComparatorComputer(options)
        }
        return defaultComparatorByOptionsComputer(options)
      case 'always':
        if (isNumericEnum) {
          return byNumericValueComparatorComputer(options)
        }
        return byNonNumericValueComparatorComputer(options)
      case 'never':
        return defaultComparatorByOptionsComputer(options)
      default: {
        throw new UnreachableCaseError(options.sortByValue)
      }
    }
  }
}

let byNonNumericValueComparatorComputer: ComparatorByOptionsComputer<
  Required<Options[number]>,
  SortEnumsSortingNode
> = options => {
  switch (options.type) {
    case 'alphabetical':
      return (a, b) =>
        compareAlphabetically(
          a.value?.toString() ?? '',
          b.value?.toString() ?? '',
          options,
        )
    case 'line-length':
      return (a, b) => compareByLineLength(a, b, options)
    case 'unsorted':
      return unsortedComparator
    case 'natural':
      return (a, b) =>
        computeNaturalSort(
          a.value?.toString() ?? '',
          b.value?.toString() ?? '',
          options,
        )
    case 'custom':
      return (a, b) =>
        compareByCustomSort(a.value ?? '', b.value ?? '', options)
    default:
      throw new UnreachableCaseError(options.type)
  }
}

let byNumericValueComparatorComputer: ComparatorByOptionsComputer<
  Required<Options[number]>,
  SortEnumsSortingNode
> = options => {
  switch (options.type) {
    case 'alphabetical':
    case 'line-length':
    case 'natural':
    case 'custom':
      return (a, b) =>
        computeNaturalSort(
          a.numericValue!.toString(),
          b.numericValue!.toString(),
          options,
        )
    case 'unsorted':
      return unsortedComparator
    default:
      throw new UnreachableCaseError(options.type)
  }
}
```

</details>

- [x] Other

### DIFF
--- a/utils/compare/compute-comparators.ts
+++ b/utils/compare/compute-comparators.ts
@@ -5,6 +5,8 @@ import type {
 import type { CommonOptions } from '../../types/common-options'
 import type { SortingNode } from '../../types/sorting-node'
 
+import { unsortedComparator } from './unsorted-comparator'
+
 export function computeComparators<
   Options extends CommonOptions,
   T extends SortingNode,
@@ -12,11 +14,14 @@ export function computeComparators<
   comparatorByOptionsComputer: ComparatorByOptionsComputer<Options, T>,
   options: Options,
 ): Comparator<T>[] {
-  return [
-    comparatorByOptionsComputer(options),
-    comparatorByOptionsComputer({
-      ...options,
-      ...options.fallbackSort,
-    }),
-  ]
+  let mainComparator = comparatorByOptionsComputer(options)
+  if (mainComparator === unsortedComparator) {
+    return []
+  }
+
+  let fallbackComparator = comparatorByOptionsComputer({
+    ...options,
+    ...options.fallbackSort,
+  })
+  return [mainComparator, fallbackComparator]
 }
--- a/test/utils/compare/compute-comparators.test.ts
+++ b/test/utils/compare/compute-comparators.test.ts
@@ -3,47 +3,59 @@ import type { Mock } from 'vitest'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 
 import type { ComparatorByOptionsComputer } from '../../../utils/compare/default-comparator-by-options-computer'
-import type { CommonOptions } from '../../../types/common-options'
+import type { CommonOptions, TypeOption } from '../../../types/common-options'
 import type { SortingNode } from '../../../types/sorting-node'
 
+import { defaultComparatorByOptionsComputer } from '../../../utils/compare/default-comparator-by-options-computer'
 import { computeComparators } from '../../../utils/compare/compute-comparators'
 
 describe('compute-comparators', () => {
-  let testComparator: Mock<
+  let comparatorByOptionsComputerMock: Mock<
     ComparatorByOptionsComputer<CommonOptions, SortingNode>
   >
 
   beforeEach(() => {
-    testComparator = vi.fn()
+    comparatorByOptionsComputerMock = vi.fn()
   })
 
-  it('should compute two comparators', () => {
-    let commonOptions = {
+  it('should compute two comparators if the main one is not the unsorted comparator', () => {
+    let commonOptions = buildCommonOptions('alphabetical')
+
+    computeComparators(comparatorByOptionsComputerMock, commonOptions)
+
+    expect(comparatorByOptionsComputerMock).toHaveBeenCalledTimes(2)
+    expect(comparatorByOptionsComputerMock.mock.calls[0]?.[0]).toEqual(
+      commonOptions,
+    )
+    expect(comparatorByOptionsComputerMock.mock.calls[1]?.[0]).toEqual({
+      ...commonOptions,
+      ...commonOptions.fallbackSort,
+    })
+  })
+
+  it('should return an empty array if the main comparator is the unsorted comparator', () => {
+    let commonOptions = buildCommonOptions('unsorted')
+
+    let comparators = computeComparators(
+      defaultComparatorByOptionsComputer,
+      commonOptions,
+    )
+
+    expect(comparators).toEqual([])
+  })
+
+  function buildCommonOptions(type: TypeOption): CommonOptions {
+    return {
       fallbackSort: {
         type: 'natural',
         order: 'desc',
       },
       specialCharacters: 'trim',
-      type: 'alphabetical',
       ignoreCase: true,
       locales: ['en'],
       order: 'asc',
       alphabet: '',
-    } as const
-
-    computeComparators(testComparator, {
-      ...commonOptions,
-      fallbackSort: {
-        type: 'natural',
-        order: 'desc',
-      },
-    })
-
-    expect(testComparator).toHaveBeenCalledTimes(2)
-    expect(testComparator.mock.calls[0]?.[0]).toEqual(commonOptions)
-    expect(testComparator.mock.calls[1]?.[0]).toEqual({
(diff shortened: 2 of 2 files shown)


utils/compare/compare-naturally.ts
```diff
@@ -0,0 +1,31 @@
...
+    locales,
+    order,
+  }: Pick<
+    CommonOptions,
+    'specialCharacters' | 'ignoreCase' | 'locales' | 'order'
+  >,
+): number {
+  let naturalCompare = createNaturalCompare({
+    locale: locales.toString(),
+  })
```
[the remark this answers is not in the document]
hugop95: From a quick investigation, `natural-orderby` actually supports an array of locales, as it uses `localeCompare`

https://github.com/yobacca/natural-orderby/blob/main/src/utils/compareUnicode.ts#L3-L10

It's a minor issue as it's unlikely users are going to pass multiple locales. This can be investigated in another issue if needed as the current behavior is unchanged here.

https://github.com/azat-io/eslint-plugin-perfectionist/blob/9632cbcc3d15f9d5b74c068bb4d9fe44e28a2c70/utils/compare/compare.ts#L285-L287

rules/sort-switch-case.ts
```diff
@@ -229,11 +229,9 @@ export default createEslintRule<Options, MessageId>({
           if (b.some(node => node.isDefaultClause)) {
             return -1
           }
-          return compare({
-            a: a.at(0)!,
-            b: b.at(0)!,
-            options,
-          })
+
+          let defaultComparator = defaultComparatorByOptionsComputer(options)
```
azat-io: Let's move `defaultComparatorByOptionsComputer(options)` to the top so that we don't have to calculate the comparator at each iteration of the comparison.

utils/compare/compute-comparators.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export function computeComparators<
+  Options extends CommonOptions,
+  T extends SortingNode,
+>(
+  comparatorByOptionsComputer: ComparatorByOptionsComputer<Options, T>,
+  options: Options,
+): Comparator<T>[] {
+  return [
+    comparatorByOptionsComputer(options),
```
azat-io: With type: `'unsorted'` we still add the fallback comparator, so the array ends up sorted, whereas the previous compare short-circuited and preserved order. Is this behavior intentional? 

If not, maybe skip adding the fallback comparator when `type === 'unsorted'` so such groups remain unsorted.

hugop95: @azat-io 

> Is this behavior intentional?

I think that if it happens, it would be due to a weird user configuration: they would put `unsorted` as the main sort type, but `alphabetical` as the fallback one (which shouldn't happen unless explicitly asked by the user as `fallbackSort.type` is `unsorted` by default).

So I don't think it would be a bug, but the fix is simple nonetheless:

[`be40557` (#617)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/617/commits/be40557fef7b1593e594648333f94fecb57c96bf)

rules/sort-switch-case.ts
hugop95: @azat-io 

👍

[`431b498` (#617)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/617/commits/431b498d6d46a6d6be50b967b66bbf2d41603fdd)

azat-io (approved):
LGTM. I will review it again tomorrow morning.

azat-io (approved):

PR #618, opened 2025-11-23
test: improve test coverage

azat-io:
### Description

Clarified and repositioned our `/* v8 ignore ... */` hints so they still align with the intended AST nodes under Vitest 4’s new coverage stack (`esbuild 0.25` + `ast-v8-to-istanbul@0.3.8`). Without this, the transpiled output moved those comments away from the relevant branches, causing them to count toward coverage again; the tests now confirm all defensive paths are exercised or explicitly ignored.

### Additional context

N/A.

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other

- [ ] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves.

### DIFF
--- a/test/rules/sort-objects.test.ts
+++ b/test/rules/sort-objects.test.ts
@@ -968,6 +968,16 @@ describe('sort-objects', () => {
         options: [options],
       })
 
+      /**
+       * TODO: add regression test for nested dependency:
+       *
+       * `let { a = nested, b: { nested } = { nested: 'default' } } = obj;`.
+       *
+       * Should be invalid because a depends on nested from b, but currently
+       * passes.
+       *
+       * @see https://github.com/azat-io/eslint-plugin-perfectionist/issues/619
+       */
       await valid({
         code: dedent`
           let {


test/rules/sort-array-includes.test.ts
```diff
@@ -5765,6 +5765,34 @@ describe('sort-array-includes', () => {
...
+          code: dedent`
+            ['c', 'b', 'a'].forEach(x => console.log(x))
+          `,
+        })
+      })
+
+      it('ignores computed property access', async () => {
+        await valid({
+          code: dedent`
+            ['c', 'b', 'a']['includes']('a')
```
hugop95, line 5501: @azat-io 

I didn't think of that use case ! 😄
I don't think it will every be used though.

test/rules/sort-objects.test.ts
```diff
@@ -1024,6 +1024,16 @@ describe('sort-objects', () => {
         `,
         options: [options],
       })
+
+      await valid({
```
hugop95, line 981: @azat-io 

I don't think that this test suits: `a` and `b` don't depend on each other.

However, I realize that the following is marked as `valid` even though it shouldn't:

```ts
let {
  a = nested,
  b: { nested } = { nested: 'default' },
} = obj;
```

It's an uncommon use case !

test/utils/get-node-decorators.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    return {
+      expression: {
+        type: 'Identifier',
+        name,
+      },
+      type: 'Decorator',
+    } as unknown as TSESTree.Decorator
+  }
+
+  function createNodeWithDecorators(
```
hugop95, line 33: @azat-io 

[nit] You can write this in a concise way as well:

```ts
function createNodeWithDecorators(
  decorators?: TSESTree.Decorator[],
): NodeWithDecoratorsParameter {
    return {
      type: 'ClassDeclaration',
      ...(decorators ? { decorators} : {}) ,
    } as unknown as NodeWithDecoratorsParameter
  }
}
```

test/rules/sort-objects.test.ts
azat-io: Yes, thanks.

I created new issue:
https://github.com/azat-io/eslint-plugin-perfectionist/issues/619

And added TODO comment before the test case.

test/utils/get-node-decorators.test.ts
azat-io: Fixed

test/rules/sort-array-includes.test.ts
azat-io: This is a new requirement for Vitest to ensure test coverage. It would have been necessary if it were a lie. We need tests to ensure that 'if' is 'false'.

PR #621, opened 2025-11-25
fix(sort-objects): handle destructured dependencies

### LINKED ISSUE #619: Bug: Missing dependency detection for nested destructuring defaults in sort-objects
### Describe the bug

The `sort-objects` rule misses dependencies when a default value uses an identifier declared deeper inside another destructuring branch. It treats the order as valid even though one property depends on another.

### Code example

```js
let {
  a = nested,
  b: { nested } = { nested: 'default' },
} = obj;
```

Expected: dependency error forcing b before a. Actual: passes with no error and no fix suggested.

### ESLint version

v9.39.1

next

### Additional comments

_No response_

### Validations

azat-io:
### Description

Fix `sort-objects` dependency detection to gather identifiers from nested destructuring patterns so default values can’t be reordered ahead of their declarations. Adds regression tests for nested defaults, rest elements, and array patterns to lock the behavior.

### Additional context

Close #619

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/rules/sort-objects.ts
+++ b/rules/sort-objects.ts
@@ -144,66 +144,6 @@ export default createEslintRule<Options, MessageId>({
         sourceCode,
       })
 
-      type DestructuringPattern =
-        | TSESTree.AssignmentPattern
-        | TSESTree.ObjectPattern
-        | TSESTree.ArrayPattern
-        | TSESTree.RestElement
-        | TSESTree.Identifier
-
-      function extractNamesFromPattern(
-        pattern: DestructuringPattern,
-      ): string[] {
-        switch (pattern.type) {
-          case AST_NODE_TYPES.AssignmentPattern:
-            return extractNamesFromPattern(pattern.left as DestructuringPattern)
-          case AST_NODE_TYPES.ObjectPattern:
-            return pattern.properties.flatMap(
-              extractNamesFromObjectPatternProperty,
-            )
-          case AST_NODE_TYPES.ArrayPattern:
-            return pattern.elements.flatMap(extractNamesFromArrayPatternElement)
-          case AST_NODE_TYPES.Identifier:
-            return [pattern.name]
-          /* v8 ignore next 2 */
-          default:
-            return []
-        }
-
-        function extractNamesFromArrayPatternElement(
-          element: TSESTree.DestructuringPattern | null,
-        ): string[] {
-          if (!element) {
-            return []
-          }
-          if (element.type === AST_NODE_TYPES.RestElement) {
-            return extractNamesFromPattern(
-              element.argument as DestructuringPattern,
-            )
-          }
-
-          return extractNamesFromPattern(element as DestructuringPattern)
-        }
-
-        function extractNamesFromObjectPatternProperty(
-          property: TSESTree.RestElement | TSESTree.Property,
-        ): string[] {
-          switch (property.type) {
-            case AST_NODE_TYPES.RestElement:
-              return extractNamesFromPattern(
-                property.argument as DestructuringPattern,
-              )
-            case AST_NODE_TYPES.Property:
-              return extractNamesFromPattern(
-                property.value as DestructuringPattern,
-              )
-            /* v8 ignore next 2 -- @preserve Exhaustive guard. */
-            default:
-              throw new UnreachableCaseError(property)
-          }
-        }
-      }
-
       function extractDependencies(init: TSESTree.Expression): string[] {
         let dependencies: string[] = []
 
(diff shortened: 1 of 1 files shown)


rules/sort-objects.ts
```diff
@@ -144,6 +144,63 @@ export default createEslintRule<Options, MessageId>({
...
+        pattern: DestructuringPattern,
+      ): string[] {
+        if (pattern.type === 'AssignmentPattern') {
+          return extractNamesFromPattern(pattern.left as DestructuringPattern)
+        }
+
+        if (pattern.type === 'ObjectPattern') {
+          let names: string[] = []
+          for (let property of pattern.properties) {
+            if (property.type === 'Property') {
```
hugop95: @azat-io 

[nit] A `switch-case` on `property.type` with `UnreachableCaseError` is safer and more readable IMO (there are 2 cases today).

```ts
function extractNamesFromPattern(pattern: TSESTree.Node): string[] {]
  switch (pattern.type) {

    [...]

    case TSESTree.AST_NODE_TYPES.ObjectPattern:
      return pattern.properties.flatMap(extractNamesFromObjectPatternProperty)
    
     [...]

  function extractNamesFromObjectPatternProperty(
    property: TSESTree.RestElement | TSESTree.Property,
  ): string[] {
    switch (property.type) {
      case TSESTree.AST_NODE_TYPES.RestElement:
        return extractNamesFromPattern(property.argument)
      case TSESTree.AST_NODE_TYPES.Property:
        return extractNamesFromPattern(property.value)
      default:
        throw new UnreachableCaseError(property)
    }
  }
```

rules/sort-objects.ts
```diff
@@ -144,6 +144,63 @@ export default createEslintRule<Options, MessageId>({
         sourceCode,
       })
 
+      type DestructuringPattern =
+        | TSESTree.AssignmentPattern
+        | TSESTree.ObjectPattern
+        | TSESTree.ArrayPattern
+        | TSESTree.Identifier
+
+      function extractNamesFromPattern(
```
hugop95: @azat-io 

Could we put this function at the bottom ? We don't benefit from nesting it here, and it's harder to read IMO.

rules/sort-objects.ts
```diff
@@ -144,6 +144,63 @@ export default createEslintRule<Options, MessageId>({
...
+        pattern: DestructuringPattern,
```
hugop95: @azat-io 

`DestructuringPattern` doesn't bring type safety here as we cast all usages to `as DestructuringPattern`.
Typing `pattern: TSESTree.Node` is more explicit here IMO.

rules/sort-objects.ts
```diff
@@ -144,6 +144,63 @@ export default createEslintRule<Options, MessageId>({
...
+      ): string[] {
+        if (pattern.type === 'AssignmentPattern') {
```
hugop95: @azat-io 

I find `switch-casing` on `pattern.type` better. There are 4 cases:
- `AssignmentPattern`
- `ObjectPattern`
- `ArrayPattern`
- `Identifier` (the latest)
- `default` (unhandled).

rules/sort-objects.ts
```diff
@@ -144,6 +144,63 @@ export default createEslintRule<Options, MessageId>({
...
+                ...extractNamesFromPattern(
+                  property.argument as DestructuringPattern,
+                ),
+              )
+            }
+          }
+          return names
+        }
+
+        if (pattern.type === 'ArrayPattern') {
```
hugop95: @azat-io 

Same thing here: I think that a `switch-case` and a dedicated function makes it easier to read:

```ts
function extractNamesFromPattern(pattern: TSESTree.Node): string[] {
  switch (pattern.type) {

   [...]

    case TSESTree.AST_NODE_TYPES.ArrayPattern:
      return pattern.elements.flatMap(extractNamesFromArrayPatternElement)

    [...]

  function extractNamesFromArrayPatternElement(
    element: TSESTree.DestructuringPattern | null,
  ): string[] {
    if (!element) {
      return []
    }
    switch (element.type) {
      case TSESTree.AST_NODE_TYPES.RestElement:
        return extractNamesFromPattern(element.argument)
      default:
        return extractNamesFromPattern(element)
    }
  }
```

rules/sort-objects.ts
```diff
@@ -278,6 +335,16 @@ export default createEslintRule<Options, MessageId>({
             }
 
             let name = getNodeName({ sourceCode, property })
+            let dependencyNames = [name]
+            if (isDestructuredObject) {
+              let namesFromPattern = extractNamesFromPattern(
```
hugop95: @azat-io 

I think that we can simplify this to:

```ts
let dependencyNames = [name]
if (isDestructuredObject) {
  dependencyNames = [
    ...new Set(extractNamesFromPattern(property.value)),
  ]
}
```

We don't need to check if `namesFromPattern.length > 0`: in the previous version, we were overriding the original `dependencyName` if the object was destructured.

azat-io:
@hugop95 Thanks for the comments. I've refactored it.

PR #626, opened 2025-11-29
refactor: simplify internal typings

hugop95:
### Description

This PR changes the following:
- Put `PartitionByCommentOption` in `common-partition-options.ts`.
- Put `GroupNewlinesBetweenOption`, `NewlinesBetweenOption` and `GroupWithOverridesOption` in `common-groups-options`.
- Remove precise `Group`-specific types for each rule: 
  - All `Group` types end up with `"unknown" | string`, meaning that they don't bring any compilation safety.
    - `string` because custom group names can be defined as the user want. 
  - The only advantage is some IDE autocompletion in some cases, but the maintenance cost of such precise/complex types is no longer worth it in my opinion.
- Remove precise `SingleCustomGroup` typings.
  - Some rules defined a `AllowedModifiersPerSelector` type that defined what modifiers were allowed per selector. However, this isn't enforced at runtime, so the typing is purely informative. The maintenance cost of such precise/complex type is no longer worth it in my opinion. 
- Put `elementNamePattern` in the common options of `customGroups`, as it's used by all rules.

- [ ] Other

### DIFF
--- a/rules/sort-enums/types.ts
+++ b/rules/sort-enums/types.ts
@@ -35,7 +35,7 @@ export interface SortEnumsSortingNode
   value: string | null
 }
 
-/** Additional configuration for a single custom group.. */
+/** Additional configuration for a single custom group. */
 interface SingleCustomGroup {
   /**
    * Regular expression pattern to match enum member values. Members with values


azat-io (approved):
Wow, that's powerful. 👍

PR #627, opened 2025-11-29
feat(sort-imports): allow `type-import-first` fallback sort

### LINKED ISSUE #593: Feature: sort imports from the same module by "import type"

hugop95:
- Resolves #593

## Description

This PR adds the following `type` value to `sort-imports`: `type-import-first`.

The primary objective is to enforce fallback sorting between same `type` and `value` imports when the user decides not to group them apart.

```ts
import { antfu } from "@antfu/eslint-config"
import type { Awaitable } from "@antfu/eslint-config"
```

### ℹ️ Disclaimer

The primary use case for this feature is to be used with like so:
- `fallbackSort: { type: "type-import-first"}`
- Or 
  ```ts
  customGroups: [
    groupName: "group",
    fallbackSort: {
      type: "type-import-first" 
    }
  ]
  ```

However, the framework natively supports all other `type` override features:
- `type: "type-import-first"`
- `groups: [{ type: "type-import-first" }]`
- ```ts
  customGroups: [
    groupName: "group",
    type: "type-import-first",
  ]
  ```

Those are bad use cases in reality. Users should use `groups` instead with an alphabetical-based sort.
As such, they are not explicitly documented (but not forbidden).

- [x] New Feature

### DIFF
--- a/docs/content/rules/sort-imports.mdx
+++ b/docs/content/rules/sort-imports.mdx
@@ -169,7 +169,7 @@ When importing a type and value import from the same module, you can enforce the
 {
   type: 'alphabetical',
   order: 'desc',
-  fallbackSort: { type: 'type-import-first', order: 'desc' } // The type import will come before its value import
+  fallbackSort: { type: 'type-import-first', order: 'asc' } // The type import will come before its value import
 }
 ```
 


rules/sort-imports/comparator-by-options-computer.ts
```diff
@@ -0,0 +1,44 @@
...
+      throw new UnreachableCaseError(options.type)
+  }
+}
+
+function compareTypeImportFirst(
+  a: SortImportsSortingNode,
+  b: SortImportsSortingNode,
+  options: Pick<CommonOptions, 'order'>,
+): number {
+  if (a.isTypeImport && b.isTypeImport) {
```
azat-io, line 36: With `fallbackSort: { type: 'type-import-first' }` without explicit `order` (i.e., with the default `'asc'`), type imports end up after value imports. The option name suggests the opposite. Is this intentional?

hugop95: @azat-io 

Good point, I initially thought it made sense to place `type` import first when the order was `desc`, but re-looking at it, it's better to make it the wanted behavior with the default order (`asc`)

[`44442d3` (#627)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/627/commits/44442d3fa5469181e1e23be2dda5d7aac9ba62a5)

[`de5b44d` (#627)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/627/commits/de5b44d60f19dbdb6245e8a83dd4c0f344eb9dff)

Issue #571, opened 2025-08-14
Feature: Support ignoring objects by variable name in sort-objects


Gajdascz:
### What rule do you want to change?

sort-objects

### Describe the problem

Currently, there does not seem to be a way to ignore objects by variable name when using sort-objects. This makes working with order-dependent objects difficult.

### Code example

In `oclif` the args object is order-dependent. Sorting it ruins the positional input. 

```ts
import { Command, Args } from '@oclif/core';

/**
 * The order of args is always intentional and should not be sorted.
 */
export default class MyCommand extends Command {
  static override args = {
    type: Args.string({
      name: 'type',
      options: [type1,type2,type3],
      required: true
    }),
    description: Args.string({
      name: 'description',
      required: true
    }),
    scope: Args.string({
      name: 'scope',
      required: false
    })
  }
}
```

### Additional comments

This deprecated rule seems to state it's possible to ignore objects by it's name. But I was not able to get it to work and regardless it's deprecated.
- https://perfectionist.dev/rules/sort-objects#deprecated-ignorepattern

The alternative to the deprecated rule according to the docs is `useConfigurationIf` with `unsorted` type. But its two rules `allNamesMatchPattern` and `callingFunctionNamePattern` don't solve this issue. `allNamesMatchPattern` only matches object keys and I'm not using functions in this context.
- https://perfectionist.dev/rules/sort-objects#useconfigurationif

### Validations

- [x] Read the [docs](https://perfectionist.dev).
- [x] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

azat-io:
Please tell me how common this case is and why using eslint-disable comments is not appropriate?

Gajdascz:
I cannot speak to how common this is across all projects, but in CLI packages like those built with oclif, order-dependent objects (e.g., the args object) are non-negotiable. Sorting them breaks intended positional behavior, so workarounds are otherwise required.

Using eslint-disable comments is not ideal because, aside from being a _perfectionist_, they introduce repetitive, local escape hatches. This can make it harder to maintain consistent rules across multiple order-dependent objects and can allow subtle bugs to go unnoticed if a comment is missed. For nested objects, the problem compounds, as every child requires its own disable comment.

hugop95:
[...]

PR #632, opened 2025-12-07
feat(sort-objects)!: drop deprecated `ignorePattern` 

hugop95:
## Description

In favor of `useConfigurationIf.declarationMatchesPattern` and `useConfigurationIf.callingFunctionNamePattern` with `type: "unsorted"`.

A fix was needed for them to be able to detect complex expressions (such as castings), but they should now offer the same possibilities as `ignorePattern`.

- [x] Other

### DIFF
--- a/test/utils/get-settings.test.ts
+++ b/test/utils/get-settings.test.ts
@@ -25,7 +25,6 @@ describe('get-settings', () => {
       specialCharacters: 'keep',
       partitionByNewLine: true,
       type: 'alphabetical',
-      ignorePattern: [],
       ignoreCase: true,
       locales: 'en-US',
       alphabet: '',
--- a/utils/get-settings.ts
+++ b/utils/get-settings.ts
@@ -1,7 +1,7 @@
 import type { TSESLint } from '@typescript-eslint/utils'
 
 import type { PartitionByCommentOption } from '../types/common-partition-options'
-import type { CommonOptions, RegexOption } from '../types/common-options'
+import type { CommonOptions } from '../types/common-options'
 
 /**
  * Global settings for the Perfectionist plugin.
@@ -20,9 +20,6 @@ export type Settings = Partial<{
 
   /** Whether to create partitions at newlines. */
   partitionByNewLine: boolean
-
-  /** Pattern to match elements that should be ignored during sorting. */
-  ignorePattern: RegexOption
 }> &
   Partial<CommonOptions>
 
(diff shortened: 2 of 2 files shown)


hugop95:
Hi @azat-io 

Related to [this comment](https://github.com/azat-io/eslint-plugin-perfectionist/pull/569#issuecomment-3606695409): with this PR, I believe that we should be iso-functional to what existed before.

I still would like to improve things as explained in https://github.com/azat-io/eslint-plugin-perfectionist/issues/571#issuecomment-3196206473:

Currently,

```ts
const ignoreMe = {
  field: <any>{
    b,
    a,
    c,
  }
}
```

with `declarationMatchesPattern: "ignoreMe"` will not be ignored because `declarationMatchesPattern` will evaluate on `field`.

It would be useful to be able to do

```ts
{
  useConfigurationIf.declarationMatchesPattern: { pattern: "ignoreMe", scole: "deep" }
}
```

in order to match all sub-objects as well.

---

I don't think it should be too hard (I still think I can get something by the end of the week hopefully), but it's not blocking for the release and can always be added later if needed.

docs/content/rules/sort-object-types.mdx
```diff
@@ -209,16 +209,6 @@ type User = {
 
 ```
 
-### [DEPRECATED] ignorePattern
```
azat-io, line 212: Should we also update `docs/content/guide/getting-started.mdx:143-152`?
Currently, this page has an example with `ignorePattern`.

docs/content/rules/sort-object-types.mdx
```diff
@@ -721,7 +711,6 @@ type User = {
                   ignoreCase: true,
                   specialCharacters: 'keep',
                   sortBy: 'name',
-                  ignorePattern: [],
```
azat-io, line 724: We still have the `ignorePattern` validation. Is it necessary?
`utils/get-settings.ts:13-85`

hugop95: @azat-io 

Good catch 👍

[`81948b6` (#632)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/632/commits/81948b6baf2fa30b80dffafd8238d706004a1bc6)

docs/content/rules/sort-object-types.mdx
hugop95: [`81948b6` (#632)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/632/commits/81948b6baf2fa30b80dffafd8238d706004a1bc6)

PR #635, opened 2025-12-09
fix: honor declaration comments for nested types

azat-io:
### Description

Fix declarationCommentMatchesPattern so declaration comments on aliases/interfaces apply to nested/intersection object types (no fallback to properties) and add a regression test for nested cases.

### Additional context

N/A.

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves).

### DIFF
--- a/test/rules/sort-object-types.test.ts
+++ b/test/rules/sort-object-types.test.ts
@@ -2419,6 +2419,27 @@ describe('sort-object-types', () => {
         ],
         code: dedent`
           // Ignore me
+          interface Interface {
+            nested: {
+              b: string
+              a: string
+            }
+          }
+        `,
+      })
+
+      await valid({
+        options: [
+          {
+            useConfigurationIf: {
+              declarationCommentMatchesPattern: '^Ignore me$',
+            },
+            type: 'unsorted',
+          },
+          options,
+        ],
+        code: dedent`
+          // Ignore me
           type Type = A & {
             nested: {
               b: string
--- a/docs/content/rules/sort-object-types.mdx
+++ b/docs/content/rules/sort-object-types.mdx
@@ -352,6 +352,8 @@ Example configuration:
 
 - `declarationCommentMatchesPattern` — A regexp pattern to specify which comments above the type declaration should match.
 
+This setting also matches any object type found at any depth inside a type alias or interface that has a matching comment.
+
 Example configuration:
 ```ts
 {
--- a/test/rules/sort-interfaces.test.ts
+++ b/test/rules/sort-interfaces.test.ts
@@ -2404,6 +2404,28 @@ describe('sort-interfaces', () => {
         `,
       })
 
+      await valid({
+        options: [
+          {
+            useConfigurationIf: {
+              declarationCommentMatchesPattern: '^Ignore me$',
+            },
+            type: 'unsorted',
+          },
+          options,
+        ],
+        code: dedent`
+          // Ignore me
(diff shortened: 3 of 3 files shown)


rules/sort-object-types.ts
```diff
@@ -340,16 +348,35 @@ export function sortObjectTypeElements<MessageIds extends string>({
...
+  declarationMatchParent:
+    | TSESTree.TSTypeAliasDeclaration
+    | TSESTree.TSInterfaceDeclaration
+    | TSESTree.TSPropertySignature
+    | null
+} {
   let parentNodes = computeParentNodesWithTypes({
     allowedTypes: [
       AST_NODE_TYPES.TSTypeAliasDeclaration,
+      AST_NODE_TYPES.TSInterfaceDeclaration,
```
hugop95, line 366: @azat-io 

[test] I think we should add a dedicated `sort-object-types` test for that.

```ts
// Ignore me
interface Interface {
  nested: {
    b: string
    a: string
  }
}
```

This will work because `sort-object-types` considers interface comments.

test/rules/sort-object-types.test.ts
```diff
@@ -2399,9 +2399,31 @@ describe('sort-object-types', () => {
...
-            a: string
+            nested: {
+              b: string
+              a: string
+            }
+          }
+        `,
+      })
+
+      await valid({
```
hugop95, line 2431: @azat-io 

[test] Can we also add tests for `sort-interfaces` ?

rules/sort-interfaces.ts
```diff
@@ -27,6 +27,23 @@ type MessageId =
   | typeof ORDER_ERROR_ID
 
 export default createEslintRule<Options, MessageId>({
+  create: context => ({
```
hugop95, line 30: @azat-io 

[docs] Maybe we can add in the `sort-object-types` documentation that *this setting will also match any object type found at any depth within a type alias or interface that has a matching comment.*. The scope is **deep** by default here.

Similar to

https://github.com/azat-io/eslint-plugin-perfectionist/blob/472b4b4d7500c349a5235f308538e34ae8aac442/docs/content/rules/sort-objects.mdx?plain=1#L394

---

Note: I think that the scope should ideally be `shallow` by default (it's the default behavior for `sort-objects`'s `declarationMatchesPattern`).

This means that in theory, we should have

```ts
// Ignore me
type T = {
  b: "b"
  a: "a"
  nested: { // Sorting is enforced here, and `scope: "deep"` support should ideally be added
    a: "a"
    b: "b"
  }
}
```

However, this requires modifying a bit more things and I think that a first version with this behavior is acceptable as well.

azat-io: Done ✅

test/rules/sort-object-types.test.ts
azat-io: Done ✅

rules/sort-object-types.ts
azat-io: Done ✅

test/rules/sort-object-types.test.ts
hugop95: @azat-io 

Thanks ! I just realized that we actually had a test that does a similar thing already for `sort-interfaces` (we can't have nested interfaces anyways).

Issue #645, opened 2025-12-18
Bug: sort-exports rule now requires `@typescript-eslint/parser`


merrywhether:
### Describe the bug

It is mentioned in the contributing docs that `@typescript-eslint/parser` is used but not in the install/config docs. I had not had an issue with this ([using the "recommended-natural" config](https://github.com/merrywhether/config/blob/main/src/eslint/index.js#L23)) until v5, when something changed around the sort-exports rule (not sure if newly added to the config or the logic was updated) which caused the following error:

```sh
ESLint: 9.39.2

UnreachableCaseError: Unreachable case: undefined
Occurred while linting /home/runner/work/config/config/src/projen/index.js:1
Rule: "perfectionist/sort-exports"
at computeExportKindModifier (file:///home/runner/work/config/config/node_modules/.pnpm/eslint-plugin-perfectionist@5.0.0_eslint@9.39.2_typescript@5.9.3/node_modules/eslint-plugin-perfectionist/dist/rules/sort-exports.js:223:13)
```

I had only been enabling `@typescript-eslint/parser` for TS files before, falling back to Espree (ESLint's default) for JS files. But Espree doesn't yield [the referenced exportKind](https://github.com/azat-io/eslint-plugin-perfectionist/blob/2cfeed963061bdfc113a35c5cbd09251320dd386/rules/sort-exports.ts#L268-L280) on Node's ([espree example](https://ast-explorer.dev/#eNo9i0EKwjAQRa8yzCYgpe5b3HoCcZVNjCO0JJkwk4hSendTKt399x9vwYADzu7t1MuUC3aY20GahaiB3+CTWQqc4CUcwfRn189qRpv+wnPSAldmuIB5ODFjC7mFi00AFpWreLp9M1kcGkd+1tB2t2vy0d1JdOK0++AKaTl8YL/9RSrZtOL6A+3sOUQ=) vs [ts-parser example](https://ast-explorer.dev/#eNoty0EKwkAMQNGrhGwGROu+xa1HcJVNHCK0TJNhEkQpvbsjun2fv2HBERd+suc218Aj1g7xrvKDk3iZ9eu5u7yqtYADPJqtkIYzD4unifQfsqkHXM3gAunOLU19tD5upACEkle+SfPZlHDsUDjEg5B0x/0D0w8uCA==)) so I ran into the unreachable case error on my JS files.

### Code example

```
// example.js
// without ts-parser specified in ESLint config
[...]

PR #646, opened 2025-12-19
fix: handle missing export kind in sort exports

azat-io:
### Description

Handle missing export kind (undefined) as value for sort-exports.

### Additional context

#645 

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/rules/sort-exports.ts
+++ b/rules/sort-exports.ts
@@ -254,7 +254,7 @@ function sortExportNodes({
 function computeExportKindModifier(
   node: TSESTree.ExportNamedDeclaration | TSESTree.ExportAllDeclaration,
 ): 'value' | 'type' {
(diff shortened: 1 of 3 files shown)


rules/sort-named-exports.ts
```diff
@@ -117,10 +117,10 @@ export default createEslintRule<Options, MessageId>({
 
         let selector: Selector = 'export'
         let modifiers: Modifier[] = []
-        if (specifier.exportKind === 'value') {
-          modifiers.push('value')
-        } else {
+        if (specifier.exportKind === 'type') {
```
hugop95, line 120: @azat-io 

Could we add a test for this as well for this rule in case this gets converted similarly to `sort-exports` ?

azat-io: Done ✅

Issue #631, opened 2025-12-05
Feature: sort in order of use


abrahamguo:
### What rule do you want to change?

`classes`, `enums`, `imports`, `modules`,  `sort-objects` (in destructuring), `variable-declarations`

### Describe the problem

What do we think about adding an option to these rules to sort by usage (in the same file)?

Taking `variable-declarations` as an example, the variable declarations would be sorted so that the first-used variable is declared first, and so on.

We would probably consider usage only within the current scope, and not usage when something is exported and used in another file.

This should actually allow us to bypass dependency tracking in most cases, because this sort method naturally takes care of non-circular dependencies.

This was previously brought up in #434 , but that issue was quite different.

- In #434, this was discussed only for `modules`, whereas this issue would add this functionality to several different rules.
- The actual functionality of sorting by usage was only briefly discussed in #434, before that issue went off-topic, and began discussing other bugs and issues in `modules`.
- #434 was closed after some minor bugs in `modules` were fixed, but the actual functionality of having an option to sort by usage wasn't implemented. 

### Code example

Correct sort order:

```
const b = 2, a = 1, c = b + a;

console.log(c);
console.log(a, b);
```

`b` and `a` are first used on line 1, while `c` is first used on line 3.
`b` is used before `a`, so it is sorted first.

Note that we didn't even need to determine that `c` depends on `a` and `b` — applying this sort order automatically and naturally takes care of dependencies!

### Validations

- [x] Read the [docs](https://perfectionist.dev).
- [x] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

hugop95:
@abrahamguo 

Thank you for putting this proposal back on the table. I definitely think that this would be a great improvement. However, I think that the main use case still lies with `sort-modules`, and that efforts should primarily be put here at first.

Things have improved a bit since the opening of https://github.com/azat-io/eslint-plugin-perfectionist/issues/434:
- We support fallback sorting.
- Our framework allows us to define custom sorting algorithms for specific rules more easily.

I'll see if I can get a convincing POC up.

azat-io:
Added in fc297a2d71dabc97228735d557c219cd72260db9.
Released in [v5.2.0](https://github.com/azat-io/eslint-plugin-perfectionist/releases/tag/v5.2.0).

Happy New Year! ⛄️

PR #648, opened 2025-12-20
feat(sort-modules): sort by order of use

hugop95:
- Resolves https://github.com/azat-io/eslint-plugin-perfectionist/issues/631.
- Resolves https://github.com/azat-io/eslint-plugin-perfectionist/issues/434.

### Description

This PR adds the following `type` option to `sort-modules`: `'usage'`. This sorting type sorts items by their usage in their group. Items not used in their group can be placed anywhere.

This is essentially an extension of the **dependency sorting** strategy that already exists to ensure compilation is preserved.

This sorting type will need to be tested in real use cases a bit more to see what can be improved and detect potentially missing detection cases.

- [x] New Feature

### DIFF
--- a/test/rules/sort-modules.test.ts
+++ b/test/rules/sort-modules.test.ts
@@ -1735,6 +1735,52 @@ describe('sort-modules', () => {
             },
           ],
           code: dedent`
+            const a = (...B) => {}
+
+            type B = 'b'
+          `,
+        })
+
+        await valid({
+          code: dedent`
+            class A {
+              a = (...B) => {}
+            }
+
+            type B = 'b'
+          `,
+          options: [
+            {
+              ...options,
+              groups: ['unknown'],
+            },
+          ],
+        })
+
+        await valid({
+          code: dedent`
+            class A {
+              a(...B) {}
+            }
+
+            type B = 'b'
+          `,
+          options: [
+            {
+              ...options,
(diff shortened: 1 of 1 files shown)


azat-io:
Looks good overall. One suggestion: add a test to avoid false positives from parameter/identifier bindings.

```ts
it('ignores parameter bindings that shadow module-level names', async () => {
  await valid({
    code: dedent`
      function foo(B: number) {}

      type B = { value: number }
    `,
    options: [
      {
        ...options,
        groups: ['unknown'],
      },
    ],
  })
})
```

hugop95:
@azat-io 

Thank you for the feedback. I've managed to improve things a bit, but unfortunately, it turns out that it's harder to handle all edge cases. Example of enforced cases that shouldn't be enforced:

```ts
type B = 'b'

function a(...B) {}
```
⬆️ *Spread arguments aren't detected as independent.*

```ts
type B = 'b'

type A<B> = void
```
⬆️ *Generics aren't detected as independent.*

For the second case for example, it's not trivial to differentiate between generic types and external types.

On one hand, the current implementation doesn't have a perfect detection, but on the other hand, I hope it should fit the large majority of use cases.

To be honest, I'm not sure if we'll ever be able to simply detect everything correctly.

azat-io:
Thanks, looks good! 👍

PR #651, opened 2025-12-21
feat: add `newlinesInside` at rule options and settings level

### LINKED ISSUE #652: Feature: Add `newLinesInside` at rule options and settings levels

All rules + Settings

### Describe the problem

Following of https://github.com/azat-io/eslint-plugin-perfectionist/issues/649

Would be great to be able to change default `newLinesInside` (`'ignore'`) at the rule and setting levels – would apply by default to:
- for all groups if set at the settings level
- for all groups for a given rule if set at the rule level

This would then apply with the following priority:
- group level
- rule level
- setting level
- default

### Code example

```
  …
  settings: {
    perfectionist: {
      newLinesInside: 0,
    }
  }
  …
```

### Additional comments

_No response_

### Validations

hugop95:
- Resolves #652.

## Description

This PR adds `newlinesInside` to all rules that support grouping (all except `sort-switch-case`).

## ⚠️ Breaking change

Previously, `newlinesBetween` different than `ignore` meant that no newline was tolerated by default within groups, as per the documented:

Screenshot 2025-12-21 at 20 45 05

Now, as `newlinesInside` is `ignore` by default for all rules but `sort-imports` (which has `newlinesInside: 0` to be isofunctional with its default `newlinesBetween: 1`), errors won't be raised anymore if a newline exist within groups.

This means that users will have to manually add `newlinesInside: 0` to their rule options.

- [x] New Feature

### DIFF
--- a/test/rules/sort-array-includes.test.ts
+++ b/test/rules/sort-array-includes.test.ts
@@ -1288,21 +1288,7 @@ describe('sort-array-includes', () => {
       },
     )
 
-    it('removes newlines between and inside groups', async () => {
-      let newlinesOptions = [
-        {
-          ...options,
-          customGroups: [
-            {
-              elementNamePattern: 'a',
-              groupName: 'a',
-            },
-          ],
-          groups: ['a', 'unknown'],
-          newlinesBetween: 0,
-        },
-      ]
-
+    it('removes newlines between and inside groups by default when "newlinesBetween" is 0', async () => {
       await invalid({
         errors: [
           {
@@ -1316,6 +1302,19 @@ describe('sort-array-includes', () => {
           {
             messageId: 'extraSpacingBetweenArrayIncludesMembers',
             data: { right: 'b', left: 'z' },
+          },
+        ],
+        options: [
+          {
+            ...options,
+            customGroups: [
+              {
+                elementNamePattern: 'a',
+                groupName: 'a',
+              },
+            ],
+            groups: ['a', 'unknown'],
+            newlinesBetween: 0,
           },
         ],
         code: dedent`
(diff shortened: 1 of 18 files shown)


azat-io:
Thanks for the feature!

It looks useful.

However, this changes the default behavior: with `newlinesBetween: 1/2/...`, empty lines inside groups are no longer prohibited unless `newlinesInside: 0` is explicitly set. This makes it a breaking change, and since v5.0.0 was just released, we'd prefer to avoid another major bump.

Would it be possible to preserve backward compatibility by making `newlinesInside` derive from `newlinesBetween` when not explicitly set? For example: `newlinesInside` defaults to `0` when `newlinesBetween` is a number, and to `'ignore'` when `newlinesBetween === 'ignore'`. 

Users who want the new behavior could then explicitly set `newlinesInside: 'ignore'`.

hugop95:
@azat-io 

I think that the breaking change was minor considering that the release was delivered very recently and that the breaking change only "removes an enforced behavior" (and doesn't enforce an other).

But we can support both.

The default value for `newlinesInside` is now `"newlinesBetween"`, except for `sort-imports` which has `newlinesInside: 0` without any breaking change.

azat-io:
Should we also update `docs/content/guide/getting-started.mdx`?

hugop95:
@azat-io 

> Should we also update docs/content/guide/getting-started.mdx?

Good catch 👍 [`7c66ea6` (#651)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/651/commits/7c66ea6b50faeb3b6e9026df35d927d686d486c4)

OlivierZal:
Awesome, thanks @hugop95 and @azat-io!

PR #659, opened 2025-12-29
feat: add `options.newlinesAfter` to `sort-imports` and `sort-exports` 

### LINKED ISSUE #658: Feature(sort-imports): enforce newline after last import

sort-imports

### Describe the problem

Hi, first of all thanks for this awesome plugin!

I'd like to propose a new option to the `sort-imports` rule to enforce a new line after the last import. Since Prettier doesn't enforce this either, you can end up with an import and export sticking together. I like to start typing in a new file without caring about formatting and style, and have Prettier and ESLint clean everything up for me once I hit save. This missing rule has always been a small annoyance for me.

I looked around and found that `eslint-plugin-import` does have such a rule: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/newline-after-import.md. Personally I don't think a configurable amount of newlines is necessary, Prettier will collapse them anyway (and I think this plugin is a great _addition_ to Prettier). A simple boolean would work for me.

### Code example

**Before**
```ts
import { URL } from 'node:url';

import { baseUrl } from './constants.ts';
export const createExampleUrl = (path: string) => new URL(path, 'https://example.org/');
```
**After**
```ts
import { URL } from 'node:url';

import { baseUrl } from './constants.ts';

export const createExampleUrl = (path: string) => new URL(path, 'https://example.org/');
```

### Additional comments

_No response_

### Validations

hugop95:
- Resolves #658

### Description

TODO

- [x] New Feature

### DIFF
--- a/rules/sort-exports/types.ts
+++ b/rules/sort-exports/types.ts
@@ -16,7 +16,9 @@ import {
  */
 export type Options = Partial<
   {
-    /** Number of newlines required after the last sorted element. */
+    /**
+     * Number of newlines required after the last sorted element.
+     */
     newlinesAfter: 'ignore' | number
   } & AllCommonOptions<
     TypeOption,
--- a/rules/sort-imports/types.ts
+++ b/rules/sort-imports/types.ts
@@ -40,7 +40,9 @@ export type Options = Partial<
      */
     useExperimentalDependencyDetection: boolean
 
-    /** Number of newlines required after the last sorted element. */
+    /**
+     * Number of newlines required after the last sorted element.
+     */
     newlinesAfter: 'ignore' | number
 
     /**
--- a/utils/get-newlines-after-errors.ts
+++ b/utils/get-newlines-after-errors.ts
@@ -14,19 +14,29 @@ interface GetNewlinesBetweenErrorsParameters<
   MessageIds extends string,
   T extends SortingNode,
 > {
-  /** Configuration options for newlines. */
+  /**
+   * Configuration options for newlines.
+   */
   options: { newlinesAfter: 'ignore' | number }
 
-  /** ESLint source code object for accessing lines. */
+  /**
+   * ESLint source code object for accessing lines.
+   */
   sourceCode: TSESLint.SourceCode
 
-  /** Error message ID for missing required newlines. */
+  /**
+   * Error message ID for missing required newlines.
+   */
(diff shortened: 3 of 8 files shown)


azat-io:
I have some doubts about this option. It seems to be out of scope for the current plugin. What do you think about it?

hugop95:
@azat-io 

> I have some doubts about this option. It seems to be out of scope for the current plugin. What do you think about it?

I think it's in a grey area. In my opinion, the pros/cons are:
- ✅ We already support `newlinesBetween` and `newlinesInside`, so it can make sense to support `newlinesAfter` (or `newlinesBefore`).
- ❌ It's no longer "inside" or "between" sorted elements, but after.

@dd-jonas Can you use both `perfectionist` and [`eslint-plugin-imports/newline-after-import`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/newline-after-import.md) at the same time to solve your issue ?

azat-io:
I want to create a new release today. Shall we wait for @dd-jonas's response? And should I wait for this PR?

dd-jonas:
@azat-io @hugop95  Hi, thanks for mentioning me and considering this so quickly.

For me, this is not important enough to justify using an extra plugin such as `eslint-plugin-imports/newline-after-import`. It is more about this plugin rule being more feature complete.

I agree this is a grey area, but I do believe this fits within the `sort-imports` rule (and `sort-exports`):
- This style rule is about the newline after the last *import*. It is not a generic rule about newlines between blocks of code, but it directly relates to the import lines.
- `sort-imports` already has some configuration options that are not related to the sort order. Grouping and newlines between groups have no influence on the sort order (unless you apply different sorting rules per group), but they are necessary to visually clean up the imports. A final newline after the last import group follows the same principal in this way.

@azat-io This is not urgent for me at all. I wouldn't block a release because of this PR, if this needs to be discussed further.

And thanks for your work creating and maintaining this great plugin!

hugop95:
@azat-io 

> And should I wait for this PR?

I agree that I don't think it's a key feature that would justify delaying the release.

hugop95:
I'm going to close this PR for now as there are mixed opinions about it fitting the scope. I can reopen if a clear decision to support it is chosen. Re-looking at it, I also believe that although it would make the plugin feature complete, it's not really in the scope of sorting elements itself.

Issue #560, opened 2025-07-17
Feature: Sort imports by specifiers


blakegearin:
### Describe the rule

I feel like I must be missing this somewhere in the settings or past issues. I'd like to sort my imports by specifiers (export names) instead of the paths they're coming from.

### Code example

```js
import { A } from 'z'
import { Z } from 'a'
import { a } from 'y'
import { z } from 'b'

import A from 'z.js'
import Z from 'a.js'
import a from 'y.js'
import z from 'b.js'
```

```js
import type { Response, Request } from 'express'

import bodyParser from 'body-parser'
import defaultsDeep from 'lodash/defaultsDeep'
import express from 'express'
import fs from 'node:fs/promises'
import map from 'lodash/map'
import mongoose from 'mongoose'
import passport from 'passport'
import path from 'node:path'
import session from 'express-session'

import User from '~/models/User'
import authRoutes from '~/routes/auth'
import initializePassport from '~/config/passport'
import logger from '~/middleware/logger'

import dbConfig from './db'
```

### Validations

- [x] Read the [docs](https://perfectionist.dev).
- [x] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

azat-io:
Thank you for your issue.

At the moment, we do not support this. How important do you think this issue is, and what do you think the API should look like?

hugop95:
@blakegearin cc @azat-io 

What should be the behavior when an import does not have a specifier/has multiple specifiers ?
- `import something`
- `import { a, z } from "something"` ➡️ Should we evaluate `a` or `z` ?
- Same for `import { z, a } from "something"`

How could we make this non-opinionated ?

In my opinion, the most straightforward implementation would be to add a `sortBy: 'specifier' | 'path'` (default `path`) option, similar to the [`sortByValue: boolean` of `sort-enums`](https://perfectionist.dev/rules/sort-enums#sortbyvalue).

azat-io:
I'm not sure if this is a common type of sorting.

On the other hand, we already have similar functionality in sort-enums.

blakegearin:
This is the way the ESLint [sort-imports](https://eslint.org/docs/latest/rules/sort-imports) rule works. Here's some that's valid with their defaults.

```js
// 'sort-imports': 'error',

import 'module-without-export.js';
import * as bar from 'bar.js';
import * as foo from 'foo.js';
[...]

PR #660, opened 2025-12-30
feat(sort-imports): allow sorting by specifiers

hugop95:
- Resolves https://github.com/azat-io/eslint-plugin-perfectionist/issues/560.

### Description

This PR adds a `sortBy: 'path' | 'specifier'` option (by default, `path`) to `sort-imports`, similar to [this rule](https://eslint.org/docs/latest/rules/sort-imports) (it differs a bit in the result, especially regarding lowercase/uppercase sorting).

It considers the **first specifier** encountered if there are multiple ones.

When used with `type: 'line-length'`, it changes nothing.

### Example

```ts
import 'foo4'
import {} from 'foo10';
const {} = require('foo11')
const [] = require('foo13')
import a from 'foo8'
import { c, l } from 'foo7'
import { default as d } from 'foo1'
import e = Foo.e
import * as f from 'foo5'
const { ...g } = require('foo3')
import h from 'foo2'
const [i] = require('foo12')
const j = require('foo6')
import { b as k } from 'foo9'
```

- [x] New Feature

### DIFF
--- a/test/rules/sort-imports.test.ts
+++ b/test/rules/sort-imports.test.ts
@@ -3552,25 +3552,30 @@ describe('sort-imports', () => {
             messageId: 'unexpectedImportsOrder',
           },
           {
-            data: { right: 'foo13', left: 'foo12' },
+            data: { right: 'foo11', left: 'foo10' },
+            messageId: 'unexpectedImportsOrder',
+          },
+          {
+            data: { right: 'foo14', left: 'foo13' },
             messageId: 'unexpectedImportsOrder',
           },
         ],
         output: dedent`
           import 'foo4'
-          import {} from 'foo10';
-          const {} = require('foo11')
-          const [] = require('foo13')
+          import {} from 'foo11';
+          const {} = require('foo12')
+          const [] = require('foo14')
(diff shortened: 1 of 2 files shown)


test/rules/sort-imports.test.ts
```diff
@@ -3527,6 +3527,77 @@ describe('sort-imports', () => {
...
+          import a from 'foo8'
+          import { c, l } from 'foo7'
+          import { default as d } from 'foo1'
+          import e = Foo.e
+          import * as f from 'foo5'
+          const { ...g } = require('foo3')
+          import h from 'foo2'
+          const [i] = require('foo12')
+          const j = require('foo6')
+          import { b as k } from 'foo9'
```
azat-io: What do you think about adding something like this?

```ts
const { foo: bar } = require('x')
```

hugop95: @azat-io 👍

[`e7927d2` (#660)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/660/commits/e7927d2ca89392f5cbb42789a9f9a8e781876792)

Issue #666, opened 2026-01-02
Bug: `sort-modules` by `usage` doesn't seem to work


aguo-ntech:
### Describe the bug

The new `usage` option for `sort-modules` (introduced by @hugop95 in #648) i documented as:

> Sort items by the order they are first used or referenced within their [group](https://perfectionist.dev/rules/sort-modules#groups).

Therefore, in this code example, I would expect that `A` would be sorted before `B`, since `A` is used/referenced before `B`. However, the plugin does not report any issues, and is happy with the current sorting order.

### Code example

```
type B = 1;

type A = 2;

type X = A | B;

### ESLint version

9.39.2

### ESLint Plugin Perfectionist version

5.2.0

### Additional comments

_No response_

### Validations

- [x] Read the [docs](https://perfectionist.dev).
- [x] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

hugop95:
Hi @aguo-ntech 

The current behavior is intended: `usage` enforces that a module element A (function, type, interface, class, enum) is placed before another module element B if B references A. However, it doesn't enforce anything else.

This sort option is necessarily opinionated, and in my opinion, enforcing what you suggest feels like too much, we should at least give the possibility to users to enforce lenient usage sorting while still giving them the freedom to sort first-used elements as they want (alphabetically, for example).

We can first work on documenting the option better to avoid confusion, and in a second time think about upgrading the option to support stricter enforcement.

aguo-ntech:
Ah.

In that case, is it correct that this `usage` option is identical to the `no-use-before-define` rule from [ESLint](https://eslint.org/docs/latest/rules/no-use-before-define) and [typescript-eslint](https://typescript-eslint.io/rules/no-use-before-define/)?

> This sort option is necessarily opinionated, and in my opinion, enforcing what you suggest feels like too much, we should at least give the possibility to users to enforce lenient usage sorting while still giving them the freedom to sort first-used elements as they want (alphabetically, for example).

> in a second time think about upgrading the option to support stricter enforcement.

I definitely think that this would be a good idea.

Everything else in `eslint-plugin-perfectionist` has a strict and specific sort order — this entire ESLint plugin is about being very opinionated — so I was very surprised that this `usage` option is _not_ opinionated.

hugop95:
@aguo-ntech 

[...]

PR #667, opened 2026-01-02
fix(sort-modules): fix `type: 'usage'` not handling `fallbackSort`

hugop95:
- Related issue: https://github.com/azat-io/eslint-plugin-perfectionist/issues/666

## Description

`type: 'usage'` currently doesn't handle `fallbackSort` correctly, because this comparator never returns `0`. This PR fixes this issue.

It also improves the documentation for this option to reduce confusion.

## Important note

There has been issues with dependency sorting before, and I'm wondering if I can't find a counter-example where the current implementation in this PR actually doesn't work. I want to spend more time investigating this before opening.

Edit: See https://github.com/azat-io/eslint-plugin-perfectionist/pull/667#discussion_r2669701543.

- [x] Bug fix

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## v5.3.1
+
+[compare changes](https://github.com/azat-io/eslint-plugin-perfectionist/compare/v5.3.0...v5.3.1)
+
+### 🐞 Bug Fixes
+
+- Align plugin configs typing with eslint ([7c44de3](https://github.com/azat-io/eslint-plugin-perfectionist/commit/7c44de3))
+
+### ❤️ Contributors
+
+- Azat S. ([@azat-io](https://github.com/azat-io))
+
 ## v5.3.0
 
 [compare changes](https://github.com/azat-io/eslint-plugin-perfectionist/compare/v5.2.0...v5.3.0)
--- a/index.ts
+++ b/index.ts
@@ -24,7 +24,10 @@ import sortEnums from './rules/sort-enums'
 import sortMaps from './rules/sort-maps'
 import sortSets from './rules/sort-sets'
 
-interface PluginConfigs {
+interface PluginConfigs extends Record<
+  string,
+  Linter.LegacyConfig | Linter.Config[] | Linter.Config
+> {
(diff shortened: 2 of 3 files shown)


azat-io:
I would like to create a release, we had an important bug fix:
https://github.com/azat-io/eslint-plugin-perfectionist/pull/670

At the same time, this pull request also seems important to me. Please advise, should I create a release today or wait for this pull request?

hugop95:
@azat-io 

> At the same time, this pull request also seems important to me. Please advise, should I create a release today or wait for this pull request?

Please don't wait for this PR, I still need to test additional things as I suspect there might be edge cases that are not covered.

rules/sort-modules/build-usage-comparator.ts
```diff
@@ -26,28 +34,63 @@ export function buildUsageComparator({
...
+      options.order,
+    )
+
+    if (sortedOrderedValue !== unsortedOrderedValue) {
+      return sortedOrderedValue
+    }
+
+    let aWithUpdatedDependencies = updatedSortingNodeByNode.get(nodeA)!
+    let bWithUpdatedDependencies = updatedSortingNodeByNode.get(nodeB)!
+    if (
```
hugop95, line 66: @reviewers

A part of me is not convinced that this will catch all cases, but I couldn't find a counter-example.

Dependency sorting is necessary topological, and we can't replicate it perfectly just with JavaScript's `sort`. This is why sorting by dependencies is always done at the end and [isn't based on `sort`](https://github.com/azat-io/eslint-plugin-perfectionist/blob/271a4811f1cb0f6ca3f4fafc64bf23f87c305516/utils/sort-nodes-by-dependencies.ts).

Here, we add a `fallback` sort after sorting by usage (which is essentially sorting by dependencies), so I'm afraid that there might be edge cases with transitive dependencies between nodes that might lead to issues.

I think it's acceptable to go with the current implementation and see where it goes, as it's quite a niche use case. The only way to really know if this algorithm works 100% of the time is to see if somebody can make it fail and reports it.

hugop95: @coderabbitai

azat-io:
Thanks!

Issue #669, opened 2026-01-02
Bug: crashes without @typescript-eslint/types


lishaduck:
### Describe the bug

https://github.com/azat-io/eslint-plugin-perfectionist/blob/5ca53c2a85697dbd26a3a0d6c80e7d19d13fa7d8/rules/sort-named-imports.ts#L3

In pnpm, when using a hermetic `node_modules` (`hoist: false`), or the global virtual store, eslint-plugin-perfectionist crashes because there's a runtime import of an undeclared dependency.

Historically, I would've said you needed to add a runtime dependency to satiate typescript, but I recently learned a trick to make TS happy: [example](https://www.typescriptlang.org/play/?#code/KYOwrgtgBAYg9nKBvAUFdUBCBDATlAXigEYAaNDHAL0KgCYUBfFFAYzhAGcAXKAQQBcsBLWJRO2bgEtOAMynBOwuADocuNhx5Yh8RETETpchUr1rsVFEA). Hopefully that's helpful!

### Code example

n/a

### ESLint version

n/a

### ESLint Plugin Perfectionist version

5.1.0

### Validations

- [x] Read the [docs](https://perfectionist.dev).
- [x] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

azat-io:
Fixed in a0ec11080e115d9c264ecb9439fa35559ea4596f.
Released in [v5.3.0](https://github.com/azat-io/eslint-plugin-perfectionist/releases/tag/v5.3.0).

PR #670, opened 2026-01-02
fix: always use `@typescript-eslint/utils` for `AST_NODE_TYPES`

hugop95:
- Fixes https://github.com/azat-io/eslint-plugin-perfectionist/issues/669.

### Description

We sometime use `@typescript-eslint/types` for `AST_NODE_TYPES` as a non-type import. However, we only have `@typescript-eslint/utils` as a `dependency`.

This PR fixes all usages to use `AST_NODE_TYPES` from `@typescript-eslint/utils`.

https://github.com/azat-io/eslint-plugin-perfectionist/blob/783001b333c9550ae2f8c3ff6718aea9a4c1cdbc/package.json#L60

- [x] Bug fix

### DIFF
--- a/rules/sort-jsx-props.ts
+++ b/rules/sort-jsx-props.ts
@@ -3,9 +3,8 @@ import type { TSESLint } from '@typescript-eslint/utils'
 
 import { AST_NODE_TYPES } from '@typescript-eslint/utils'
 
-import type { Modifier, Selector } from './sort-jsx-props/types'
+import type { Modifier, Selector, Options } from './sort-jsx-props/types'
 import type { SortingNode } from '../types/sorting-node'
-import type { Options } from './sort-jsx-props/types'
 
 import {
   buildUseConfigurationIfJsonSchema,


azat-io:
Back in August 2023 (commit f5ca1d7), we deliberately moved away from importing `AST_NODE_TYPES` constants to using string literals instead.

The main reasons were:

- Smaller bundle size (no runtime constant imports)
- Slightly better performance
- Cleaner code with fewer dependencies

TypeScript still provides type checking for us, so we're protected from typos at compile time.

What do you think about that alternative?

hugop95:
@azat-io 

> TypeScript still provides type checking for us, so we're protected from typos at compile time.

Typescript actually doesn't handle string vs enums values in the same way at compilation time:

```ts
let objectType = object.type
switch (objectType) {
  case AST_NODE_TYPES.ObjectExpression:
    break
  case AST_NODE_TYPES.ObjectPattern:
    break
  default:
    throw new UnreachableCaseError(objectType)
}
```

 will work but

```ts
let objectType = object.type
switch (objectType) {
  case 'ObjectExpression':
    break
  case 'ObjectPattern':
    break
  default:
    throw new UnreachableCaseError(objectType)
}
```

will actually raise an error at `throw new UnreachableCaseError(objectType)`.

image

---

At the moment, we already have a runtime dependency for `@typescript-eslint/utils`

https://github.com/azat-io/eslint-plugin-perfectionist/blob/a285b33177bc19aca83a9d78bb87a95acffb99ad/package.json#L60

because of

https://github.com/azat-io/eslint-plugin-perfectionist/blob/a285b33177bc19aca83a9d78bb87a95acffb99ad/utils/get-node-range.ts#L93

So in my opinion, if we really want to have a relevant impact, we should also look to remove `@typescript-eslint/utils` from runtime dependencies (and copy the `ASTUtils.isParenthesized` function for example).

---

I think we should also keep the `UnreachableCaseError` pattern as it provides good compilation safety. The good news is that removing `AST_NODE_TYPES` while still keeping `UnreachableCaseError` can be done, but only without a `switch`:

```ts
    if (object.type === 'ObjectExpression') {
      // ...
    } else if (object.type === 'ObjectPattern') {
      // ...
    } else {
      throw new UnreachableCaseError(object)
    }
```

However:
- ESLint will complain about the last `else if`
  image
  
  - So we will need to `eslint-disable` it each time we use this pattern.
- I find it a bit less readable than a `switch/case`.

In my opinion, removing the dependency to `@typescript-eslint/utils` can be interesting, but I don't think it's a huge deal either: 
- The [package](https://www.npmjs.com/package/@typescript-eslint/utils) size is small (196 kB).
- We need to plan on replacing the remaining usages in `getNodeRange` first before a relevant size impact can be seen.
- I estimate that the large majority of users indirectly have this dependency as long as they use typescript (as they likely use `typescript-eslint` as well), so removing this dependency would only impact JS users.

---

So overall, I personally would go with:
- Keep using `AST_NODE_TYPES` from `@typescript-eslint/utils` everywhere for now as it's been a runtime dependency since at least May 2023.
  - With this [new lint rule](https://github.com/azat-io/eslint-plugin-perfectionist/pull/670/commits/84ea435491dac63cc6d959ad1491b401346998e5), we at least make sure we don't directly import `@typescript-eslint/types`.
- In a second time, investigate if it's interesting to completely remove `@typescript-eslint/types` by replacing all `AST_NODE_TYPES` usages (will the readability cost be worth it ?) **and** finding a way to replace `ASTUtils.isParenthesized`. 
  - At this stage, we should also make sure to never directly import `@typescript-eslint/utils` other than for types.

azat-io (approved):
Thank you for your detailed explanation!

PR #673, opened 2026-01-04
feat: fallback sort subgroup order

Lievesley:
### Description

Adds explicit support and test coverage for fallbackSort: "subgroup-order" so subgroup ordering is applied consistently as the fallback sorting strategy, with focused comparator and group-order tests to validate behavior.

### Additional context

This makes public-get-method come before public-set-method when they otherwise tie under the primary sort:

```json
'perfectionist/sort-classes': [
  'error',
  {
    type: 'alphabetical',
    fallbackSort: { type: 'subgroup-order' },
    groups: [
      ['public-get-method', 'public-set-method'],
    ],
  },
]
```
---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -4,6 +4,3 @@ configDependencies:
 onlyBuiltDependenciesFile: node_modules/.pnpm-config/@pnpm/plugin-trusted-deps/allow.json
 
 sideEffectsCache: false
-
-packages:
-  - .


test/rules/sort-enums.test.ts
```diff
@@ -5710,4 +5716,60 @@ describe('sort-enums', () => {
       })
     })
   })
+
+  describe('comparatorByOptionsComputer', () => {
```
hugop95: @Lievesley 

I have two blocking points here:

### Point 1

This is not a fitting test here (same thing for `sort-imports`, `sort-modules` and `sort-object-types`) as you mock nodes (and tests here don't mock them). 

It would go in a `buildComparatorByOptionsComputer.test` file associated to `buildComparatorByOptionsComputer.ts`, but the interest of unit testing this like that is limited, so I don't recommend that either. It feels like it was just added to satisfy coverage, and that's sweeping dust under the rug:

In theory, it's better to test the full behavior. In practice, because it's not a critical feature for those rules, it's better to:
- Remove the tests you added.
- Add a `/* v8 ignore next X -- @preserve Untested for now as not a relevant sort for this rule. */` where needed until a proper test is added for all rules (because all rules now support this type). 

### Point 2

There is no test for the actual use case here in `sort-classes`. We should at least add the following test:

```ts
    it('allows grouping and sorting getters and setters', async () => {
      await invalid({
        options: [
          {
            ...options,
            type: 'alphabetical',
            fallbackSort: {type: 'subgroup-order'},
            groups: [['get-method', 'set-method'], 'unknown'],
          },
        ],
        errors: [
          // ...
        ],
        output: dedent`
          class Class {
            get a() {}
            set a() {},
            get b() {}
            set b() {}
          }
        `,
        code: dedent`
          class Class {
            set b() {}
            set a() {},
            get b() {}
            get a() {}
          }
        `,
      })
    })
```

test/utils/compare/default-comparator-by-options-computer.test.ts
```diff
@@ -0,0 +1,32 @@
+import { describe, expect, it } from 'vitest'
+
+import type { SortingNode } from '../../../types/sorting-node'
+
+import { defaultComparatorByOptionsComputer } from '../../../utils/compare/default-comparator-by-options-computer'
```
hugop95: @Lievesley 

Same comment as above, I don't think it's an interesting test overall, because it doesn't test the interesting part (that it returns the correct value depending on `groups` subarrays).

You can remove this file and add a proper test in `sort-classes` (see comment above), and coverage should be guaranteed for `default-comparator-by-options-computer`.

utils/sort-nodes-by-groups.ts
```diff
@@ -141,9 +148,28 @@ export function sortNodesByGroups<
 
     sortedNodes.push(
       ...sortNodes({
+        comparatorByOptionsComputer: optionsToCompare => {
```
hugop95: @Lievesley 

We shouldn't do that ! `comparatorByOptionsComputer` shouldn't be overridden in this file (I think we don't need to touch this file at all).

We should instead create a dedicated `buildSubgroupOrderComparator` comparator (like `buildLineLengthComparator`) that given the groups and order, returns a comparator that:
- Checks if `a` and `b` is within a same subgroup array.
  - If not, return 0.
- Otherwise, return 1 or -1 based on the order in the subgroup and the `order` value.

Then we can use this comparator in `defaultComparatorByOptionsComputer`.

docs/content/rules/sort-imports.mdx
```diff
@@ -173,6 +173,17 @@ When importing a type and value import from the same module, you can enforce the
 }
 ```
 
+You can also sort by subgroup order inside composite groups using `subgroup-order`.
```
hugop95: @Lievesley 

I'd rather not add this documentation until this is fully tested for this rule (and add it in `sort-classes.mdx` instead).

Lievesley:
@hugop95 I made an update, if you could please confirm this fits your overall idea that would be great.

utils/sort-nodes-by-groups.ts
Lievesley: Should be resolved now.

test/utils/compare/default-comparator-by-options-computer.test.ts
Lievesley: removed

test/rules/sort-enums.test.ts
Lievesley: removed and added sort-classes test

docs/content/rules/sort-imports.mdx
hugop95: @Lievesley cc https://github.com/azat-io/eslint-plugin-perfectionist/pull/673/files#r2659628053

Can you update `sort-classes.mdx` instead ?

test/rules/sort-classes.test.ts
```diff
@@ -1835,6 +1835,36 @@ describe('sort-classes', () => {
       })
     })
 
+    it('sorts getters and setters using subgroup-order fallback', async () => {
```
hugop95, line 1838: @Lievesley 

I think we can make this test a bit stronger by ensuring subgrouping is still done correctly. Otherwise, you can get the same output (albeit with a different error) with the following configuration:

```ts
options: [
  {
    ...options,
    groups: ['get-method', 'set-method', 'unknown']
  },
],
```

This test is stronger:

```ts
it('sorts getters and setters using subgroup-order fallback', async () => {
  await invalid({
    errors: [
      {
        messageId: 'unexpectedClassesOrder',
      },
      {
        messageId: 'unexpectedClassesOrder',
      },
    ],
    options: [
      {
        ...options,
        groups: [['get-method', 'set-method'], 'unknown'],
        fallbackSort: { type: 'subgroup-order' },
      },
    ],
    output: dedent`
      class Class {
        get a() {}
        set a(value) {}
        get b() {}
        set b(value) {}
      }
    `,
    code: dedent`
      class Class {
        set b(value) {}
        set a(value) {}
        get b() {}
        get a() {}
      }
    `,
  })
})
```

rules/sort-enums/build-comparator-by-options-computer.ts
```diff
@@ -48,10 +49,15 @@ export function buildComparatorByOptionsComputer(
 }
 
 let byNonNumericValueComparatorComputer: ComparatorByOptionsComputer<
-  Required<Options[number]>,
+  Required<Options[number]> & { subgroupOrder?: string[] | null },
```
hugop95: @Lievesley 

I've wrote a new comment regarding where to put `getSubgroupOrder`. This line shouldn't be updated in the end (same for other similar cases).

test/utils/sort-nodes-by-groups.subgroup-order.test.ts
```diff
@@ -0,0 +1,112 @@
+import { describe, expect, it, vi } from 'vitest'
```
hugop95: @Lievesley 

Those tests (and the ones added in `sortNodesByGroups`) shouldn't be needed as `sortNodesByGroups` doesn't need to know that this new sort type exists.

`sort-nodes-by-groups` should be unchanged.

utils/compare/build-subgroup-order-comparator.ts
```diff
@@ -0,0 +1,27 @@
+import type { Comparator } from './default-comparator-by-options-computer'
+import type { OrderOption } from '../../types/common-options'
+import type { SortingNode } from '../../types/sorting-node'
+
+export function buildSubgroupOrderComparator(
```
hugop95: @Lievesley 

The `getSubgroupOrder` function in `sort-nodes-by-groups` should be here instead, and `sortNodesByGroups` should be completely unchanged.

Here is the pseudo-code I had envisioned:

```ts
export function buildSubgroupOrderComparator({
  order,
  groups,
}: Pick<CommonGroupsOptions<unknown, unknown, string>, "groups">
  & Pick<CommonOptions, "order">) {
  return (a, b) => {
    let subgroupContainingA = computeSubgroupContainingNode(a, groups) // This function searches `groups` based on the group of the node entered
    let subgroupContainingB = computeSubgroupContainingNode(b, groups)

    if (subgroupContainingA !== subgroupContainingB) {
      return 0
    }

    let result = // Compare order in the subgroup
    return computeOrderedValue(result, order) // Take into account the entered `order` 
  }
}

function computeSubgroupContainingNode(node: SortingNode, groups): string[] | null {
   return groups.find(...)
}
```

Then in the comparator computers: 

```ts
case 'subgroup-order':
  return buildSubgroupOrderComparator(options)
```

Note that this pseudo-code is not optimized, we could certainly improve things by precomputing things in a map, but a first basic implementation focusing on readability first should be more than enough considering that we will only go there when fallback sorting (which is rare).

Lievesley: @hugop95 it seems we also need getSubgroupOrder in compute-overridden-options-by-group-index.ts
are you ok with that dependency?

hugop95: @Lievesley 

> it seems we also need getSubgroupOrder in compute-overridden-options-by-group-index.ts
are you ok with that dependency?

This would be a blocking change: `computeOverriddenOptionsByGroupIndex` should not be aware of anything related to subgrouping order.

The following `buildSugroupOrderComparator` implementation isolates the notion of subroup ordering here without leaking it elsewhere:

```ts
import type {
  CommonGroupsOptions,
  GroupsOptions,
} from '../../types/common-groups-options'
import type { Comparator } from './default-comparator-by-options-computer'
import type { CommonOptions } from '../../types/common-options'
import type { SortingNode } from '../../types/sorting-node'

import { isGroupWithOverridesOption } from '../is-group-with-overrides-option'
import { isNewlinesBetweenOption } from '../is-newlines-between-option'
import { UnreachableCaseError } from '../unreachable-case-error'
import { computeOrderedValue } from './compute-ordered-value'

export function buildSubgroupOrderComparator({
  groups,
  order
}: Pick<CommonOptions, "order">
  & {groups: GroupsOptions}
): Comparator<SortingNode> {
  return (a, b) => {
    let subgroupContainingA = computeSubgroupContainingNode(a, groups)
    let subgroupContainingB = computeSubgroupContainingNode(b, groups)

    if (
      !subgroupContainingA ||
      !subgroupContainingB ||
      subgroupContainingA !== subgroupContainingB
    ) {
      return 0
    }

    let indexOfAInSubgroup = subgroupContainingA.indexOf(a.group)
    let indexOfBInSubgroup = subgroupContainingB.indexOf(b.group)
    
    let result = indexOfAInSubgroup - indexOfBInSubgroup
    return computeOrderedValue(result, order)
  }
}

function computeSubgroupContainingNode(
  sortingNode: SortingNode,
  groups: CommonGroupsOptions<unknown, unknown, string>['groups'],
): string[] | null {
  for (let group of groups) {
    if (isNewlinesBetweenOption(group)) {
      continue
    }
    if (typeof group === 'string' || Array.isArray(group)) {
      if (doesStringSubgroupContainsNode(sortingNode, group)) {
        return group
      }
      continue;
    }
    if (isGroupWithOverridesOption(group)) {
      if (doesStringSubgroupContainsNode(sortingNode, group.group)) {
        return group.group
      }
      continue;
    }

    /* v8 ignore next -- @preserve Exhaustive guard for unsupported group option. */
    throw new UnreachableCaseError(group)
  }

  return null
}

function doesStringSubgroupContainsNode(sortingNode: SortingNode, subgroup: string[] | string): subgroup is string[] {
  if (typeof subgroup === 'string') {
    return false
  }
  return subgroup.includes(sortingNode.group);
}
```

Then you can update `defaultComparatorByOptionsComputer` to:

```ts
type Options = Pick<
  CommonOptions<TypeOption>,
  'specialCharacters' | 'ignoreCase' | 'alphabet' | 'locales' | 'order' | 'type'
> &
  Pick<CommonOptions, 'fallbackSort'> & {
  groups?: GroupsOptions
}

export let defaultComparatorByOptionsComputer: ComparatorByOptionsComputer<
  Options,
  SortingNode
> = options => {
  switch (options.type) {
    case 'subgroup-order':
      return buildSubgroupOrderComparator({
        ...options,
        groups: options.groups ?? [], // `sort-switch-case` currently doesn't support `groups`
      })
[...]
```

---

Like this, you only end up updating comparators, and not the rest of the framework.

hugop95: @Lievesley N.B: 

We could unit test `buildSubgroupOrderComparator` similarly to how `buildLineLengthComparator` is.

Lievesley: @hugop95 please check the latest, should be roughly what your looking for.
If so I can continue fixing tests

types/common-options.ts
```diff
@@ -92,6 +92,12 @@ export type CommonOptions<
  *   // Shorter lines appear before longer lines
  */
 export type TypeOption =
+  /**
+   * Sorts by the subgroup order defined in `groups` arrays. Useful as a
```
hugop95: @Lievesley [nit] Can we use `Sort` like in line 108 ?

utils/compare/build-subgroup-order-comparator.ts
hugop95: @Lievesley 

Thank you! It looks good 👍

Regarding tests, you can remove tests added in `sort-nodes-by-groups`: they test a specific implementation of `defaultComparatorByOptionsComputer`, and not really the logic behind the `sort-nodes-by-group` file.

Fully unit-testing `buildSubgroupOrderComparator` alongside specification tests in `sort-classes` (that you have already added) should be enough to have good coverage while not adding noisy tests.

If needed for remaining uncovered cases (`buildComparatorByOptionsComputer` for `sort-enums`, `sort-modules`, etc), it's ok to `// v8 ignore` them for now as it's not meant to really be useful for those rules for now, and there are no specification tests on that for them right now.

Lievesley: Should be done now buildSubgroupOrderComparator has 100% coverage and added v8 ignore in various places

types/common-options.ts
Lievesley: Done

test/utils/sort-nodes-by-groups.subgroup-order.test.ts
Lievesley: Reverted

rules/sort-enums/build-comparator-by-options-computer.ts
Lievesley: Reverted

rules/sort-imports/comparator-by-options-computer.ts
```diff
@@ -18,6 +18,7 @@ export let comparatorByOptionsComputer: ComparatorByOptionsComputer<
   switch (options.type) {
     case 'type-import-first':
       return (a, b) => compareTypeImportFirst(a, b, options)
+    case 'subgroup-order':
```
hugop95, line 22: @Lievesley 

You're probably missing a `/* v8 ignore next 2 -- @preserve Untested for now as not a relevant sort for this rule. */` here for 100% coverage.

rules/sort-modules/build-comparator-by-options-computer.ts
```diff
@@ -24,6 +24,7 @@ export function buildComparatorByOptionsComputer({
 > {
   return options => {
     switch (options.type) {
+      case 'subgroup-order':
```
hugop95, line 28: @Lievesley Same thing here regarding the missing `/* v8 ignore next -- @preserve Untested for now as not a relevant sort for this rule. */`

utils/compare/build-subgroup-order-comparator.ts
```diff
@@ -0,0 +1,76 @@
...
+    if (isNewlinesBetweenOption(group)) {
+      continue
+    }
+    if (typeof group === 'string' || Array.isArray(group)) {
+      if (doesStringSubgroupContainsNode(sortingNode, group)) {
+        return group
+      }
+      continue
+    }
+    if (isGroupWithOverridesOption(group)) {
```
hugop95, line 55: @Lievesley 

You're missing coverage on the `else` case here, you need to add `/* v8 ignore else -- @preserve Exhaustive guard for unsupported group option. */` before, similarly to

https://github.com/azat-io/eslint-plugin-perfectionist/blob/de28a71501081f044ece00064dc45e9a3a137848/utils/compute-group-name.ts#L23-L28

utils/compare/default-comparator-by-options-computer.ts
```diff
@@ -18,13 +20,20 @@ type Options = Pick<
...
 
 export let defaultComparatorByOptionsComputer: ComparatorByOptionsComputer<
   Options,
   SortingNode
 > = options => {
   switch (options.type) {
+    case 'subgroup-order':
+      return buildSubgroupOrderComparator({
+        ...options,
+        groups: options.groups ?? [],
```
hugop95: @Lievesley 

We're missing coverage here because `sort-switch-case` has no specification test that allows going through the `[]` branch.

For now, you can simply write

```ts
    case 'subgroup-order':
      /* v8 ignore start -- @preserve sort-switch-case specific */
      if (!options.groups) {
        return unsortedComparator
      }
      /* v8 ignore stop -- @preserve sort-switch-case specific */
      return buildSubgroupOrderComparator({
        ...options,
        groups: options.groups,
      })
```

test/utils/sort-nodes-by-groups.subgroup-order.test.ts
hugop95: @Lievesley 

Those tests (`sort-nodes-by-groups.subroup-order.test.ts`) are still here 😁

test/rules/sort-classes.test.ts
Lievesley: Done

docs/content/rules/sort-imports.mdx
Lievesley: Done

rules/sort-imports/comparator-by-options-computer.ts
Lievesley: Done

utils/compare/build-subgroup-order-comparator.ts
Lievesley: Done

utils/compare/default-comparator-by-options-computer.ts
Lievesley: Done

rules/sort-modules/build-comparator-by-options-computer.ts
Lievesley: Done

test/utils/sort-nodes-by-groups.subgroup-order.test.ts
```diff
@@ -0,0 +1,121 @@
...
```
hugop95: @Lievesley 

Could you remove these noisy tests please ?

Lievesley: Deleted

Lievesley:
@hugop95 seems its complete. Thanks for your help!

Lievesley:
@hugop95 are you able to approve the workflow so this PR can be merged?

hugop95:
@Lievesley 

Please wait for @azat-io's review, they are the one who can approve the workflow.

pnpm-workspace.yaml
```diff
@@ -4,3 +4,7 @@ configDependencies:
 onlyBuiltDependenciesFile: node_modules/.pnpm-config/@pnpm/plugin-trusted-deps/allow.json
 
 sideEffectsCache: false
+
+packages:
```
azat-io: These changes are unnecessary; this project is not a monorepo.

azat-io (changes requested):
Can you ran `pnpm test:format --write`?

pnpm-workspace.yaml
Lievesley: pnpm install - fails without it

pnpm requires a packages list if pnpm-workspace.yaml exists. Even for a single-package repo, packages: - . is the way to declare the root as the workspace package; without it, pnpm install fails.

If you want to remove it, the alternative is to remove pnpm-workspace.yaml entirely and move those settings elsewhere—but that changes tooling behavior.

Lievesley: @azat-io thanks for the review! and let me know if you agree.
I'm interested how the team installs the packages. Maybe it wasn't necessary with an older pnpm version?

azat-io: Thanks! I tested on pnpm 10.27.0: `pnpm install` works without `packages`. Our `pnpm-workspace.yaml` isn’t empty (has `configDependencies` / `onlyBuiltDependenciesFile` / `sideEffectsCache`), so https://github.com/pnpm/pnpm/issues/7307 (empty file in pnpm 8.10.4) doesn’t apply.

Can you share the exact error, pnpm version, and where you run the command? It sounds like an older pnpm (corepack) or a different/empty workspace file is being picked up. If it reproduces on latest pnpm, I’m happy to revisit packages: `['.']`, but note it changes workspace semantics.

Lievesley: You're right. Seems that pnpm 9.5 was used in my eslint-plugin-perfectionist workspace. I updated it to 10.27.0 and pnpm install works without the change so I reverted it.

Error was:

eslint-plugin-perfectionist % pnpm install
 ERROR  packages field missing or empty

azat-io:
Thanks!

Issue #376, opened 2024-11-19
Bug(sort-modules): Specific `export decorated` classes not handled correctly 


hugop95:
### Describe the bug

ESLint's parser doesn't seem to detect the decorator ranges correctly with the following syntax:

### Code example

```ts
class B {}

@A
export class A {}
```

This will result in

```ts
export class A {}

@A
class B {}
```

Removing `export` will make it work correctly.

Writing 

```ts
class B {}

export @A class A {}
```

will also make it work, but this is not an usual syntax.

### ESLint version

4.0.0

### ESLint Plugin Perfectionist version

8.57.0

### Validations

- [X] Read the [docs](https://perfectionist.dev).
- [X] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

azat-io:
Released in [v4.0.1](https://github.com/azat-io/eslint-plugin-perfectionist/releases/tag/v4.0.1).

PR #675, opened 2026-01-05
feat(sort-modules): allow exported then decorated classes to be sorted

hugop95:
- Related comment: https://github.com/azat-io/eslint-plugin-perfectionist/issues/434#issuecomment-3710506816

### Description

Since https://github.com/azat-io/eslint-plugin-perfectionist/issues/376, any exported + decorated classes are ignored in `sort-modules`. The root cause is an `estree` [incompatibility](https://github.com/estree/estree/issues/315), making decorators stay in place.

While the syntax is unusual, exported **then** decorated classes can actually be handled correctly.

```ts
export @A class A {}  // ✅ Decorator will follow the node correctly
```

This PR adds support for this case, meaning that users will always have a fallback way to sort these nodes if truly needed, rather than have to `// eslint-disable` (cf https://github.com/azat-io/eslint-plugin-perfectionist/issues/434#issuecomment-3710506816).

Documentation is updated to inform of this limitation:

image

- [x] New Feature

### DIFF
--- a/docs/content/rules/sort-modules.mdx
+++ b/docs/content/rules/sort-modules.mdx
@@ -171,7 +171,9 @@ This rule sorts the following module members:
 The following elements are not sorted by this rule:
 - `imports` (see the `sort-imports` rule).
 - `'from' exports` (see the `sort-exports` rule).
-- any other `expression`, in order to ensure compilation and runtime behavior.
+- Decorated **then** exported classes due to a known ESLint [limitation](https://github.com/estree/estree/issues/315).
+Place the `export`/`export default` keyword **before** the decorators in order to sort those classes.
+- Any other `expression`, in order to ensure compilation and runtime behavior.
 
 ## Options
 


azat-io:
Should I wait for the PR, or can I release the package now?

hugop95:
@azat-io 

Sorry, I didn't mean to rebase, I should have merged dev instead.

> Should I wait for the PR, or can I release the package now?

Moving things to resolve https://github.com/azat-io/eslint-plugin-perfectionist/pull/675#pullrequestreview-3631295351 should be quick.

azat-io:
Thank you! I'll wait!

hugop95:
@azat-io 

> Let's also check the following cases:

I see those are the unhandled cases added in https://github.com/azat-io/eslint-plugin-perfectionist/pull/648.

Unfortunately, it's not easy to fix them: the current [dependency parsing algorithm](https://github.com/azat-io/eslint-plugin-perfectionist/blob/845fa8011b6b87ed23f5a427488a315f71754e23/rules/sort-modules/compute-dependencies.ts) checks various fields in each node.

Removing a condition such as `if ('members' in nodeToCheck) { traverseNode(nodeToCheck.members) }` means that we now won't detect some other (valid) dependencies.

The fix would be to "remember" as we parse the AST tree if we are in specific invalid nodes or not. It's possible but not simple (it would at least need another PR).

Additionally, I'm experimenting with a new [dependency detection algorithm](https://github.com/azat-io/eslint-plugin-perfectionist/pull/674) that uses ESLint's built-in **scope manager**, the same used for the [`no-use-before-define`](https://eslint.org/docs/latest/rules/no-use-before-define) rule.

The good news is that this new algorithm resolves all the unhandled cases automatically because it can not generate false positives. The drawback is that it can miss some dependencies still (but this can be patched).

azat-io:
I think we can accept the current PR and make the fixes later?

hugop95:
@azat-io 

> I think we can accept the current PR and make the fixes later?

Yes: those unhandled cases only happen for `type: 'usage'` and were not introduced in this PR 👍

azat-io (approved):
Thanks!

PR #678, opened 2026-01-06
chore: set the project's `fallbackSort` to `{ type: 'alphabetical', order: 'asc' }`

hugop95:
### Description

The project uses `type: 'line-length'` with the default configuration, meaning that `fallbackSort` is `unsorted`. 

I sometimes accidently end up permutating two imports or properties which pollutes my commits without being corrected.

As such, I suggest setting `fallbackSort` to `{ type: 'alphabetical', order: 'asc'` } in order for everything to be consistent.

- [x] Other

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## v5.3.1
+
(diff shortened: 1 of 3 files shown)


hugop95:
@azat-io 

OK 👍 
Let's keep it as it is.

PR #680, opened 2026-01-08
feat(sort-objects): allow `sortBy: 'name' | 'value'`

### LINKED ISSUE #679: Feature: Sort object values alphabetically

sort-objects

### Describe the problem

Hi there! I love this plugin and I have found it very useful. One thing that I recently wanted to do, but it doesn't seem like it's currently possible, is alphabetically sort an object by its *values*, not *keys*.

### Code example

For example, if I have a sitemap object where the paths (values) are more important than the keys:

```ts
const Sitemap = {
  Health: '/health',
  Home: '/',
  Projects: '/projects',
  Login: '/login',
  Project: '/project/:id',
} as const;
```

It would be nice to sort the _paths_ alphabetically:

```ts
const Sitemap = {
  Home: '/',
  Health: '/health',
  Login: '/login',
  Project: '/project/:id',
  Projects: '/projects',
} as const;
```

The closest option I see for sorting by value is `line-length`, but that's not quite what I want:

```js
      'error',
      {
        type: 'line-length',
        useConfigurationIf: { declarationMatchesPattern: 'Sitemap' },
      },
      { type: 'unsorted' },
    ],
```

### Additional comments

_No response_

### Validations

hugop95:
- Resolves #679

### Description

This PR adds a `sortBy: 'name' | 'value'` option to `sort-objects`, similarly to `sort-object-types`/`sort-interfaces`.

It natively supports:
- `fallbackSort.sortBy`.
- `groups` overriding
  - `groups: [{group: 'property', sortBy: 'value' }]`
  - `groups: [{group: 'property', fallbackSort: { sortBy: 'value' }}]`
- `customGroups` (same as above, supports `sortBy` and `fallbackSort.sortBy`). 

- [x] New Feature

### DIFF
--- a/test/rules/sort-objects.test.ts
+++ b/test/rules/sort-objects.test.ts
@@ -4146,7 +4146,7 @@ describe('sort-objects', () => {
         })
       })
 
-      it('puts functions first', async () => {
+      it('treats functions as empty in non-destructured objects', async () => {
         await invalid({
           errors: [
             {
@@ -4183,6 +4183,33 @@ describe('sort-objects', () => {
           ],
           output: dedent`
             let obj = {
+              b: function b() {},
+              a: 'a',
+            }
+          `,
+          code: dedent`
+            let obj = {
+              a: 'a',
+              b: function b() {},
+            }
+          `,
+          options: [
+            {
+              ...options,
+              sortBy: 'value',
+            },
+          ],
+        })
+
+        await invalid({
+          errors: [
+            {
+              messageId: 'unexpectedObjectsOrder',
+              data: { right: 'b', left: 'a' },
+            },
+          ],
+          output: dedent`
+            let obj = {
               b: () => {},
               a: 'a',
             }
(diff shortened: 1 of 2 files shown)


test/rules/sort-objects.test.ts
```diff
@@ -4084,6 +4084,358 @@ describe('sort-objects', () => {
...
+            }
+          `,
+          code: dedent`
+            let obj = {
+              a: 'b',
+              a: 'a',
+            }
+          `,
+        })
+      })
```
azat-io, line 4501: Shall we add a test with destructuring?

```ts
it('treats function defaults as empty in destructured objects', async () => {
  await invalid({
    output: dedent`
      let Func = ({
        a = function foo() {},
        b = 'b',
      }) => {
        // ...
      }
    `,
    code: dedent`
      let Func = ({
        b = 'b',
        a = function foo() {},
      }) => {
        // ...
      }
    `,
    errors: [
      {
        messageId: 'unexpectedObjectsOrder',
      },
    ],
    options: [
      {
        ...options,
        sortBy: 'value',
      },
    ],
  })
})
```

hugop95: @azat-io 👍

[`54b3727` (#680)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/680/commits/54b37277ab4373bca253c0dd648747f103220de4)

PR #686, opened 2026-01-15
docs: fix default group names in sort-imports

dbaty:
### Description

The [sort-imports section of the documentation](https://perfectionist.dev/rules/sort-imports#groups) refers to  `*-type` groups, which have been removed from 5.0.0 (as per f0894880df92bf5). They are now available under `type-*`. This PR fixes the documentation.

I also added `type-import` which was missing.

- [x] Documentation update


azat-io (approved):
Thanks!

PR #692, opened 2026-02-05
feat: import partition sorting and type separation

cmath10:
### Description

Adds optional import partition sorting and type separation behavior:
- `sort-imports`: type-first partition sorting.
- `sort-imports`: specifier-aware partition splitting to avoid interleaved order.
- `separate-type-imports`: split mixed type/value imports.
- `separate-type-partitions`: enforce a blank line between type and value blocks.

Motivation: provide a more neat separation and sorting of imports when desired.
All behavior is optional, so existing setups should be unaffected unless new
options are explicitly enabled in the ESLint config.

Alongside the new tests (keeping 100% coverage as required by the vitest config),
we also tried these options on our own codebase with 1200+ modules and 160k+
lines of code.

We have not updated docs/ yet; we wanted to get a review first, and if anything
needs adjustments we will apply them before doing the documentation work to
avoid doing it twice.

This is not intended as a final version. Happy to add/change/remove any details you request.

### Additional context

Please focus on rule behavior and option naming; I can adapt the config or output
format if you want adjustments.

Usage examples:

```ts
// sort-imports: type-first partition sorting
{
  "rules": {
      "partitionByNewLine": true,
      "partitionSorting": "type-first",
      "newlinesBetween": "ignore",
      "newlinesInside": "ignore"
    }]
  }
}

// before
import type { User } from '@/types/user'

import { createClient } from '@/lib/api'

import type { BillingPlan } from '@/billing/types'

// after
import type { User } from '@/types/user'

import type { BillingPlan } from '@/billing/types'

import { createClient } from '@/lib/api'
```

```ts
// sort-imports: specifier-aware partition splitting
{
  "rules": {
      "sortBy": "specifier",
      "partitionImportsSplitOnSort": true,
      "groups": ["unknown"]
    }]
  }
}

// before
import { getUser, useCurrentUser } from '@/features/users'
import { auditEvent } from '@/services/audit'

// after
import { getUser } from '@/features/users'
import { auditEvent } from '@/services/audit'
import { useCurrentUser } from '@/features/users'
```

```ts
// separate-type-imports: split mixed type/value specifiers
{
  "rules": {
    "perfectionist/separate-type-imports": "error"
  }
}

// before
import { type UserDTO, fetchUser } from '@/api/users'

// after
import type { UserDTO } from '@/api/users'

import { fetchUser } from '@/api/users'
```

```ts
// separate-type-partitions: enforce blank line between type/value blocks
{
  "rules": {
    "perfectionist/separate-type-partitions": "error"
  }
}

// before
import type { Order } from '@/types/orders'
import { calculateTotal } from '@/lib/orders'

// after
import type { Order } from '@/types/orders'

import { calculateTotal } from '@/lib/orders'
```

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

- [x] Check that there isn't already a PR that solves the problem the same way
      to avoid creating a duplicate.
- [x] Provide a description in this PR that addresses **what** the PR is
      solving, or reference the issue that it solves (e.g. `fixes #123`).
- [x] Ideally, include relevant tests that fail without this PR but pass with
      it.
- [x] Read
      [contribution guidelines](https://github.com/azat-io/eslint-plugin-perfectionist/blob/main/contributing.md).

### DIFF
--- a/rules/separate-type-imports.ts
+++ b/rules/separate-type-imports.ts
@@ -115,8 +115,7 @@ export default createEslintRule<Options, MessageId>({
       useImportType: 'Use "import type" for type-only named imports.',
     },
     docs: {
-      description:
-        'Require separating inline type imports from value imports.',
+      description: 'Require separating inline type imports from value imports.',
       recommended: true,
     },
     type: 'suggestion',
@@ -188,8 +187,7 @@ function buildSeparatedImportsFix({
     namespaceText:
       /* v8 ignore next -- @preserve Namespace specifiers cannot coexist with named specifiers in valid syntax. */
       namespaceSpecifier ? sourceCode.getText(namespaceSpecifier) : null,
-    defaultText:
-      defaultSpecifier ? sourceCode.getText(defaultSpecifier) : null,
+    defaultText: defaultSpecifier ? sourceCode.getText(defaultSpecifier) : null,
     namedText: valueNamedText,
     importKind: 'value',
     attributesText,
@@ -240,9 +238,7 @@ function buildTypeOnlyImport({
 
   let segments = buildNamedSpecifierSegments(namedSpecifiers, sourceCode)
 
-  let typeSegments = segments.map(segment =>
-    stripTypeKeyword(segment.text),
-  )
+  let typeSegments = segments.map(segment => stripTypeKeyword(segment.text))
   let typeNamedText = buildNamedListText(typeSegments)
 
   let sourceText = sourceCode.getText(node.source)
@@ -261,7 +257,7 @@ function getPartitionStart(
   node: TSESTree.ImportDeclaration,
   sourceCode: TSESLintSourceCode,
 ): TSESTree.ImportDeclaration | null {
-  let {parent} = node
+  let { parent } = node
   /* v8 ignore next 2 -- @preserve ImportDeclaration nodes belong to Program. */
   if (parent.type !== AST_NODE_TYPES.Program) {
     return null
@@ -362,12 +358,9 @@ function getClosingBraceToken(
   sourceCode: TSESLintSourceCode,
   specifier: TSESTree.ImportSpecifier,
 ): TSESTree.Token {
-  let token = sourceCode.getTokenAfter(
-    specifier,
-    {
-      includeComments: false,
-    },
-  )!
+  let token = sourceCode.getTokenAfter(specifier, {
+    includeComments: false,
+  })!
   if (token.value === ',') {
     return sourceCode.getTokenAfter(specifier, {
       filter: nextToken => nextToken.value === '}',
@@ -380,10 +373,7 @@ function getClosingBraceToken(
 function buildNamedListText(segments: string[]): string {
   let combined = segments.join('').trim()
   combined = combined.replace(/,\s*$/u, '')
-  if (
-    /\r?\n/u.test(combined) &&
-    !/\/\*|\/\//u.test(combined)
-  ) {
+  if (/\r?\n/u.test(combined) && !/\/\*|\/\//u.test(combined)) {
     combined = combined
       .replaceAll(/\s*\n\s*/gu, ' ')
       .replaceAll(/\s+/gu, ' ')
--- a/rules/sort-imports.ts
+++ b/rules/sort-imports.ts
@@ -3,7 +3,12 @@ import type { TSESTree } from '@typescript-eslint/types'
 
 import { AST_NODE_TYPES } from '@typescript-eslint/utils'
 
-import type { SortImportsSortingNode, Modifier, Selector, Options, } from './sort-imports/types'
+import type {
+  SortImportsSortingNode,
+  Modifier,
+  Selector,
+  Options,
+} from './sort-imports/types'
 import type { CustomOrderFixesParameters } from '../utils/make-fixes'
 
 import {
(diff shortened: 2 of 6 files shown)


hugop95 (changes requested):
@cmath10 

Thank you for your proposals. I haven't done an extensive review yet, and I'm going to focus on the `sort-imports` changes, as I don't think the two new rules (`separate-type-imports` and `separate-type-partitions`) fit `perfectionist`:
- They don't use the same naming as any other rule (all start with `sort-X`).
- They don't use any of the plugin's utility functions, so they are independent and could also be put in a separate package.

Regarding `partitionImportsSplitOnSort`, this doesn't fit the plugin functionally speaking. We don't regroup imports into a single one either and leave that to other rules.

For `partitionSorting`: I feel like your proposal is trying to mix two very independent features together: being able to sort partitioned elements together (and by principle, partitioned elements are independent).

Regarding `partitionSortingStable`, I didn't understand what this did from my overview of the changes.

---

I think that the approach toward fixing your pain point doesn't fit the current state of the plugin: today, sorting is put behind a framework that is common to all rules. To put it simply:
- The user partitions things as they want.
- Within a given partition, the user groups things as they want, and orders those groups (`groups` option).
- Within a group, the user sorts things as they want (`type` option, etc).
- Additionally, the user can choose to add newlines between each element of a group, or between groups (`newlinesBetween` option, etc).

These options work together nicely, and when they don't, it's very explicit: `partitionByNewline` is not compatible with `newlinesBetween` for example.

This framework works for all rules (except `sort-switch-case` which actually sorts multiple things at once), and in the current state, I can't see how we could make other rules benefit from your proposals.

I don't see how your new options fit with the other options, and there seems to be a lot of new implicit incompatibilities created:
- How do they work with `groups`, `customGroups`, `newlinesBetween`, etc?
- Is dependency sorting mandatory for runtime well handled with these options?

In my opinion, we should instead:
- See what is your pain point with the current rule.
  - It's not explicit in your PR description, and there is no issue associated. Those are important factors regarding maintainability: your PR adds a huge tech debt to a single rule, and I suspect the feature/maintenance ROI to be extremely low.
- Once the pain point has been identified, we see if we can find a pattern than can be reproduced with other rules. In particular:
  - Is this related specifically to `sort-imports`?
  - Is this something related to the notion of **partitioning**, **sorting**, **grouping**, etc?
    - In that case, we need to investigate how to gracefully solve this pain point and make the new option(s)/concept fit with the existing options/concepts.

cmath10:
@hugop95 

Thank you for the feedback. The main pain points are:

* **Dependency analysis and cognitive load.**
* **Merge conflicts** — our experience shows that **70-80% of merge conflicts** occur specifically within import declarations.

At my company, we have a massive codebase (the mentioned above 1200 modules is a single workspace inside, not the whole code we have), and it is quite normal for us to have import blocks exceeding 40 lines. We needed a consistent approach to reduce friction while managing a high pace of changes, so we adopted these rules:

* **Focus on *what*** is being imported (responsibility) rather than *where* it comes from.
* **The Hierarchy:** `Types` (always first) → `Namespaces` → `Classes` → `Components` → `Functions` → `Constants` → `Enums`.
* **The Requirement:** Every specifier must be easily discoverable via alphabetical sorting, and types must be strictly isolated to prevent them from being mixed into value-logic during refactoring.

I understand the limitations of ESLint analysis and **do not intend** to implement a deep static analyzer or complex heuristics. However, we need a "bare minimum" that includes:

* Separating types and placing type imports above all others.
* Sorting specifiers inside a partition alphabetically and forcing their position, even if splitting an import is needed to accomplish the goal.

I tried different configurations and rules and found it impossible to configure the sorting as needed. I added the `separate-X` rules because they were part of the initial problem I was solving; I was unaware of the philosophy you described at the time and thought they might be useful. If your plugin is strictly about sorting, I will follow your suggestion and create a separate plugin for that.

**About partitionSortingStable:**

Controls whether the partition order within the same category remains stable when `partitionSorting` is set to `"type-first"`.

**Behavior when `false`:**
1) Partitions are grouped by kind: type-only partitions first, then value partitions.
2) Within the same kind, partitions are sorted by the name of the first import in the partition (`left.nodes[0].name`), using the same comparison rules as regular import sorting (`order`, `ignoreCase`, `locales`, `specialCharacters`).
3) If the first-import names are equal, the original order is preserved via a fallback to the partition start position.

**Examples:**

```ts
// options: partitionByNewLine: true, partitionSorting: 'type-first',
// partitionSortingStable: false

// before
import type { Zeta } from 'zeta'

import type { Alpha } from 'alpha'

import { Omega } from 'omega'

// after
import type { Alpha } from 'alpha'

import type { Zeta } from 'zeta'

import { Omega } from 'omega'
```

If the PR does not fit the philosophy of your project completely, I can continue addressing these pain points in my fork. In that case, feel free to close the PR. The bare minimum for my company's projects has already been accomplished with the code I introduced here.

cmath10:
I’ve decided to move forward with my fork. Thanks for the review and for clarifying the project’s philosophy.

Final thoughts: after a quick look through the documentation, I think it would be helpful to include that philosophy description in CONTRIBUTING.md.

Thanks for your time, and have a great day!

hugop95:
@cmath10 

Thank you for the explanations. Your use case is an interesting one, and while I definitely don't have the full context, I am wondering if we wouldn't have been able to fix this pain by:
- Having `separate-type-imports` as a custom ESLint rule on your side (or in another package).
- Having a custom ESLint rule as well to split imports that have X specifiers into X imports.
- Using `groups` to force type-imports to be at above value imports.
- Within each group, using `sortBy: 'specifier'` to sort by specifier. 

I understand that you use `partitionByNewline: true`, and it looks like this is where limitations come-in, as I understand you would like to sort things from different partitions together (maybe I misunderstood that part).

> Final thoughts: after a quick look through the documentation, I think it would be helpful to include that philosophy description in CONTRIBUTING.md.

Thank you for the feedback, I'll see how we can improve this part 👍

Have a good day!

Issue #662, opened 2025-12-31
Bug: `sort-classes` crashes on class bodies containing nodes without `key` property (e.g., Ember's GlimmerTemplate)


lougreenwood:
### Describe the bug

`perfectionist/sort-classes` crashes with `TypeError: Cannot read properties of undefined (reading 'type')` when linting files parsed by `ember-eslint-parser`. This parser produces `GlimmerTemplate` AST nodes inside class bodies for Ember's `<template>` tag syntax, and these nodes don't have a `key` property.

### Error

```
TypeError: Cannot read properties of undefined (reading 'type')
Occurred while linting /path/to/component.gts:277
Rule: "perfectionist/sort-classes"
    at sort-classes.js:242:33
```

### Root Cause

The `sort-classes` rule assumes all class body members (except `StaticBlock` and `TSIndexSignature`) have a `key` property:

```typescript
if (member.type === 'StaticBlock') {
  name = 'static'
} else if (member.type === 'TSIndexSignature') {
  name = sourceCode.text.slice(...)
} else if (member.key.type === 'Identifier') {  // crashes here
  ;({ name } = member.key)
} else {
  name = sourceCode.getText(member.key)
}
```

According to the [ESTree ES2022 spec](https://github.com/estree/estree/blob/master/es2022.md#classbody), `ClassBody.body` can contain:
- `MethodDefinition` (has `key`)
- `PropertyDefinition` (has `key`)  
- `StaticBlock` (**does not have `key`**)

The rule correctly handles `StaticBlock`, but third-party parsers like `ember-eslint-parser` can introduce additional node types (e.g., `GlimmerTemplate`) that also lack a `key` property. This causes the rule to crash.

### Context: Ember's `<template>` syntax

Ember.js uses [First-Class Component Templates](https://rfcs.emberjs.com/id/0779-first-class-component-templates/) with a `<template>` tag that can appear inside class bodies:

```typescript
// component.gts
export default class MyComponent extends Component {
  someProperty = 'value';
  
  <template>
    <div>{{this.someProperty}}</div>
  </template>
}
```

When parsed by `ember-eslint-parser`, the `<template>` tag becomes a `GlimmerTemplate` node in the class body, which doesn't conform to standard ESTree class member types.

### Related Issues

[...]

PR #694, opened 2026-02-07
fix(sort-classes): avoid crashes on unknown class elements

azat-io:
### Description

Fixes `sort-classes` to safely handle unknown class elements (treating them as partition boundaries) and adds tests reproducing the crash with a custom parser.

### Additional context

#662

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/rules/sort-classes.ts
+++ b/rules/sort-classes.ts
@@ -165,9 +165,6 @@ export default createEslintRule<Options, MessageId>({
           member,
         ) => {
           if (!isKnownClassElement(member)) {
-            if (accumulator.at(-1)?.length) {
-              accumulator.push([])
-            }
             return accumulator
           }
 
--- a/rules/sort-classes/is-known-class-element.ts
+++ b/rules/sort-classes/is-known-class-element.ts
@@ -5,16 +5,13 @@ import { AST_NODE_TYPES } from '@typescript-eslint/utils'
 /**
  * Checks whether a class element is supported by the sort-classes rule.
  *
- * Unknown elements should be treated as partition boundaries to avoid
- * reordering across them and to prevent rule crashes with non-standard
- * parsers.
+ * Unknown elements should be ignored to avoid crashes with non-standard parsers
+ * while letting known elements keep their ordering behavior.
  *
  * @param member - The class element to check.
  * @returns True when the element is a known, supported class member.
  */
-export function isKnownClassElement(
-  member: TSESTree.ClassElement | { type: string },
-): boolean {
+export function isKnownClassElement(member: TSESTree.ClassElement): boolean {
   switch (member.type) {
     case AST_NODE_TYPES.TSAbstractPropertyDefinition:
     case AST_NODE_TYPES.TSAbstractMethodDefinition:
@@ -26,6 +23,11 @@ export function isKnownClassElement(
     case AST_NODE_TYPES.StaticBlock:
       return true
     default:
+      assertIsNotKnownClassElement(member)
       return false
   }
+}
+
+function assertIsNotKnownClassElement(_member: never): void {
+  // Compilation check only.
 }
--- a/test/rules/sort-classes.test.ts
+++ b/test/rules/sort-classes.test.ts
@@ -7,46 +7,54 @@ import { validateRuleJsonSchema } from '../utils/validate-rule-json-schema'
 import { Alphabet } from '../../utils/alphabet'
 import rule from '../../rules/sort-classes'
 
+type ParsedProgram = ReturnType<typeof typescriptParser.parseForESLint>['ast']
+
 function injectUnknownClassElement({
-  insertIndex,
+  elementToMarkAsUnknown,
+  code,
   ast,
 }: {
-  ast: { body: unknown[] }
-  insertIndex: number
+  elementToMarkAsUnknown: string
+  ast: ParsedProgram
+  code: string
 }): void {
-  let classDeclaration = ast.body.find(
-    node =>
-      !!node &&
-      typeof node === 'object' &&
-      (node as { type?: string }).type === 'ClassDeclaration',
-  ) as { body: { body: unknown[] } }
-  let classBody = classDeclaration.body
-
-  let baseNode =
-    classBody.body.find(
-      node =>
-        !!node &&
-        typeof node === 'object' &&
-        (node as { type?: string }).type === 'StaticBlock',
-    ) ?? classBody.body[0]
-  let unknownNode = {
-    ...(baseNode as Record<string, unknown>),
-    type: 'GlimmerTemplate',
+  let classDeclaration = ast.body.find(node => node.type === 'ClassDeclaration')
+  if (!classDeclaration) {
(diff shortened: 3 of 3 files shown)


rules/sort-classes/is-known-class-element.ts
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Unknown elements should be treated as partition boundaries to avoid
+ * reordering across them and to prevent rule crashes with non-standard
+ * parsers.
+ *
+ * @param member - The class element to check.
+ * @returns True when the element is a known, supported class member.
+ */
+export function isKnownClassElement(
+  member: TSESTree.ClassElement | { type: string },
```
hugop95: @azat-io 

For maximal compilation safety (we want to make sure that all official class elements are supported), I suggest the following:
- Type `member` as `TSESTree.ClassElement`.
- Update the `default` clause like the following:

```ts
export function isKnownClassElement(member: TSESTree.ClassElement): boolean {
  switch (member.type) {
    case AST_NODE_TYPES.TSAbstractPropertyDefinition:
    case AST_NODE_TYPES.TSAbstractMethodDefinition:
    case AST_NODE_TYPES.TSAbstractAccessorProperty:
    case AST_NODE_TYPES.PropertyDefinition:
    case AST_NODE_TYPES.MethodDefinition:
    case AST_NODE_TYPES.AccessorProperty:
    case AST_NODE_TYPES.TSIndexSignature:
    case AST_NODE_TYPES.StaticBlock:
      return true
    default:
      assertIsNotKnownClassElement(member)
      return false
  }
}

function assertIsNotKnownClassElement(_member: never): void {
  // Do nothing, compilation check
}
```

rules/sort-classes.ts
```diff
@@ -163,6 +164,13 @@ export default createEslintRule<Options, MessageId>({
           >[][],
           member,
         ) => {
+          if (!isKnownClassElement(member)) {
```
hugop95, line 167: @azat-io 

IMO this is too opinionated: I expect those unknown elements to simply be ignored rather than act as partition delimiters.

Ideally, we should probably use the `isNodeIgnored` attribute, as it ensures that unknown nodes stay at the same place after auto-linting.

In practice, simply doing 

```ts
if (!isKnownClassElement(member)) {
  return accumulator
}
```

is enough and still allows users to put back the unknown nodes where they want after auto-linting.

test/rules/sort-classes.test.ts
```diff
@@ -7,6 +7,50 @@ import { validateRuleJsonSchema } from '../utils/validate-rule-json-schema'
 import { Alphabet } from '../../utils/alphabet'
 import rule from '../../rules/sort-classes'
 
+function injectUnknownClassElement({
```
hugop95, line 12: @azat-io 

I find this a bit complex, as it implicitly asks tests to have a `static` element among things, or reads code comments.

I find the following implementation more explicit using `parserOptions.elementToMarkAsUnknown`: a string that matches the element to consider as unknown.

Additionally, adding `ast: ReturnType<typeof parse>` allows us to automatically benefit from AST typings.

```ts
let unknownClassElementParser = {
  ...typescriptParser,
  parseForESLint(
    code: string,
    parserOptions?: Parameters<typeof typescriptParser.parseForESLint>[1] & {
      elementToMarkAsUnknown: string
    },
  ) {
    if (!parserOptions?.elementToMarkAsUnknown) {
      throw new Error('parserOptions.elementToMarkAsUnknown option is required')
    }

    let result = typescriptParser.parseForESLint(code, parserOptions)
    markClassElementAsUnknown(result.ast, parserOptions.elementToMarkAsUnknown)
    return result

    function markClassElementAsUnknown(
      ast: ReturnType<typeof parse>,
      elementToMarkAsUnknown: string,
    ): void {
      let classDeclaration = ast.body.find(
        node => node.type === AST_NODE_TYPES.ClassDeclaration,
      )
      if (!classDeclaration) {
        throw new Error('No class declaration found')
      }

      let matchingNodeIndex = classDeclaration.body.body.findIndex(
        node =>
          code.slice(node.range[0], node.range[1]) === elementToMarkAsUnknown,
      )
      if (matchingNodeIndex < 0) {
        throw new Error(
          `No class element found with content "${elementToMarkAsUnknown}".`,
        )
      }

      classDeclaration.body.body[matchingNodeIndex] = {
        type: 'AnUnknownType',
      } as unknown as TSESTree.ClassElement
    }
  },
}
```

test/rules/sort-classes.test.ts
```diff
@@ -14314,6 +14363,32 @@ describe('sort-classes', () => {
       ).resolves.not.toThrowError()
     })
 
+    it('does not crash when class body contains unknown elements', async () => {
```
hugop95, line 14377: @azat-io 

With the `unknownClassElementParser` implementation suggested above, you can make this test more explicit:

```ts
it('does not crash when class body contains unknown elements', async () => {
  await validWithUnknownClassElement({
    code: dedent`
      class Example {
        a() {}
        anUnknownElement
        b() {}
      }
    `,
    parserOptions: {
      elementToMarkAsUnknown: 'anUnknownElement',
    },
    options: [{}],
  })
})
```

test/rules/sort-classes.test.ts
```diff
@@ -14314,6 +14363,32 @@ describe('sort-classes', () => {
...
+            static {}
+            a() {}
+            b() {}
+          }
+        `,
+        options: [{}],
+      })
+    })
+
+    it('does not reorder members across unknown elements', async () => {
```
hugop95: @azat-io 

With the `unknownClassElementParser` implementation suggested above, you can make this test more explicit (assuming that we choose to ignore unknown elements):

```ts
it('does not take into account unknown elements', async () => {
  await invalidWithUnknownClassElement({
    errors: [
      {
        messageId: 'unexpectedClassesOrder',
      },
    ],
    output: dedent`
      class Example {
        a() {}
        anUnknownElement
        b() {}
      }
    `,
    code: dedent`
      class Example {
        b() {}
        anUnknownElement
        a() {}
      }
    `,
    parserOptions: {
      elementToMarkAsUnknown: 'anUnknownElement',
    },
    options: [{}],
  })
})
```

rules/sort-classes/is-known-class-element.ts
azat-io: Nice idea. 👍 
Fixed

rules/sort-classes.ts
azat-io: Fixed

test/rules/sort-classes.test.ts
azat-io: Yes, agree with you. I fixed it

test/rules/sort-classes.test.ts
azat-io: 👍

hugop95 (commented):
@azat-io 🙏

hugop95 (approved):

Issue #695, opened 2026-02-11
Bug: declarationCommentMatchesPattern and callingFunctionNamePattern does not work


agarwalvaibhav0211:
### Describe the bug

In Tanstack Start, we need to _not_ re-order properties used for route creation for proper type inference to work. For that, I'm looking to disable perfectionist for all the createRouter calls. I tried with declarationCommentMatchesPattern and callingFunctionNamePattern. Both do not work

Ref: https://tanstack.com/router/latest/docs/eslint/create-route-property-order

### Code example

```ts
import { createFileRoute } from "@tanstack/react-router";
import { zodValidator } from "@tanstack/zod-adapter";
import { z } from "zod";

/* do not sort */
export const Route = createFileRoute("/items/$itemGroupDetails/")({
	component       : ()=>{
           return <>Dummy</>
},
	loaderDeps: ({ search: { limit, offset } }) => {
		return { limit, offset };
	},
	loader(ctx) {
	},
	validateSearch: zodValidator(z.object({
		limit : z.number().default(10),
		offset: z.number().default(0),
	}).default({
		limit : 10,
		offset: 0,
	})),
});
```

### ESLint version

9.39.2

### ESLint Plugin Perfectionist version

5.4.0

### Additional comments

The following Rule is configured

```ts
"perfectionist/sort-objects": [
                "warn",
                {
                    useConfigurationIf: {
                        declarationCommentMatchesPattern: {
                            pattern: 'do not sort',
                            scope: 'deep',
                        },
                    },
                    type: 'unsorted',
                },
                {
                    useConfigurationIf: {
                        callingFunctionNamePattern: {
                            pattern: `(createFileRoute\(\s*["'][^"']*["']\s*\))`,
                            scope: 'deep',
                        },
                    },
                    type: 'unsorted',
                },
                {
                    "ignoreCase": false,
                    "type": "natural",
                    "partitionByNewLine": true
                },
            ],
```

### Validations

- [x] Read the [docs](https://perfectionist.dev).
- [x] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

hugop95:
Hi @agarwalvaibhav0211 

There is indeed a bug with `declarationCommentMatchesPattern`

Regarding `callingFunctionNamePattern`, can you please try using it simply with `createFileRoute`?

agarwalvaibhav0211:
[...]

PR #696, opened 2026-02-11
fix(sort-objects): find declaration comments before exported variables

azat-io:
### Description

Handle comments before `export const` by reading them from `ExportNamedDeclaration` instead of `VariableDeclaration`.

### Additional context

#695 

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/test/rules/sort-objects.test.ts
+++ b/test/rules/sort-objects.test.ts
@@ -2682,88 +2682,6 @@ describe('sort-objects', () => {
         })
       })
 
-      it('matches callingFunctionNamePattern for chained function calls', async () => {
-        await valid({
-          options: [
-            {
-              ...options,
-              useConfigurationIf: {
-                callingFunctionNamePattern: 'foo',
-              },
-              type: 'unsorted',
-            },
-            {
-              type: 'alphabetical',
-            },
-          ],
-          code: dedent`
-            let a = foo("arg")({
-              b: "b",
-              a: "a",
-            })
-          `,
-        })
-
-        await valid({
-          options: [
-            {
-              ...options,
-              useConfigurationIf: {
-                callingFunctionNamePattern: {
-                  pattern: String.raw`^foo\(`,
-                  scope: 'deep',
-                },
-              },
-              type: 'unsorted',
-            },
-            {
-              type: 'alphabetical',
-            },
-          ],
-          code: dedent`
-            export const v = foo("arg")({
-              d: () => null,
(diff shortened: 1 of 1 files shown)


test/rules/sort-objects.test.ts
```diff
@@ -2682,6 +2682,88 @@ describe('sort-objects', () => {
...
+            export const v = foo("arg")({
+              d: () => null,
+              c: ({ search: { limit } }) => ({ limit }),
+              b() {},
+              a: () => ({}),
+            })
+          `,
+        })
+
+        await invalid({
```
hugop95: @azat-io 

IMO this test doesn't fit the `matches callingFunctionNamePattern for chained function calls` description (there is no chained call).

I'm not sure it's really a useful test considering the other ones in the `useConfigurationIf.callingFunctionNamePattern` block.

test/rules/sort-objects.test.ts
```diff
@@ -2682,6 +2682,88 @@ describe('sort-objects', () => {
         })
       })
 
+      it('matches callingFunctionNamePattern for chained function calls', async () => {
```
hugop95: @azat-io 

[nit] I would put this block after the `matches shallow and deep calls at the same time` test block just below, so it's not between shallow/deep-related tests ( `matches deep calls` and `matches shallow and deep calls at the same time`).

test/rules/sort-objects.test.ts
```diff
@@ -3448,6 +3530,99 @@ describe('sort-objects', () => {
...
 
+      it('matches declarationCommentMatchesPattern with exported declarations', async () => {
+        await valid({
+          options: [
+            {
+              ...options,
+              useConfigurationIf: {
+                declarationCommentMatchesPattern: {
+                  pattern: 'do not sort',
+                  scope: 'deep',
```
hugop95: @azat-io [nit] We can remove `scope: 'deep'` as a shallow scope will work as well here.

test/rules/sort-objects.test.ts
azat-io: Fixed ✅

test/rules/sort-objects.test.ts
azat-io: Fixed ✅

test/rules/sort-objects.test.ts
azat-io: Fixed ✅

Issue #407, opened 2024-12-02
Bug: sort-intersection-types conflicts with Prettier on object literal and parenthesized union 


JoshuaKGoldberg:
### Describe the bug

`npx eslint . --fix` on the following code autofixes to a layout that Prettier then autofixes right back. 

### Code example

Original code, as formatted by Prettier, with a lint report on the `{\n ... \n}`:

```ts
type ExtensionlessExportOrImport = (
	| ts.ExportDeclaration
	| ts.ImportDeclaration
) & {
	moduleSpecifier: ts.StringLiteral;
};
```

Code after `perfectionist/sort-intersection-types` autofixes it:

```ts
type ExtensionlessExportOrImport = {
	moduleSpecifier: ts.StringLiteral;
} & (
	| ts.ExportDeclaration
	| ts.ImportDeclaration
);
```

Code after formatting with Prettier:

```ts

type ExtensionlessExportOrImport = {
	moduleSpecifier: ts.StringLiteral;
} & (ts.ExportDeclaration | ts.ImportDeclaration);
//   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
// eslint(perfectionist/sort-intersection-types)
// Expected "ts.ExportDeclaration | ts.ImportDeclaration" to come before "{ moduleSpecifier: ts.StringLiteral;
```

### ESLint version

9.16.0

### ESLint Plugin Perfectionist version

4.1.2

### Additional comments

Not sure if the bug is here or in Prettier... 🤔 

### Validations

- [X] Read the [docs](https://perfectionist.dev).
- [X] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

hugop95:
Hi @JoshuaKGoldberg 👋

Could you send me the `.prettierrc` + `perfectionist` options that you use?

JoshuaKGoldberg:
D'oh, sorry! I got so used to quickly reporting issues I forgot to respect the checklist. 🙈 

Full repro here: https://github.com/JoshuaKGoldberg/repros/tree/repro-perfectionist-sort-intersection-types-conflict-prettier

```js
import js from "@eslint/js";
import perfectionist from "eslint-plugin-perfectionist";
import tseslint from "typescript-eslint";

export default tseslint.config(
  js.configs.recommended,
  tseslint.configs.recommended,
[...]

PR #698, opened 2026-02-17
fix: normalize parenthesized union and intersection member names

azat-io:
### Description

Fixes the Prettier conflict in `sort-intersection-types` by normalizing leading `|`/`&` before sorting.

### Additional context

#407

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


rules/sort-union-types.ts
```diff
@@ -223,7 +224,10 @@ export function sortUnionOrIntersectionTypes<MessageIds extends string>({
           break
       }
 
-      let name = sourceCode.getText(type)
+      let name = normalizeTypeMemberName({
```
hugop95: @azat-io 

[nit] I find it a bit more logical to have a `computeNodeName` function (like we have for other rules) that takes `type` and `sourceCode`.

```ts
export function computeNodeName({
  sourceCode,
  type,
}: {
  sourceCode: TSESLint.SourceCode
  type: TSESTree.TypeNode
}): string {
  let name = sourceCode.getText(type)

  if (
    type.type !== AST_NODE_TYPES.TSUnionType &&
    type.type !== AST_NODE_TYPES.TSIntersectionType
  ) {
    return name
  }

  return name.replace(/^\s*[&|][\s&|]*/u, '')
}
```

rules/sort-union-types/normalize-type-member-name.ts
```diff
@@ -0,0 +1,34 @@
...
+  name: string
+}): string {
+  if (
+    type.type !== AST_NODE_TYPES.TSUnionType &&
+    type.type !== AST_NODE_TYPES.TSIntersectionType
+  ) {
+    return name
+  }
+
+  return name.replace(/^\s*[&|][\s&|]*/u, '')
```
hugop95: @azat-io [q] Can't we simply do `^\s*[&|]\s*` ?

hugop95 (approved):

rules/sort-union-types.ts
azat-io: Fixed ✅

rules/sort-union-types/normalize-type-member-name.ts
azat-io: Fixed ✅

Issue #422, opened 2024-12-17
Feature: Customizable `sort-array`


hugop95:
### Describe the rule

I often build enums from constant arrays a lot in my projects:

```ts
export const NOTIFICATION_OPTIONS = ["userAdded", "userModified"] as const;
export type NotificationOption = typeof NOTIFICATION_OPTIONS[number];
```

I would like all of those arrays to be sorted.

Today, `sort-array-includes` and `sort-sets` don't affect them.

# Current rules

`sort-array-includes` and `sort-sets` generally do the same things: they sort `arrays` under different condition. This feature proposal is also related to sorting `arrays`, but with another different condition.

### Code example

# Proposal

Create a new rule `sort-array` while allowing users to choose the configuration they want depending on their needs. The other rules could be deprecated until next major version release.

Example configuration with `useConfigurationIf` (placeholder option names)

```ts
{
  'perfectionist/sort-array': [
    'error',
    {
      useConfigurationIf: {
        anyOf: [
		  {
			isSetDeclaration: true // Sort sets
		  },
		  {
			isArrayDeclaration: true,
			calledMethodMatchesPattern: "^includes$" // Sort array includes
		  },
		  {
			isConstantDeclaration: true,
			hasNoCalledMethod: true // Sort array enums
		  }
		]
      },
    },
	{
	  type: 'unsorted' // Fallback configuration: don't sort anything else
	}
  ],
}
```

### Validations

- [X] Read the [docs](https://perfectionist.dev).
- [X] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

hugop95:
@azat-io @OlivierZal 

What do you think about this? Curious to hear your opinions?

hugop95:
@azat-io We could keep `sort-array-includes` and `sort-sets` while also having `sort-arrays`, users would not have to migrate immediately.

OlivierZal:
@hugop95, I like the idea very much: powerful and I confess I would use it since there are many contexts where order doesn't matter.

Just wondering if `sort-arrays` isn't too scary, an array are ordered by definition, and it's maybe to narrow to reflect it could also sort `sets` and others. Maybe `sort-array-like`?

azat-io:
I think it's better to have two separate `sort-arrays` and `sort-sets` rules.

hugop95:
[...]

PR #701, opened 2026-02-22
feat(sort-array-includes): add `useConfigurationIf.matchesAstSelector` support

hugop95:
- Related to #422

## Description

ESLint allows the usage of [AST selectors](https://eslint.org/docs/latest/extend/selectors) to precisely target specific elements.

While very technical, it's also a very powerful tool. It also turns out that it's nicely built-in within ESLint itself, making it "relatively" easy to add.

I believe that we could make multiple rules benefit from it (in particular, `sort-objects` and `sort-object-types`).

This PR adds support for a new `useConfigurationIf.matchesAstSelector` option for `sort-array-includes` and `sort-sets`.

While it still only affects arrays using `includes` and sets, it will be much more useful for a broader [`sort-array`](https://github.com/azat-io/eslint-plugin-perfectionist/issues/422) rule.

## How to use

Users need to pass an AST selector that matches an `ArrayExpression` element.

## Example

Don't sort set elements that are declared as `const` variables.

```ts
{
  'perfectionist/sort-array-includes': [
    'error',
    {
      useConfigurationIf: {
        matchesAstSelector: 'VariableDeclaration[kind="const"] ArrayExpression',
      },
      type: 'unsorted'
    },
    {
    }
  ],
}
```

### DIFF
--- a/docs/content/rules/sort-array-includes.mdx
+++ b/docs/content/rules/sort-array-includes.mdx
@@ -299,6 +299,7 @@ This option is only applicable when [`partitionByNewLine`](#partitionbynewline) 
       | string[]
       | { pattern: string; flags: string }
       | { pattern: string; flags: string }[]
+    matchesAstSelector?: string
   }
   ```
 </sub>
@@ -334,6 +335,26 @@ Example configuration:
       useConfigurationIf: {
         allNamesMatchPattern: '^r|g|b$',
       },
+    },
+    {
+      type: 'alphabetical' // Fallback configuration
+    }
+  ],
+}
+```
+
+- `matchesAstSelector` — An [AST selector](https://eslint.org/docs/latest/extend/selectors) matching an `ArrayExpression` node.
+
+Example configuration: don't sort array-include elements that are declared as `const` variables.
+```ts
+{
+  'perfectionist/sort-array-includes': [
+    'error',
+    {
+      useConfigurationIf: {
+        matchesAstSelector: 'VariableDeclaration[kind="const"] ArrayExpression',
+      },
+      type: 'unsorted'
     },
     {
       type: 'alphabetical' // Fallback configuration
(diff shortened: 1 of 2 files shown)


rules/sort-array-includes.ts
```diff
@@ -78,6 +78,8 @@ export let jsonSchema: JSONSchema4 = {
 
 export default createEslintRule<Options, MessageId>({
   create: context => {
+    let alreadyParsedExpressions = new Set<TSESTree.Expression>()
```
hugop95, line 86: @reviewers

<details>
<summary>❓ Why do we need to keep track of parsed arrays?</summary>

If a given array is successfully parsed by a specific AST listener, other AST listeners should ignore it.

</details>

rules/sort-array-includes.ts
```diff
@@ -94,7 +97,8 @@ export default createEslintRule<Options, MessageId>({
 
     return {
       ...Object.fromEntries(allAstSelectorMatchers),
-      MemberExpression: buildFromMemberExpressionArraySorter({
+      'MemberExpression:exit': buildFromMemberExpressionArraySorter({
```
hugop95, line 105: @reviewers

<details>
<summary>❓ Why `:exit` ?</summary>

If the user adds a new AST listener from `userConfigurationIf.matchesAstSelector`, we end up with two listeners:

- `MemberExpression` (the default one).
- Their listener (for example, `ArrayExpression`).

We want the default listener to be called at the end, hence the `MemberExpression:exit`.

⚠️ Users shouldn't be using `:exit` in their own AST selectors, or correct behavior may not be guaranteed (the default listener might be called first).

</details>

rules/sort-array-includes.ts
```diff
@@ -144,157 +124,96 @@ export default createEslintRule<Options, MessageId>({
...
+    sortArrayFromMemberExpression({
+      alreadyParsedExpressions,
+      node: node.parent,
+      astSelector,
+      context,
+    })
   }
-
-  return computeArrayElements(memberExpression.object)
 }
```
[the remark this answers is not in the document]
hugop95: @coderabbitai

> The documentation mentions that the selector must match "an ArrayExpression node" and shows a correct descendant selector example, but it should be more explicit that invalid selectors produce no warning—they simply don't activate the custom configuration.

I find it explicit enough

PR #702, opened 2026-02-23
feat(sort-objects): add `useConfigurationIf.matchesAstSelector` support

hugop95:
- Similar PR: #701

## Description

This PR adds support for `useConfigurationIf.matchesAstSelector` option for `sort-objects`.

## How to use

Users need to pass an AST selector that matches an `ObjectExpression` or `ObjectPattern` (for destructured objects) node.

## Example

Don't sort objects that are declared as `const` variables.

```ts
{
    'error',
    {
      useConfigurationIf: {
        matchesAstSelector: 'VariableDeclaration[kind="const"] ObjectExpression',
      },
      type: 'unsorted'
    },
    {
    }
  ],
}
```

### DIFF
--- a/rules/sort-objects.ts
+++ b/rules/sort-objects.ts
@@ -4,6 +4,7 @@ import type { TSESTree } from '@typescript-eslint/types'
 import { AST_NODE_TYPES } from '@typescript-eslint/utils'
 
 import type { MessageId, Options } from './sort-objects/types'
+import type { Settings } from '../utils/get-settings'
 
 import {
   additionalCustomGroupMatchOptionsJsonSchema,
@@ -35,6 +36,7 @@ import { buildCommonGroupsJsonSchemas } from '../utils/json-schemas/common-group
 import { scopedRegexJsonSchema } from '../utils/json-schemas/scoped-regex-json-schema'
 import { defaultOptions, sortObject } from './sort-objects/sort-object'
 import { createEslintRule } from '../utils/create-eslint-rule'
+import { getSettings } from '../utils/get-settings'
 
 export default createEslintRule<Options, MessageId>({
   meta: {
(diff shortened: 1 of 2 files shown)


rules/sort-objects.ts
```diff
@@ -1,3 +1,8 @@
+import type { RuleContext } from '@typescript-eslint/utils/ts-eslint'
```
hugop95, line 1: Note to reviewers: there are likely utility functions we can extract, but I prefer to add at least a third use case before doing that.

rules/sort-objects.ts
```diff
@@ -340,76 +98,74 @@ export default createEslintRule<Options, MessageId>({
...
+        }),
+      'ObjectPattern:exit': node =>
+        sortObject({
+          alreadyParsedExpressions,
+          astSelector: null,
+          context,
+          node,
+        }),
+    }
+  },
```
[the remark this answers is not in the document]
hugop95: @coderabbitai It doesn't matter, it's the same listener.

rules/sort-objects/sort-object.ts
```diff
@@ -0,0 +1,268 @@
...
+  alreadyParsedNodes: Set<TSESTree.ObjectExpression | TSESTree.ObjectPattern>
+  context: Readonly<TSESLint.RuleContext<MessageId, Options>>
+  node: TSESTree.ObjectExpression | TSESTree.ObjectPattern
+  astSelector: string | null
+}): void {
+  if (!isSortable(node.properties)) {
+    return
+  }
+
+  let settings = getSettings(context.settings)
```
azat-io: Previously, `getSettings(context.settings)` was called once at the top of `create()`, so settings were computed a single time when the rule listener was created.

Now it is called inside `sortObject()`, which means it runs for every visited object node (`ObjectExpression` / `ObjectPattern`) in the file.

This is not entirely free because `getSettings()` also validates settings. Can we move it back to `create()` and pass the result down, as before?

hugop95: @azat-io Fair 👍

[`daa1379` (this PR)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/702/commits/daa1379dcf02d70f728451b50d26dab6498681b4)

PR #703, opened 2026-02-24
feat(sort-enums): add `useConfigurationIf.matchesAstSelector` support

hugop95:
## Description

This PR adds support for a `useConfigurationIf.matchesAstSelector` option for `sort-enums`.

## How to use

Users need to pass an AST selector that matches a `TSEnumDeclaration` node.

### DIFF
--- a/docs/content/rules/sort-enums.mdx
+++ b/docs/content/rules/sort-enums.mdx
@@ -302,7 +302,7 @@ Example configuration:
         },
       ],
       useConfigurationIf: {
-        allNamesMatchPattern: '^r|g|b$',
+        allNamesMatchPattern: '^[rgb]$',
       },
     },
     {
--- a/docs/content/rules/sort-array-includes.mdx
+++ b/docs/content/rules/sort-array-includes.mdx
@@ -333,7 +333,7 @@ Example configuration:
         },
       ],
       useConfigurationIf: {
-        allNamesMatchPattern: '^r|g|b$',
+        allNamesMatchPattern: '^[rgb]$',
       },
     },
     {
--- a/docs/content/rules/sort-interfaces.mdx
+++ b/docs/content/rules/sort-interfaces.mdx
@@ -381,7 +381,7 @@ Example configuration:
         },
       ],
       useConfigurationIf: {
-        allNamesMatchPattern: '^r|g|b$',
+        allNamesMatchPattern: '^[rgb]$',
       },
     },
     {
--- a/docs/content/rules/sort-jsx-props.mdx
+++ b/docs/content/rules/sort-jsx-props.mdx
@@ -317,7 +317,7 @@ Example configuration:
         },
       ],
       useConfigurationIf: {
-        allNamesMatchPattern: '^r|g|b$',
+        allNamesMatchPattern: '^[rgb]$',
       },
     },
     {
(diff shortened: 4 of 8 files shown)


docs/content/rules/sort-enums.mdx
```diff
@@ -257,6 +257,81 @@ or [`customGroups`](#customgroups) options.
...
+          elementNamePattern: '^g$',
+          groupName: 'g',
+        },
+        {
+          elementNamePattern: '^b$',
+          groupName: 'b',
+        },
+      ],
+      useConfigurationIf: {
+        allNamesMatchPattern: '^r|g|b$',
```
azat-io:
```suggestion
        allNamesMatchPattern: '^(r|g|b)$',
```

The regular expression in the example is incorrect.

```js
/^r|g|b$/.test('r-blabla') // true
/^r|g|b$/.test('any-word-with-g-letter') // true
/^r|g|b$/.test('test-b') //true
```

rules/sort-enums.ts
```diff
@@ -255,115 +82,67 @@ export default createEslintRule<Options, MessageId>({
     type: 'suggestion',
     fixable: 'code',
   },
+  create: context => {
+    let settings = getSettings(context.settings)
+
+    let alreadyParsedNodes = new Set<TSESTree.TSEnumDeclaration>()
+
+    let allAstSelectors = context.options
```
azat-io, line 90: Is it expected that a `matchesAstSelector` config can override an earlier generic config because of the separate listener pass and `alreadyParsedNodes`, despite the "first matching configuration" rule?`

docs/content/rules/sort-enums.mdx
hugop95: @azat-io 

I've fixed all occurrences:
- In tests: [`5330ae4` (this PR)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/703/commits/5330ae4131a8090d26d34e4a20a52c90f105c4b8)
- In the documentation: [`0470d99` (this PR)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/703/commits/0470d99e036c3def7f156f8bc184807845ee0c1e)

rules/sort-enums.ts
hugop95: @azat-io 

I would say that it's an acceptable limitation for now: The [ESLint documentation](https://eslint.org/docs/latest/extend/selectors#listening-for-selectors-in-rules) mentions the following:

> If two or more selectors match the same node, their listeners will be called in order of increasing specificity. The specificity of an AST selector is similar to the specificity of a CSS selector:
>
> - When comparing two selectors, the selector that contains more class selectors, attribute selectors, and pseudo-class selectors (excluding :not()) has higher specificity.
> - If the class/attribute/pseudo-class count is tied, the selector that contains more node type selectors has higher specificity.
> If multiple selectors have equal specificity, their listeners will be called in alphabetical order for that node.

So for now, users will want to make sure that the selectors with highest specificity (which gets called first) are referenced in the first configurations, and I think that it makes sense with how our configuration system works. 👍

It's possible to improve this, but we would need to:
- Parse all listeners without sorting first.
- Then in `Program:exit`, sort things.

I think it will be more interesting to investigate this improvement further in a second time.

PR #708, opened 2026-03-02
feat(sort-objects): add `partitionByComputedKey` option

### LINKED ISSUE #707: Feature: (sort-objects) Add ignoreComputedKeys option

sort-objects

### Describe the problem

I sometimes have computed properties in objects that would break if they were to be sorted. Do you have any plans to add an `ignoreComputedKeys` option to the `sort-objects` rule, similar to what they have in the ESLint `sort-keys` rule?

https://github.com/eslint/eslint/pull/19162

### Code example

```javascript
const key = 'propertyName';

const someObject = {
  propertyName: 'propertyValue',
  [key]: 'differentPropertyValue',
};
```

### Additional comments

_No response_

### Validations

hugop95:
- Resolves #707

## Description

This PR simply adds a new `partitionByComputedKey` (default: `false`) option to `sort-objects`.

### DIFF
--- a/docs/content/rules/sort-objects.mdx
+++ b/docs/content/rules/sort-objects.mdx
@@ -290,7 +290,20 @@ const user = {
 };
 ```
 
-Each group of keys (separated by empty lines) is treated independently, and the order within each group is preserved.
+### partitionByComputedKey
+
+<sub>default: `false`</sub>
+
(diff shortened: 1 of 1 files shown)


azat-io:
Thanks!

PR #711, opened 2026-03-05
feat(sort-named-exports): add `useConfigurationIf.matchesAstSelector` support

hugop95:
## Description

This PR adds support for a `useConfigurationIf.matchesAstSelector` option for `sort-named-exports`.

## How to use

Users need to pass an AST selector that matches a `ExportNamedDeclaration` node.

### DIFF
--- a/test/rules/sort-array-includes.test.ts
+++ b/test/rules/sort-array-includes.test.ts
@@ -1533,6 +1533,7 @@ describe('sort-array-includes', () => {
                 allNamesMatchPattern: '^[ab]$',
               },
               type: 'alphabetical',
+              order: 'desc',
             },
             {
               type: 'unsorted',
@@ -1541,22 +1542,22 @@ describe('sort-array-includes', () => {
           errors: [
             {
               data: {
-                right: 'a',
-                left: 'b',
+                right: 'b',
+                left: 'a',
               },
               messageId: 'unexpectedArrayIncludesOrder',
             },
           ],
           output: dedent`
             const a = [
-              a,
               b,
+              a,
             ].includes(value)
           `,
           code: dedent`
             const a = [
-              b,
               a,
+              b,
             ].includes(value)
           `,
         })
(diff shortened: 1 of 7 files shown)


rules/sort-named-exports/compute-matched-context-options.ts
```diff
@@ -0,0 +1,46 @@
...
+  let nodeNames = node.specifiers.map(specifier =>
+    computeNodeName(specifier, true),
+  )
+
+  let matchedContextOptions = filterOptionsByAllNamesMatch({
+    contextOptions: context.options,
+    nodeNames,
+  })
+
+  return matchedContextOptions.find(isContextOptionMatching)
```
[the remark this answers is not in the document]
hugop95: @coderabbitai I think it's OK for simplicity for now.

rules/sort-named-exports/sort-named-export.ts
```diff
@@ -0,0 +1,185 @@
...
+  alreadyParsedNodes: Set<TSESTree.ExportNamedDeclaration>
+  context: TSESLint.RuleContext<MessageId, Options>
+  node: TSESTree.ExportNamedDeclaration
+  astSelector: string | null
+  settings: Settings
+}): void {
+  if (!isSortable(node.specifiers)) {
+    return
+  }
+
```
azat-io, line 80: Should we add early return here?

```ts
  let matchedContextOptions = computeMatchedContextOptions({...})
  if (!matchedContextOptions && astSelector) { return } 
```

hugop95: @azat-io 

Indeed, thanks for catching that.
I realize that tests were covering this branch, but not failing. I've updated them:
- [`16d87e1` (this PR)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/711/commits/16d87e1ed89abb736405ccbcf94291135e4ce4db)
- [`33cd8de` (this PR)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/711/commits/33cd8defdb342f214d8fcc950411c901629a5c00)

PR #714, opened 2026-03-06
fix: make selector match priority deterministic across overlapping options

azat-io:
### Description

This PR fixes non-deterministic option selection when multiple `useConfigurationIf.matchesAstSelector` entries match the same AST node. 

Previously, the applied config depended on ESLint listener execution order rather than the position in `context.options`. 

Now, AST selector matches are collected during enter listeners and resolved once on `:exit`, ensuring stable "first matching option wins" precedence. Fallback behavior for options without `matchesAstSelector` is preserved.               

### Reviewer notes

N/A.

---

### Pre-merge checklist

### DIFF
--- a/test/rules/sort-named-exports.test.ts
+++ b/test/rules/sort-named-exports.test.ts
@@ -2902,6 +2902,116 @@ describe('sort-named-exports', () => {
           `,
         })
       })
+
+      it('applies first matching option when selectors overlap', async () => {
+        await valid({
+          options: [
+            {
+              ...options,
+              useConfigurationIf: {
+                matchesAstSelector: 'ExportNamedDeclaration',
+              },
+              type: 'unsorted',
+            },
+            {
+              ...options,
+              useConfigurationIf: {
+                matchesAstSelector: '* > ExportNamedDeclaration',
+              },
+              type: 'alphabetical',
+            },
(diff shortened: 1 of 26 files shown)


hugop95 (approved):
👍

test/rules/sort-named-exports.test.ts
```diff
@@ -2902,6 +2902,116 @@ describe('sort-named-exports', () => {
...
+          output: dedent`
+            export { a, b }
+          `,
+          code: dedent`
+            export { b, a }
+          `,
+        })
+      })
+
+      it('prioritizes selector-based options over fallback options', async () => {
```
hugop95, line 2964: Hi @azat-io 👋

I just realized this change here, and I'm not sure it's better: `useConfigurationIf` configurations have always been ordered, and the documentation explicitly says that *The first matching options configuration will be used*.

Here, `matchesAstSelector` bypasses that. I think that we should follow the general rule and apply the first configuration that matches all the time.

Do you agree?

---

Edit: PR fixing it: https://github.com/azat-io/eslint-plugin-perfectionist/pull/715

azat-io: Yes, agree. Looks good. 👍

PR #717, opened 2026-03-08
feat(sort-named-imports): add 2 `useConfigurationIf` options: `allNamesMatchPattern` and `matchesAstSelector`

hugop95:
## Description

This PR adds support for the following `sort-named-imports` options:
- `useConfigurationIf.allNamesMatchPattern`.
- `useConfigurationIf.matchesAstSelector`.

## How to use `useConfigurationIf.matchesAstSelector`

Users need to pass an AST selector that matches a `ImportDeclaration` node.

### DIFF
--- a/docs/content/rules/sort-named-exports.mdx
+++ b/docs/content/rules/sort-named-exports.mdx
@@ -218,7 +218,8 @@ Specifies filters to match a particular options configuration for a given named 
 
 The first matching options configuration will be used. If no configuration matches, the default options configuration will be used.
 
-- `allNamesMatchPattern` — A regexp pattern that all named export names must match.
+- `allNamesMatchPattern` — A regexp pattern that all named export names must match. If [`ignoreAlias`](#ignorealias)
+is `true`, the pattern will be tested against the export aliases, otherwise against the local names.
 
 Example configuration:
 ```ts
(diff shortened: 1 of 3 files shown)


rules/sort-named-imports/compute-matched-context-options.ts
```diff
@@ -0,0 +1,60 @@
...
+  context: Readonly<RuleContext<MessageIds, Options>>
+  matchedAstSelectors: ReadonlySet<string>
+  node: TSESTree.ImportDeclaration
+}): Options[number] | undefined {
+  let nodeNames = node.specifiers
+    .filter(
+      (specifier): specifier is TSESTree.ImportSpecifier =>
+        specifier.type === 'ImportSpecifier',
+    )
+    .map(specifier => computeNodeName(specifier, true))
```
azat-io: Should `allNamesMatchPattern` respect `ignoreAlias` option here? Right now this is hardcoded to true (`computeNodeName(specifier, true)`), which can diverge from actual sorting behavior when `ignoreAlias` is `false`.

hugop95: @azat-io 

I initially wanted to do that in a separate PR as it needs a refactor, but we can do that here as well and update both `sort-named-imports` and `sort-named-exports`.

[`3297dc6` (this PR)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/717/commits/3297dc6766263b0ebc53eaccdf514526272960da)
[`b641574` (this PR)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/717/commits/b641574363508641faad5ba4adf06774c1153ddd)

PR #718, opened 2026-03-09
feat(sort-object-types): add `useConfigurationIf.matchesAstSelector` support

hugop95:
## Description

This PR adds support for a `useConfigurationIf.matchesAstSelector` option for `sort-object-types` and `sort-interfaces`.

## How to use

Users need to pass an AST selector that matches a `TSTypeLiteral` or `TSInterfaceDeclaration` node.

### DIFF
--- a/test/rules/sort-array-includes.test.ts
+++ b/test/rules/sort-array-includes.test.ts
@@ -1359,7 +1359,7 @@ describe('sort-array-includes', () => {
     })
 
     describe('useConfigurationIf.matchesAstSelector', () => {
-      it('matches configuration based off matchesAstSelector', async () => {
+      it('skips config when selector does not match the sorted node type', async () => {
         await invalid({
           options: [
             {
@@ -1392,7 +1392,9 @@ describe('sort-array-includes', () => {
             ].includes(value)
           `,
         })
+      })
 
+      it('applies config when selector matches the sorted node type', async () => {
         await valid({
           options: [
             {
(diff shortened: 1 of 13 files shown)


azat-io (approved):
Looks good to me. Thank you!

We probably want similar `useConfigurationIf.matchesAstSelector` support for the following rules as well:

- sort-decorators
- sort-export-attributes
- sort-exports
- sort-import-attributes
- sort-imports
- sort-intersection-types
- sort-modules
- sort-switch-case
- sort-union-types

Should I wait or is it time to release it?

hugop95:
@azat-io 

> We probably want similar `useConfigurationIf.matchesAstSelector` support for the following rules as well:

Unfortunately, not all remaining rules will be able to benefit from it: `useConfigurationIf.matchesAstSelector` targets rules that have a common "parent" node to match (in `sort-objects`, it's the object itself for example).

Rules like `sort-imports` are "file-scoped" and a group of imports doesn't really have a parent to match.

So IMO, the remaining rules to update are:
- `sort-export-attributes`.
- `sort-import-attributes`.
- `sort-intersection-types`.
- `sort-union-types`.

`sort-decorators` could potentially work, but:
- It would probably need additional work as there are multiple decorator types.
- It's not the most important rule.

So I don't think it's a blocker not to release the option for it ASAP.

> Should I wait or is it time to release it?

The 4 remaining rules should be quick to update, I'll open some PRs today. I think we can wait a few days and ship all of them at once 👍

PR #721, opened 2026-03-10
docs(integrations): add VS Code formatter info

unimorphic:
### Description

After recently discovering this awesome ESLint plugin, I found myself wanting to use it as a formatter in VS Code but still also wanting to use Prettier, so I made a VS Code extension to do this: https://github.com/unimorphic/vscode-format-att

Added info to the docs for using the ESLint VS Code extension as a formatter with and without my extension

---

### Pre-merge checklist

- [ ] Tests added/updated when needed

### DIFF
--- a/docs/content/guide/integrations.mdx
+++ b/docs/content/guide/integrations.mdx
@@ -48,11 +48,14 @@ Then add the following code to your `settings.json`:
 
 ### Formatter
 
-The ESLint VS Code extension can also be used as a formatter:
+The ESLint VS Code extension can also be used as a formatter. This example also sets it as the default formatter for TypeScript:
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


azat-io:
Thanks for the PR and for your interest in the plugin!

I don't have a problem with self-promotion by itself, but I'm not convinced this belongs in our docs yet. Format ATT is currently at version 0.0.1 with essentially no adoption on the Marketplace, so it feels too early to present it as a recommended integration.

I'm also not sure we should add `eslint.format.enable` guidance here — our current VS Code docs are based on `codeActionsOnSave`, and I'd prefer to keep that as the documented approach for now.

I'd rather skip this addition for now. If Format ATT gains more traction and establishes itself, we can definitely revisit this later.

PR #724, opened 2026-03-23
fix: guard getAllComments return value against undefined

StephenODea54:
### Description

Hey! I ran into this while using `@antfu/eslint-config` (related issue: antfu/eslint-config#837). When ESLint processes non-JS files (e.g. markdown code blocks via `@eslint/markdown`), `sourceCode.getAllComments()` can return `undefined`, causing a crash:

```
TypeError: sourceCode.getAllComments is not a function or its return value is not iterable
```

The fix is a `?? []` guard. I ripped this pattern from `utils/get-node-decorators.ts`, for what I assume is the same reason, so it felt like the right approach here too.

I didn't open an issue beforehand since the fix is so small, but happy to do so if that's a hard requirement. First OSS contribution so feel free to roast me.

### Reviewer notes

The fix is a single line in `utils/get-eslint-disabled-lines.ts`. The test covers the specific case where `getAllComments` returns `undefined`.

---

### Pre-merge checklist

### DIFF
--- a/test/utils/get-eslint-disabled-lines.test.ts
+++ b/test/utils/get-eslint-disabled-lines.test.ts
@@ -1,0 +1,24 @@
+import type { TSESLint } from '@typescript-eslint/utils'
+
+import { describe, expect, it } from 'vitest'
+
+import { getEslintDisabledLines } from '../../utils/get-eslint-disabled-lines'
+
+describe('getEslintDisabledLines', () => {
+  it('returns empty array when getAllComments returns undefined', () => {
+    expect(
+      getEslintDisabledLines({
+        sourceCode: getSourceCodeMock(undefined),
+        ruleName: 'perfectionist/sort-imports',
+      }),
+    ).toStrictEqual([])
+  })
+})
+
+function getSourceCodeMock(
+  comments: undefined,
+): TSESLint.SourceCode {
+  return {
+    getAllComments: () => comments,
+  } as unknown as TSESLint.SourceCode
+}
--- a/utils/get-eslint-disabled-lines.ts
+++ b/utils/get-eslint-disabled-lines.ts
@@ -52,7 +52,8 @@ export function getEslintDisabledLines(props: {
   let { sourceCode, ruleName } = props
   let returnValue: number[] = []
   let lineRulePermanentlyDisabled: number | null = null
-  for (let comment of sourceCode.getAllComments()) {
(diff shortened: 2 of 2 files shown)


StephenODea54:
> Hi @StephenODea54
> 
> * The issue was introduced when a dependency project was misconfigured. The root cause is that JS-only rules are being applied to Markdown files without restricting the files scope in the flat config, and IMO the fix should come from that project, or we would just be treating the symptom here.
> * `getAllComments()` is a JS-specific method that simply doesn't exist on `@eslint/markdown`'s SourceCode implementation. If another ESLint utility behaves the same way for another non-JS language plugin, will that mean we will keep adding guard clauses indefinitely?
> 
> For comparison, PR #385 (the decorator guard) makes more functional sense to me: it handles projects that legitimately don't use TypeScript, which is a real compatibility concern, not a configuration mistake.
> 
> With that being said, not fixing it here would mean `perfectionist` wouldn't run any of its rules in Markdown files containing JS. I'm not convinced it's a big use-case, but I could hear that argument pushing toward making a fix here.
> 
> That's just my opinion, @azat-io might view things differently!

Hey, thanks for the detailed feedback!

Honestly, I was on the fence about opening this PR in the first place. I mainly drew the parallel to [#385](https://github.com/azat-io/eslint-plugin-perfectionist/pull/385) and figured a similar guard made sense here, but your breakdown made me realize the two situations are actually pretty different. The node.decorators fix in 385 is a genuine parser compatibility issue, whereas this is really a symptom of JS rules being applied to Markdown files without proper scoping in the consumer's config.

Makes sense for the fix to belong upstream. Happy to close this out if that's the move, just let me know!

azat-io:
Thanks for the PR and welcome to OSS! I agree with @hugop95 - this looks more like an upstream/config scoping issue than a compatibility bug in Perfectionist. I'd rather not add a one-off guard here without deciding on broader non-JS SourceCode support, so I'm going to close this PR. Hope to see you back with another contribution!

StephenODea54:
> Thanks for the PR and welcome to OSS! I agree with @hugop95 - this looks more like an upstream/config scoping issue than a compatibility bug in Perfectionist. I'd rather not add a one-off guard here without deciding on broader non-JS SourceCode support, so I'm going to close this PR. Hope to see you back with another contribution!

Sounds good! Appreciate the review!

Issue #685, opened 2026-01-13
Feature: (sort-modules) add `arrow` modifier for function selector


SimonGolms:
### Describe the rule

The `sort-modules` rule currently supports sorting function declarations with modifiers like `declare`, `default`, `async`, and `export` (e.g., `async-function`, `export-default-function`). However, arrow functions defined as variable declarations (e.g., `export const myFunc = () => {}`) are not sorted at all by this rule.

I propose adding an `arrow` modifier to the `function` selector, allowing arrow functions to be included in the sorting alongside regular function declarations. This would enable better organization and provide granular control over the ordering of different function styles.

For example: `arrow-function`, `export-arrow-function`, `async-arrow-function`, `export-async-arrow-function`, etc.

### Code example

**Current behavior:**

```ts
// Arrow functions are NOT sorted - they remain in original position
export const fetchUser = async () => { /* ... */ };

export function processData() { /* ... */ }

export const validateInput = (input: string) => { /* ... */ };

export async function handleRequest() { /* ... */ }
```

**Expected behavior with the following configuration**

```js
// In configuration:
{
  groups: [
    ['enum'],
    ['interface', 'type'],
    ['class'],
    ['arrow-function'],
    ['function'],
    ['unknown']
  ]
}
```

```ts
export const fetchUser = async () => { /* ... */ };
export const validateInput = (input: string) => { /* ... */ };

export function processData() { /* ... */ }
export async function handleRequest() { /* ... */ }
```

This would allow teams to:

- Group all functions together: `['function', 'arrow-function']`
- Separate function declarations from arrow functions: `['function'], ['arrow-function']`
- Mix and match with existing modifiers: `['async-function', 'async-arrow-function']`

### Additional comments

- Arrow functions exported as const declarations (`export const myFunc = () => {}`) are extremely common in modern JavaScript/TypeScript codebases
- Currently, these are not sorted at all by the `sort-modules` rule
- Adding arrow as a modifier maintains consistency with the existing modifier pattern for functions (`async`, `export`, `declare`, `default`)
- This approach provides maximum flexibility - teams can group arrow functions with regular functions or separate them as needed
- Would work seamlessly with existing modifiers: `export-arrow-function`, `async-arrow-function`, etc.

### Validations

- [x] Read the [docs](https://perfectionist.dev).
[...]

PR #727, opened 2026-03-26
feat: support arrow-function modifier for const arrow declarations

AndrewDongminYoo:
## Summary

- Adds `arrow` modifier to `sort-modules` rule, enabling `arrow-function`, `export-arrow-function`, and `async-arrow-function` group patterns
- Only `const` declarations with a single `Identifier` declarator and `ArrowFunctionExpression` initializer are recognized; `let`/`var` and multi-declarator forms remain partition boundaries
- Adds `TSESTree.VariableDeclaration` to `SortModulesNode` to reflect that const arrow declarations are now valid sorting nodes
- Updates `defaultOptions.groups` to bundle `export-arrow-function` with `export-function` and `arrow-function` with `function`

## Test plan

- [x] `sorts arrow functions` — basic alphabetical ordering of const arrow functions
- [x] `sorts exported arrow functions` — `export const` arrow functions
- [x] `sorts async arrow functions` — `const foo = async () => {}`
- [x] `separates exported and non-exported arrow functions by group` — verifies `export` modifier is correctly assigned via the `ExportNamedDeclaration` path
- [x] `still creates partitions at let/var arrow function declarations` — non-`const` forms remain partition boundaries
- [x] `still creates partitions at non-arrow/destructured/multi-declarator variable declarations`
- [x] 100% coverage on `sort-modules.ts` and `rules/sort-modules/`

### DIFF
--- a/rules/sort-modules.ts
+++ b/rules/sort-modules.ts
@@ -83,8 +83,8 @@ let defaultOptions: Required<Options[number]> = {
     'class',
     'export-class',
     'declare-function',
-    ['export-function', 'export-arrow-function'],
-    ['function', 'arrow-function'],
+    'export-function',
+    'function',
   ],
   useExperimentalDependencyDetection: true,
   fallbackSort: { type: 'unsorted' },
@@ -201,7 +201,6 @@ function analyzeModule({
       case AST_NODE_TYPES.TSModuleDeclaration:
         break
       case AST_NODE_TYPES.VariableDeclaration:
-        break
       case AST_NODE_TYPES.ExpressionStatement:
         sortingNodeGroupsWithoutOverloadSignature.push([])
         continue
--- a/rules/sort-modules/compute-node-details.ts
+++ b/rules/sort-modules/compute-node-details.ts
@@ -143,17 +143,16 @@ export function computeNodeDetails({
         if (
           nodeToParse.declarations.length === 1 &&
           nodeToParse.kind === 'const' &&
-          declarator !== undefined &&
           declarator.id.type === AST_NODE_TYPES.Identifier &&
           declarator.init?.type === AST_NODE_TYPES.ArrowFunctionExpression
         ) {
-          let arrowFn = declarator.init
+          let arrowFunction = declarator.init
           selector = 'function'
           modifiers.push('arrow')
-          if (arrowFn.async) {
+          if (arrowFunction.async) {
             modifiers.push('async')
           }
-          name = declarator.id.name
+          ;({ name } = declarator.id)
           addSafetySemicolonWhenInline = true
         } else {
           shouldPartitionAfterNode = true
--- a/test/rules/sort-modules.test.ts
+++ b/test/rules/sort-modules.test.ts
@@ -358,67 +358,6 @@ describe('sort-modules', () => {
       })
     })
 
-    it('separates exported and non-exported arrow functions by group', async () => {
-      await invalid({
-        options: [
-          {
-            ...options,
-            groups: ['export-arrow-function', 'arrow-function'],
-          },
-        ],
-        errors: [
-          {
-            data: {
-              rightGroup: 'export-arrow-function',
-              leftGroup: 'arrow-function',
-              right: 'b',
-              left: 'a',
-            },
-            messageId: 'unexpectedModulesGroupOrder',
-          },
-        ],
-        output: dedent`
-          export const b = () => {}
-          const a = () => {}
-        `,
-        code: dedent`
-          const a = () => {}
-          export const b = () => {}
-        `,
-      })
-    })
-
-    it('does not sort arrow functions when arrow-function groups are not configured', async () => {
-      await valid({
-        options: [
-          {
-            ...options,
-            groups: [],
-          },
-        ],
-        code: dedent`
-          const b = () => {}
-          const a = () => {}
-        `,
-      })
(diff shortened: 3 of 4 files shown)


rules/sort-modules.ts
```diff
@@ -83,8 +83,8 @@ let defaultOptions: Required<Options[number]> = {
     'class',
     'export-class',
     'declare-function',
-    'export-function',
-    'function',
+    ['export-function', 'export-arrow-function'],
```
hugop95: @AndrewDongminYoo 

IMO changing the default configuration would be a breaking change and should be kept for a major version upgrade.

rules/sort-modules/compute-node-details.ts
```diff
@@ -138,9 +138,28 @@ export function computeNodeDetails({
...
         moduleBlock = nodeToParse.body ?? null
         break
-      case AST_NODE_TYPES.VariableDeclaration:
-        shouldPartitionAfterNode = true
+      case AST_NODE_TYPES.VariableDeclaration: {
+        let [declarator] = nodeToParse.declarations
+        if (
+          nodeToParse.declarations.length === 1 &&
+          nodeToParse.kind === 'const' &&
+          declarator !== undefined &&
```
hugop95: @AndrewDongminYoo  `declarator` can't be `undefined` (cf linting errors).

test/rules/sort-modules.test.ts
```diff
@@ -274,6 +274,215 @@ describe('sort-modules', () => {
...
+          const a = () => {}
+        `,
+        code: dedent`
+          const a = () => {}
+          export const b = () => {}
+        `,
+      })
+    })
+
+    it('does not sort arrow functions when arrow-function groups are not configured', async () => {
```
hugop95: @AndrewDongminYoo I don't think this test is interesting (we could apply this logic to any other selector/modifier).

test/rules/sort-modules.test.ts
```diff
@@ -274,6 +274,215 @@ describe('sort-modules', () => {
...
+          const b = async () => {}
+        `,
+        code: dedent`
+          const b = async () => {}
+          const a = async () => {}
+        `,
+      })
+    })
+
+    it('separates exported and non-exported arrow functions by group', async () => {
```
hugop95: @AndrewDongminYoo 

I think this test can be improved:
- We can rename it to `prioritizes arrow modifier over async modifier for arrow functions` and place it before to the `prioritizes async modifier over export modifier for functions`. 
- It will test the modifiers priority.

```ts
    it('prioritizes arrow modifier over async modifier for arrow functions', async () => {
      await invalid({
        errors: [
          {
            data: {
              rightGroup: 'arrow-function',
              leftGroup: 'interface',
              left: 'Interface',
              right: 'f',
            },
            messageId: 'unexpectedModulesGroupOrder',
          },
        ],
        options: [
          {
            ...options,
            groups: ['arrow-function', 'interface', 'async-function'],
          },
        ],
        output: dedent`
          const f = async () => {}

          interface Interface {}
        `,
        code: dedent`
          interface Interface {}

          const f = async () => {}
        `,
      })
    })
```

test/rules/sort-modules.test.ts
```diff
@@ -274,6 +274,215 @@ describe('sort-modules', () => {
...
+          },
+        ],
+        code: dedent`
+          const b = () => {}
+          const a = () => {}
+        `,
+      })
+    })
+
+    it('still creates partitions at non-arrow variable declarations', async () => {
```
hugop95: @AndrewDongminYoo [nit] We could remove `still` to those tests to align with the existing `creates partitions at function call expressions` test.

rules/sort-modules.ts
```diff
@@ -201,6 +201,7 @@ function analyzeModule({
       case AST_NODE_TYPES.TSModuleDeclaration:
         break
       case AST_NODE_TYPES.VariableDeclaration:
+        break
```
hugop95: @AndrewDongminYoo 

This can create issues due to dependencies. Here are a few examples that come to mind (with `groups: ["unknown"]`)

```ts
const b = () => {}
class a {
  static {
    b()
  }
}
```

```ts
const b = () => {}
class a {
  static value = b();
}
```

```ts
const a = () => c()
class b {
  static {
    a()
  }
}
const c = () => 42
```

AndrewDongminYoo:
Thanks for the detailed review.

I understand the concern around introducing variable-declaration sorting into sort-modules, especially regarding compilation/runtime safety.

Given that, I agree this approach is too risky in its current form. I’ll revert the changes related to the arrow modifier and clean up the remaining ESLint issues.

If helpful, I can open a separate discussion later to explore whether a safe subset of variable-declaration sorting is possible, or if it belongs in a separate rule.

hugop95:
@AndrewDongminYoo 

> If helpful, I can open a separate discussion later to explore whether a safe subset of variable-declaration sorting is possible, or if it belongs in a separate rule.

Feel free to add things to the dedicated issue: https://github.com/azat-io/eslint-plugin-perfectionist/issues/685 if you want.

I think it functionally fits this rule. It's just technically challenging because variable declarations create a ton of explicit and implicit dependencies at runtime which we ideally want to cover completely.

An alternative would be to allow users to enable this option "at their own risk", but that's not ideal.

PR #728, opened 2026-03-26
fix(sort-modules): add missing partition cases

hugop95:
### Description

`sort-modules` doesn't handle several AST node types that can appear at module or namespace level and ignore them. In theory, this can create issues (see the tests).

This PR adds the following list of new node types that now trigger partitioning:

- `TSImportEqualsDeclaration`
- `TSExportAssignment`
- `DoWhileStatement`
- `LabeledStatement`
- `SwitchStatement`
- `WhileStatement`
- `ForInStatement`
- `ForOfStatement`
- `ThrowStatement`
- `BlockStatement`
- `ForStatement`
- `TryStatement`
- `IfStatement`

### DIFF
--- a/rules/sort-modules.ts
+++ b/rules/sort-modules.ts
@@ -199,8 +199,6 @@ function analyzeModule({
       case AST_NODE_TYPES.ImportDeclaration:
       case AST_NODE_TYPES.DebuggerStatement:
       case AST_NODE_TYPES.ContinueStatement:
-      /* v8 ignore next 4 -- @preserve These statements cannot appear at module/namespace level in valid code. */
-      // eslint-disable-next-line no-fallthrough
       case AST_NODE_TYPES.ReturnStatement:
       case AST_NODE_TYPES.EmptyStatement:
       case AST_NODE_TYPES.BreakStatement:


rules/sort-modules.ts
```diff
@@ -208,7 +232,9 @@ function analyzeModule({
       case AST_NODE_TYPES.TSEnumDeclaration:
       case AST_NODE_TYPES.ClassDeclaration:
         break
+      /* v8 ignore next 2 -- @preserve Exhaustive guard. */
       default:
+        assertIsNever(node)
```
hugop95, line 237: reviewers

I'm not doing `throw new UnreachableCaseError` to preserve current behavior in case custom parsers return other types.

test/rules/sort-modules.test.ts
```diff
@@ -285,6 +285,204 @@ describe('sort-modules', () => {
...
+    it('creates partitions at TypeScript import-equals declarations', async () => {
+      await valid({
+        code: dedent`
+          enum B { b = 'b' }
+          import A = B
+          enum C { c = 'c' }
+        `,
+        options: [options],
+      })
+    })
```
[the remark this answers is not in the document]
hugop95: Fixed.

rules/sort-modules.ts
```diff
@@ -193,22 +194,49 @@ function analyzeModule({
   >[][] = [[]]
   for (let node of module.body) {
     switch (node.type) {
+      case AST_NODE_TYPES.TSNamespaceExportDeclaration:
+      case AST_NODE_TYPES.ExportAllDeclaration:
+      case AST_NODE_TYPES.ImportDeclaration:
+      case AST_NODE_TYPES.DebuggerStatement:
+      case AST_NODE_TYPES.ContinueStatement:
+      /* v8 ignore next 4 -- @preserve These statements cannot appear at module/namespace level in valid code. */
```
azat-io: Is this `v8 ignore` still correct? `EmptyStatement` is valid at top level (`;`), so the comment seems too broad.

test/rules/sort-modules.test.ts
```diff
@@ -285,6 +285,204 @@ describe('sort-modules', () => {
...
+    it('sorts across empty statements', async () => {
+      await valid({
+        code: dedent`
+          enum A { a = 'a' }
+          ;
+          enum B { b = 'b' }
+        `,
+        options: [options],
+      })
+    })
```
azat-io, line 446: Could these `sorts across ...` cases be turned into `invalid(...)` tests with `B ... skipped node ... A`?

Right now they use already sorted input, so they still pass even if those skipped nodes accidentally become partition boundaries. Coverage is green, but I don't think those tests actually pin the behavior down yet.

hugop95: @azat-io 

Good point, there are indeed useless as of now!

[`4379a9c` (this PR)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/728/commits/4379a9c80db271a347291a6a2d98b9e001026e2a) 🤦

rules/sort-modules.ts
hugop95: @azat-io 

It was from an early implementation, we are testing everything now, it's useless 👍

[`20a9a99` (this PR)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/728/commits/20a9a99ef34f8012b5a99b778bafbbf7eecb7a66)

PR #731, opened 2026-04-04
feat: export class option types

### LINKED ISSUE #730: Feature: Expose types for rule option verification
I'd like to see the types in `/types` exposed so they can be used programmatically.

### Code example

**Crude** example: I'm creating a shareable config that uses Perfectionist. However, some of the rules I want to use are going to be common. But I'd also like the rule options to be verified against the types this plugin uses internally.

```ts
// src/configs/perfectionist.ts
import type { CommonOptions, TypeOption } from 'eslint-plugin-perfectionist/types/common-options';

type PerfectionistOptions = Omit<CommonOptions<TypeOption>, 'alphabet'>;

const commonOptions = {
    fallbackSort: { order: 'asc', type: 'alphabetical' },
    ignoreCase: false,
    order: 'asc',
    type: 'natural',
    // ...
} satisfies PerfectionistOptions;

export const perfectionist = async (options: OptionsPefectionist): Promise<FlatConfigItem> => {
    return [
        {
            name: 'config/perfectionist',
            rules: {
                'perfectionist/sort-classes': ['error', { ...commonOptions }],
                'perfectionist/sort-decorators': ['error', { ...commonOptions }],
                'perfectionist/sort-interfaces': ['error', { ...commonOptions }],
                'perfectionist/sort-named-exports': ['error', { ...commonOptions }],
                'perfectionist/sort-named-imports': ['error', { ...commonOptions }],
            },
        },
    ];
};
```

However, because `CommonOptions` and `TypeOption` aren't exposed as types, I have to add a `path` to my `tsconfig.json` for it to resolve.

I don't mind the long path when I do `import type`, I'd just prefer my `tsconfig.json` to be "clean", if that makes sense.

### Additional comments

_No response_

### Validations

hugop95:
- Resolves #730.

## Description

This PR exposes each rule's `Options`.

Other internal types are not exposed as in theory, exposed types should avoid being modified to prevent breaking changes.

## Example

```ts
import type { SortNamedExportsOptions } from 'eslint-plugin-perfectionist';
```

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -147,3 +147,27 @@ export default {
   ...plugin,
   configs,
 } as { configs: PluginConfigs } & ESLint.Plugin
+
+export type { Options as SortVariableDeclarationsOptions } from './rules/sort-variable-declarations/types'
+export type { Options as SortIntersectionTypesOptions } from './rules/sort-intersection-types/types'
+export type { Options as SortImportAttributesOptions } from './rules/sort-import-attributes/types'
+export type { Options as SortExportAttributesOptions } from './rules/sort-export-attributes/types'
+export type { Options as SortHeritageClausesOptions } from './rules/sort-heritage-clauses/types'
+export type { Options as SortArrayIncludesOptions } from './rules/sort-array-includes/types'
+export type { Options as SortNamedImportsOptions } from './rules/sort-named-imports/types'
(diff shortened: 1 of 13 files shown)


azat-io (approved):
Thanks!

PR #732, opened 2026-04-05
fix(sort-imports): add submodule imports as internal patterns

### LINKED ISSUE #723: Bug: sort-imports does not recognise submodule imports as internal
### Describe the bug

When using [subpath imports](https://nodejs.org/api/packages.html#subpath-imports) the internal imports start with a `#` character.  At present, these are not treated as internal, and so you end up with imports like this:

```js
import type { Logger } from "#src/shared/cli/logger.js";

import { DefaultAzureCredential } from "@azure/identity";
```

But I'd prefer them to be like this:

```js
import { DefaultAzureCredential } from "@azure/identity";

import type { Logger } from "#src/shared/cli/logger.js";
```

### Code example

It's possible to customise in the `eslint.config.js` like so:

```js
export default defineConfig([
  {
    files: ["**/*.js", "**/*.ts"],
    rules: {
      "perfectionist/sort-imports": [
        "error",
        {
          internalPattern: ["^#.+"],
        },
      ],
    },
  },
]);
```

Should this be the default behaviour?

### ESLint version

10.0.2

5.6.0

### Additional comments

_No response_

### Validations

hugop95:
- Fixes #723

### Description

This PR marks submodule imports as `internal` by default.

This may affect existing configurations, but [as it's considered a bug](https://github.com/azat-io/eslint-plugin-perfectionist/issues/723#issuecomment-4095607392), this is deemed acceptable.

`internalPattern` went from `['^~/.+', '^@/.+']` to `['^~/.+', '^@/.+', '^#.+']`.


azat-io (approved):
Thanks!

PR #737, opened 2026-05-02
docs: add agent-ready support

azat-io:
### Description

I checked the website’s agent readiness:
https://isitagentready.com/perfectionist.dev

This pull request improves the documentation site’s accessibility for AI agents. Specifically, it:

- adds Markdown content negotiation for docs pages on Netlify
- adds Link response headers for agent discovery
- adds Content-Signal directives to robots.txt
- adds machine-readable Markdown versions of documentation pages, including the homepage and 404 responses
- adds a local Netlify CLI workflow so agent-related behavior can be built, tested, and deployed consistently

Result after the changes:

isitagentready com_perfectionist dev

---

### Pre-merge checklist


docs/utils/mdx-to-markdown.ts
```diff
@@ -0,0 +1,313 @@
...
+const TRANSFORMERS: Record<
+  string,
+  ((node: JsxElement) => Node[] | null) | undefined
+> = {
+  CodeExample: transformCodeExample,
+  Important: transformImportant,
+  CodeTabs: transformCodeTabs,
+}
+
+export async function mdxToMarkdown(mdx: string): Promise<string> {
```
hugop95, line 65: @azat-io 

I think it would be interesting to add full tests for that, just so we know that nothing is broken/forgotten when modifying the docs.

I was thinking of something similar to that:

```ts
import { describe, expect, it } from 'vitest'
import fs from 'node:fs/promises'
import path from 'node:path'

import { mdxToMarkdown } from '../mdx-to-markdown.ts'

describe('mdx-to-markdown', () => {
  // Maybe adding some unit tests too?

  it('should convert mdx files to markdown correctly', async () => {
    let sortClasses = await fs.readFile(
      path.join(process.cwd(), './docs/content/rules/sort-classes.mdx'), // Text documentation files
      { encoding: 'utf8' },
    )

    let result = await mdxToMarkdown(sortClasses)

    let expected = await fs.readFile(
      path.join(process.cwd(), './docs/utils/tests/sort-classes.md'), // Place the expected md files somewhere
      { encoding: 'utf8' },
    )
    expect(result).to.equal(expected)
  })
})
```

The question is what would be the best ROI between: 
- Having a test for every doc page.
  - ✅ Exhaustive: we have to make sure each time a doc is updated, its agent `md` file looks good.
  - ⚠️ Painful: each doc change means updating those tests. 
- Having a test for one doc page chosen arbitrarily (`sort-classes` for example).
  - ✅ Doesn't require updating tests most of the time.
  - ✅ Should probably catch most of the unexpected changes as most doc files follow similar patterns. 
- Having a test for a random `md` file we generate.
  - ✅ Updating the doc doesn't require updating tests.
  - ⚠️ Weak.

docs/pages/index.md.ts
```diff
@@ -0,0 +1,139 @@
+import type { APIRoute } from 'astro'
+
+import { getCollection } from 'astro:content'
+
+const TAGLINE = 'Take Your Code to a Beauty Salon'
+
+const DESCRIPTION =
+  'Automatically sort and organize objects, imports, types, enums, and JSX props. Ensure a clean and maintainable codebase with minimal effort.'
```
hugop95: @azat-io 

Here and in `guide.md.ts`/`configs.md.ts`/`rules.md.ts`, we're duplicating content that is already in `astro` files.

I'm wondering if it would be interesting to put string content in a dedicated file and use them both for agent queries and WEB through Astro.

Otherwise, we need to remember to update multiple places when updating general docs.

hugop95 (commented):
@azat-io 

Interesting topic! I'm just wondering if it would be worth adding tests regarding the conversion from `mdx` to `md`. It's not as simple as I would have expected.

Additionally, I've added a comment about future maintainability, considering that we duplicate content that already exists.

docs/utils/mdx-to-markdown.ts
azat-io: I thought about it and decided to keep tests scoped to the plugin's functional code - I don't want to maintain tests for docs tooling. The conversion was verified by running it over all 30 MDX content pages (zero dropped code examples, no JSX artifacts) and end-to-end against netlify serve

docs/pages/index.md.ts
azat-io: Done: shared content now lives in `docs/data/home.ts` / `docs/data/collections.ts`, and `llms.txt` is generated from the same collections

azat-io:
@hugop95 I think everything's fine now, and it makes sense to merge this

hugop95:
@azat-io 

Sorry, I didn't see the comments! I'll take a look soon.

docs/utils/mdx-to-markdown.ts
hugop95: @azat-io Fair!

hugop95 (approved):

PR #738, opened 2026-05-13
feat: new rule - `sort-constructor-parameters`

hugop95:
# Description

This rule proposal allows users to sort class constructors in a similar fashion than `sort-arrays`:
- The rule is disabled by default (not present in recommended configs), and it **requires** a defined `useConfigurationIf` option.
- The implementation is quite similar overall (arrays and function parameters have a lot in common).

I came up with this use case when going back on an old Angular project: frameworks providing dependency injection generally allow constructor parameters to be entered in any order (Angular, NestJS, TSyringe, ...).

```ts
@Component({
  imports: [ButtonModule],
  ...
})
export class MyComponent {
  public constructor(
    private readonly _i18nService: I18nService,
    private readonly _snackService: SnackService
  ) {}
```

In my case, I automatically sort relevant constructors with the following `matchesAstSelector`: `ClassDeclaration:has(Decorator[expression.callee.name=/^(Component|Injectable|Pipe)$/]) > ClassBody > MethodDefinition[kind="constructor"]`. (Note that `[kind="constructor"]` is not necessary in reality for the rule).

---

## Selectors

- `parameter`

##  Modifiers

- `'decorated'` — Matches parameters with at least one decorator.
- `'override'` — Matches parameters declared with the `override` keyword.
- `'readonly'` — Matches parameters declared with the `readonly` keyword.
- `'protected'` — Matches parameters declared with the `protected` keyword.
- `'private'` — Matches parameters declared with the `private` keyword.
- `'public'` — Matches parameters declared with the `public` keyword, or without visibility keyword.
- `'optional'` — Matches optional parameters.

## Other support

- `decoratorNamePattern`.


hugop95:
Hi @azat-io 👋

I came up with this new rule idea while working on an old Angular side project, the idea is basically to allow users to sort class constructors of their choosing, similarly to `sort-arrays`, by leveraging `useConfigurationIf.matchesAstSelector`.

What do you think of the idea?

---

In theory, we could expand this to a `sort-function-parameters`/`sort-method-parameters` rule, but I don't see a use case in which function/method parameters ordering has no impact, other than in those constructors.

---

Side note: I'll be in holidays for about a month, I might take some time to answer! 🏖️

azat-io:
Sorry for the late reply - I was on an extended vacation.

Thanks for the proposal! The implementation looks solid overall, and I can see this being useful for Angular/NestJS-style DI constructors.

The use case is a bit niche, and I probably wouldn't use it personally, but I like the idea as an opt-in rule.

Before merging, I think we should tighten a couple of safety details:

- parameter decorators should be preserved/moved correctly, especially for cases like `@Inject(TOKEN) service`, since that is exactly the DI use case this rule targets
- default values that depend on earlier parameters should not be reordered if that can change runtime behavior

Also, since the rule sorts constructor parameters rather than constructors themselves, I’d consider renaming it before release, for example to `sort-constructor-parameters`.

Overall: I’m open to this, but I’d like these safety cases covered with tests first. 👍

readme.md
```diff
@@ -187,32 +187,32 @@ module.exports = {
...
-| [sort-variable-declarations](https://perfectionist.dev/rules/sort-variable-declarations) | Enforce sorted variable declarations                                   | 🔧  | ✅  |
+| Name                                                                                       | Description                                                            | 🔧  | ✅  |
+| :----------------------------------------------------------------------------------------- | :--------------------------------------------------------------------- | :-- | :-- |
+| [sort-arrays](https://perfectionist.dev/rules/sort-arrays)                                 | Enforce sorted arrays based on custom conditions                       | 🔧  |     |
+| [sort-array-includes](https://perfectionist.dev/rules/sort-array-includes)                 | Enforce sorted arrays before include method                            | 🔧  | ✅  |
+| [sort-constructor-parameters](https://perfectionist.dev/rules/sort-constructor-parameters) | Enforce sorted class constructor parameters based on custom conditions | 🔧  |     |
```
azat-io: Let's sort them alphabetically

hugop95: @azat-io 

Oops, fixed in [`362e1af` (this PR)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/738/commits/362e1aff52e1acd997c4befa0b85b20e2acfe0cb)

---

I'll fix the other issues soon!

hugop95:
Temporarily opening to get an early review from the rabbit 🐇.

rules/sort-constructor-parameters.ts
```diff
@@ -0,0 +1,148 @@
...
+  | typeof ORDER_ERROR_ID
+
+let defaultOptions: Required<Options[number]> = {
+  fallbackSort: { type: 'unsorted' },
+  newlinesInside: 'newlinesBetween',
+  specialCharacters: 'keep',
+  partitionByComment: false,
+  partitionByNewLine: false,
+  newlinesBetween: 'ignore',
+  useConfigurationIf: {},
```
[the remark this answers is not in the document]
hugop95: @coderabbitai 

[`bae12f7` (this PR)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/738/commits/bae12f7499b52ec4c7e948b23d6e3e8d879f49da)

`sort-arrays` doesn't enforce `matchesAstSelector`, and I think we shouldn't either, in case the user wants to use `allNamesMatchPattern`.

rules/sort-constructor-parameters.ts
```diff
@@ -0,0 +1,149 @@
...
+    messages: {
+      [DEPENDENCY_ORDER_ERROR_ID]: DEPENDENCY_ORDER_ERROR,
+      [MISSED_SPACING_ERROR_ID]: MISSED_SPACING_ERROR,
+      [EXTRA_SPACING_ERROR_ID]: EXTRA_SPACING_ERROR,
+      [GROUP_ORDER_ERROR_ID]: GROUP_ORDER_ERROR,
+      [ORDER_ERROR_ID]: ORDER_ERROR,
+    },
+    docs: {
+      url: 'https://perfectionist.dev/rules/sort-constructor-parameters',
+      description: 'Enforce sorted constructor parameters parameters.',
```
azat-io:
```suggestion
      description: 'Enforce sorted constructor parameters.',
````

File: index.ts [hunk 4]
```diff
@@ -1,6 +1,7 @@
 import type { ESLint, Linter } from 'eslint'
 
 import { version as packageVersion, name as packageName } from './package.json'
+import SortConstructorParameters from './rules/sort-constructor-parameters'
```
azat-io: Why PascalCase, not camelCase?

test/rules/sort-constructor-parameters.test.ts
```diff
@@ -0,0 +1,7140 @@
+import { createRuleTester } from 'eslint-vitest-rule-tester'
```
azat-io, line 1: Maybe we should add a few more tests:

```ts
  it('constructor keeps its body when a trailing line comment moves last', async () => {
    let { result } = await invalid({
      code: dedent`
        class Foo {
          constructor(
            b, // b comment
            a) {}
        }
      `,
      options: [options], // { useConfigurationIf: {}, type: 'alphabetical', order: 'asc' }
    })
    let { ast } = typescriptParser.parseForESLint(result.output!)
    let classDeclaration = ast.body[0] as unknown as {
      body: { body: [{ value: { body: unknown } }] }
    }
    expect(classDeclaration.body.body[0].value.body).toBeTruthy()
  })
```

```ts
  it('trailing block comment stays attached to its parameter inside the list', async () => {
    await invalid({
      errors: [{ messageId: 'unexpectedConstructorParametersOrder', data: { right: 'a', left: 'b' } }],
      code: dedent`
        class Foo {
          constructor(
            b, /* b comment */
            a) {}
        }
      `,
      output: dedent`
        class Foo {
          constructor(
            a,
            b /* b comment */) {}
        }
      `,
      options: [options],
    })
  })
```

```ts
  it('does not break working code when a closure captures a later parameter', async () => {
    await valid({
      code: dedent`
        class Foo {
          constructor(b = () => a, a = b) {}
        }
      `,
      options: [options],
    })
  })
```

```ts
  it('does not report dependency order for a closure-only forward reference', async () => {
    await valid({
      code: dedent`
        class Foo {
          constructor(a = () => z, z = 1) {}
        }
      `,
      options: [options],
    })
  })
```

```ts
  it('leaves unknown-group parameters in their original order', async () => {
    await valid({
      code: dedent`
        class Foo {
          constructor(b, a) {}
        }
      `,
      options: [{ ...options, groups: ['private-parameter'] }],
    })
  })
```

```ts
  it('BUG-4a: skips constructor overload signatures', async () => {
    await valid({
      code: dedent`
        class Foo {
          constructor(b: B, a: A);
          constructor(a: any, b?: any) {}
        }
      `,
      options: [options],
    })
  })
```

```ts
  it('skips declare-class constructor signatures', async () => {
    await valid({
      code: dedent`
        declare class Foo {
          constructor(b: B, a: A);
        }
      `,
      options: [options],
    })
  })
```

rules/sort-constructor-parameters.ts
hugop95: @azat-io 

Thanks!

[`296be92` (this PR)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/738/commits/296be921962a458ea964dc2ba79a132f6b5f935d)

index.ts
hugop95: @azat-io 

Oops!

[`2955f24` (this PR)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/738/commits/2955f2474cd3ba5774253d7d13d7fd3152202eef)

test/rules/sort-constructor-parameters.test.ts
hugop95: Tests/fixes coming soon.

Issue #739, opened 2026-06-08
Bug: (sort-modules) `usage` order `desc` doesn't work with diamond deps


Keyruu:
### Describe the bug

`sort-modules` with `type: 'usage'` + `order: 'desc'` enters an infinite autofix loop when module members form diamond dependency patterns (a leaf node referenced by multiple callers).

### Code example

```ts
export interface BuildOptions {
  seed?: number;
}

export type Grid = string[];

export function build(opts: BuildOptions): Grid {
  const rng = prng(opts.seed ?? 42);
  const candidates = select(opts);
  return shuffle(candidates, rng);
}

export function load(filename: string): string {
  const text = filename;
  try {
    return text;
  } catch (err) {
    return formatPath(err);
  }
}

function select(opts: BuildOptions): string[] {
  const rng = prng(opts.seed ?? 42);
  return ['a', 'b', 'c'];
}

function shuffle<T>(arr: T[], _rng: () => number): T[] {
  return [...arr];
}

function prng(a: number): () => number {
  return () => {
    a = (a + 0x6d2b79f5) | 0;
    return a / 0x100000000;
  };
}

function formatPath(err: unknown): string {
  return String(err);
}
```

### ESLint version

v10.4.1

### ESLint Plugin Perfectionist version

v5.9.0

### Validations

- [x] Read the [docs](https://perfectionist.dev).
- [x] Check that there isn't [already an issue](https://github.com/azat-io/eslint-plugin-perfectionist/issues) that reports the same bug to avoid creating a duplicate.

azat-io:
Fixed in b649b49bca47d44ad5f786882778e42b9478a3af.
Released in [v5.9.1](https://github.com/azat-io/eslint-plugin-perfectionist/releases/tag/v5.9.1).

PR #740, opened 2026-06-15
fix(sort-modules): prevent usage-sort autofix loop with order desc

azat-io:
### Description

`sort-modules` with type: `'usage'` + `order: 'desc'` caused an infinite autofix loop on diamond-shaped dependencies. 

The usage comparator negated the dependency-resolved index for `desc`, which conflicted with the unconditional forward-topological post-pass - the order flipped on every pass. 

#739

### Reviewer notes

N/A. 

---

### Pre-merge checklist

### DIFF
--- a/rules/sort-modules/build-usage-comparator.ts
+++ b/rules/sort-modules/build-usage-comparator.ts
@@ -1,14 +1,13 @@
 import type { TSESLint } from '@typescript-eslint/utils'
 
 import type { Comparator } from '../../utils/compare/default-comparator-by-options-computer'
-import type { SortModulesSortingNode, SortModulesNode, Options } from './types'
+import type { SortModulesSortingNode, SortModulesNode } from './types'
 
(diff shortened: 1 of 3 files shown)


rules/sort-modules/build-usage-comparator.ts
```diff
@@ -59,16 +55,18 @@ export function buildUsageComparator({
...
-      sortedOrderA - sortedOrderB,
-      options.order,
-    )
-    let unsortedOrderedValue = computeOrderedValue(
-      unsortedOrderA - unsortedOrderB,
-      options.order,
-    )
-
-    if (sortedOrderedValue !== unsortedOrderedValue) {
+    /**
```
hugop95: [very nit] [subject] I think this comment isn't really useful as we have tests covering that case, so it's quite self-explicit 🤷.

hugop95 (approved):

rules/sort-modules/build-usage-comparator.ts
azat-io: Fixed

PR #741, opened 2026-06-23
fix(sort-modules): allow sorting of decorated exported classes

hugop95:
- Related issues:
  - https://github.com/azat-io/eslint-plugin-perfectionist/issues/376
  - https://github.com/azat-io/eslint-plugin-perfectionist/pull/675 

## Description

ESLint's parser doesn't seem to detect (https://github.com/estree/estree/issues/315) the decorator ranges correctly with the following syntax:

### Code example

```ts
class B {}

@A
export class A {}
```

This will result in

```ts
export class A {}

@A
class B {}
```

Because of that, decorated and exported classes are not sorted in `sort-modules`.

Support of this is necessary for the upcoming [`sort-constructor-parameters`](https://github.com/azat-io/eslint-plugin-perfectionist/pull/738), as a similar issue exists here with parameter decorators.

The idea is to allow rules to pass an `implicitDecorators` field to `SortingNode`, referencing decorators that are not directly associated with the evaluated node.

When `implicitDecorators` are passed, we examine the first of these decorators and adjust the permutation range when sorting nodes.

### DIFF
--- a/test/rules/sort-modules.test.ts
+++ b/test/rules/sort-modules.test.ts
@@ -3478,34 +3478,46 @@ describe('sort-modules', () => {
         })
       })
 
-      it('ignores decorated then exported classes when sorting', async () => {
+      it('correctly sorts exported classes with decorators', async () => {
         await invalid({
           output: dedent`
-            @B
-            class B {}
-
-            @A
+            @A1
+            @A2
+            // A
             export class A {}
 
-            @C
+            @B1
+            @B2
+            /* B */
+            class B {}
+
+            @C1
+            @C2
+            // C
             class C {}
+          `,
+          code: dedent`
+            @C1
+            @C2
+            // C
+            class C {}
+
+            @A1
+            @A2
+            // A
+            export class A {}
+
+            @B1
+            @B2
+            /* B */
+            class B {}
           `,
           errors: [
             {
               messageId: 'unexpectedModulesOrder',
-              data: { right: 'B', left: 'C' },
+              data: { right: 'A', left: 'C' },
             },
           ],
-          code: dedent`
-            @C
-            class C {}
-
-            @A
-            export class A {}
-
-            @B
-            class B {}
-          `,
           options: [
             {
               ...options,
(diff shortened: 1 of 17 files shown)


test/rules/sort-modules.test.ts
```diff
@@ -8311,34 +8341,40 @@ describe('sort-modules', () => {
...
+          @B2
           class BB {}
         `,
+        errors: [
+          {
+            messageId: 'unexpectedModulesOrder',
+            data: { right: 'AAA', left: 'C' },
+          },
+        ],
         options: [
```
azat-io, line 8498: Should we add tests about saving comments?

```ts
    it('keeps a line comment attached to its exported decorated class', async () => {
      await invalid({
        code: dedent`
          @Bd
          class B {}

          // doc for A
          @Ad
          export class A {}
        `,
        output: dedent`
          // doc for A
          @Ad
          export class A {}

          @Bd
          class B {}
        `,
        options: [{ ...options, groups: ['unknown'] }],
        errors: [{ messageId: 'unexpectedModulesOrder', data: { left: 'B', right: 'A' } }],
      })
    })

    it('keeps a JSDoc block comment attached to its exported decorated class', async () => {
      await invalid({
        code: dedent`
          /** doc for B */
          @Bd
          class B {}

          /** doc for A */
          @Ad
          export class A {}
        `,
        output: dedent`
          /** doc for A */
          @Ad
          export class A {}

          /** doc for B */
          @Bd
          class B {}
        `,
        options: [{ ...options, groups: ['unknown'] }],
        errors: [{ messageId: 'unexpectedModulesOrder', data: { left: 'B', right: 'A' } }],
      })
    })
```

And maybe something like this?

```ts
    it('matches an exported decorated class by decoratorNamePattern', async () => {
      await invalid({
        code: dedent`
          function func() {}

          @HelloDecorator
          export class C {}
        `,
        output: dedent`
          @HelloDecorator
          export class C {}

          function func() {}
        `,
        options: [
          {
            ...options,
            customGroups: [
              { groupName: 'classesWithHello', decoratorNamePattern: 'Hello', selector: 'class' },
            ],
            groups: ['classesWithHello', 'unknown'],
          },
        ],
        errors: [
          {
            messageId: 'unexpectedModulesGroupOrder',
            data: { left: 'func', leftGroup: 'unknown', right: 'C', rightGroup: 'classesWithHello' },
          },
        ],
      })
    })

    it('already-sorted decorated exports are valid', async () => {
      await valid({
        code: dedent`
          @Ad
          export class A {}

          @Bd
          class B {}
        `,
        options: [{ ...options, groups: ['unknown'] }],
      })
    })
```

hugop95: @azat-io 

> Should we add tests about saving comments?

Good catch, there is a side-effect when a comment is above an exported decorated class. We need to look at the comments above that decorator in that case.

[`1d60d39` (this PR)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/741/commits/1d60d399b1b534533be832796b616cc07eb2c918)

---

> And maybe something like this?

Added 👍

[`c2e9074` (this PR)](https://github.com/azat-io/eslint-plugin-perfectionist/pull/741/commits/c2e9074d0a7dc4581cd8028f3c20dcfc04083bc6)

azat-io (approved):

PR #744, opened 2026-07-03
feat(sort-intersection-types): add ignore callable types option

### LINKED ISSUE #743: Bug: sort-intersection-types breaks TypeScript function overload semantics by reordering function intersections
### Describe the bug

In TypeScript, intersection types are generally commutative, meaning A & B is identical to B & A. However, there is a strict exception: function types.

When function signatures are intersected, TypeScript treats them as function overloads and evaluates them strictly from left to right. Therefore, the order of the operands directly affects type inference.

Currently, the sort-intersection-types rule sorts these function intersections mechanically (e.g., alphabetically or by line length). This auto-fix silently changes the order of the operands, which breaks the overload resolution priority and alters the inferred return types. Since this changes the actual behavior of the code rather than just its stylistic formatting, it is quite dangerous.

### Code example

```typescript
type SpecificFunc = (x: string) => string;
type AnyFunc = (x: any) => any;

// Before auto-fix
// The developer intentionally placed `SpecificFunc` first for higher priority.
type OverloadedFunc = SpecificFunc & AnyFunc;
declare const myFunc: OverloadedFunc;
const res1 = myFunc("hello"); // Inferred as `string`

// After auto-fix
// Alphabetical sorting forces `AnyFunc` to come first, altering the semantics.
type OverloadedFunc = AnyFunc & SpecificFunc;
declare const myFunc2: OverloadedFunc;
const res2 = myFunc2("hello"); // Inferred as `any`, breaking the original type strictness
```

### ESLint version

v9.39.4

v5.9.1

### Additional comments

Since reordering function intersections alters the core semantics of TypeScript code, it might be safer to completely ignore sorting when the operands are function types, or provide a way/option to opt-out function intersections.

Thank you for maintaining this amazing plugin! I really appreciate your work.

### Validations

hugop95:
- Resolves #743

### Description

This PR adds a `ignoreCallableTypes` (by default `true`) option to `sort-intersection-types` to ignore sorting intersection types that contain inline function types.

> [!IMPORTANT]
> Named types are not considered as callable types, and as such will still be sorted.


azat-io (approved):
Cool 👍 
Thank you!

PR #745, opened 2026-07-05
fix(sort-switch-case): skip condition-shaped switch statements

azat-io:
### Description

Switches whose discriminant or any case test is condition-shaped (boolean literal, negation, comparison, or logical expression) are no longer sorted or reported: there the case order is the program logic, because the first matching case wins.

Example:

```js
function grade(score) {
  switch (true) {
    case score >= 90: return 'A';
    case score >= 80: return 'B';
  }
  return 'C';
}

function gradeSorted(score) { // 'score >= 80' < 'score >= 90'
  switch (true) {
    case score >= 80: return 'B';
    case score >= 90: return 'A';
  }
  return 'C';
}

console.log(grade(95));       // 'A'
console.log(gradeSorted(95)); // 'B' 
```

---

### Pre-merge checklist

### DIFF
--- a/rules/sort-switch-case.ts
+++ b/rules/sort-switch-case.ts
@@ -29,6 +29,19 @@ const ORDER_ERROR_ID = 'unexpectedSwitchCaseOrder'
 
 type MessageId = typeof ORDER_ERROR_ID
 
+let conditionOperators = new Set<string>([
+  'instanceof',
+  '===',
+  '!==',
+  '==',
+  '!=',
+  '<=',
+  '>=',
+  'in',
+  '<',
+  '>',
+])
+
 let defaultOptions: Required<Options[number]> = {
   fallbackSort: { type: 'unsorted' },
   specialCharacters: 'keep',
@@ -54,10 +67,13 @@ export default createEslintRule<Options, MessageId>({
       validateCustomSortConfiguration(options)
 
       let { sourceCode } = context
-      let isDiscriminantTrue =
-        switchNode.discriminant.type === AST_NODE_TYPES.Literal &&
-        switchNode.discriminant.value === true
-      if (isDiscriminantTrue) {
+      let isConditionCaseSwitch =
+        isConditionExpression(switchNode.discriminant) ||
+        switchNode.cases.some(
+          caseNode =>
+            caseNode.test !== null && isConditionExpression(caseNode.test),
+        )
+      if (isConditionCaseSwitch) {
         return
       }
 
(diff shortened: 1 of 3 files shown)


rules/sort-switch-case.ts
```diff
@@ -324,6 +340,32 @@ function reduceCaseSortingNodes(
...
+ *
+ * @param node - The expression AST node to check.
+ * @returns True if the expression is condition-shaped.
+ */
+function isConditionExpression(node: TSESTree.Expression): boolean {
+  switch (node.type) {
+    case AST_NODE_TYPES.LogicalExpression:
+      return true
+    case AST_NODE_TYPES.BinaryExpression:
+      return conditionOperators.has(node.operator)
```
hugop95: @azat-io 

I think we can simplify this and return `true` for any binary expression. I don't see a case where that wouldn't be the case.

rules/sort-switch-case.ts
```diff
@@ -324,6 +340,32 @@ function reduceCaseSortingNodes(
...
+ *
+ * Condition-shaped expressions produce boolean values: boolean literals,
+ * logical negations, comparison binary expressions and logical expressions.
+ * Switch statements built on such expressions (e.g. `switch (true)`) encode
+ * their program logic in the case order, so sorting them is unsafe.
+ *
+ * @param node - The expression AST node to check.
+ * @returns True if the expression is condition-shaped.
+ */
+function isConditionExpression(node: TSESTree.Expression): boolean {
```
hugop95: @azat-io 

Could we put that in it's own file in the `sort-switch-case` folder?

test/rules/sort-switch-case.test.ts
```diff
@@ -1012,6 +1012,40 @@ describe('sort-switch-case', () => {
         options: [{}],
       })
     })
+
+    it('keeps autofix even when reordered cases may match the same value', async () => {
```
hugop95: @azat-io 

[nit] I find `sorts even ...` better than `keeps autofix`.

test/rules/sort-switch-case.test.ts
```diff
@@ -3273,6 +3435,106 @@ describe('sort-switch-case', () => {
       })
     })
 
+    describe('condition-shaped switches', () => {
+      it('not works if discriminant and case tests are comparisons', async () => {
```
hugop95: @azat-io 

[nit] `does not sort switches where [...]` is a better than `not works` IMO.

azat-io: Done ✅

test/rules/sort-switch-case.test.ts
azat-io: Done ✅

rules/sort-switch-case.ts
azat-io: Done ✅

rules/sort-switch-case.ts
azat-io: Done ✅

hugop95 (approved):

PR #746, opened 2026-07-05
fix: prevent reordering members with equal runtime keys

azat-io:
### Description

`sort-objects`, `sort-maps` and `sort-jsx-props` autofixes could reorder members whose keys are provably equal at runtime (`16` vs `0x10`, `'a'` vs `` `a` ``, `1` vs `+1`, `undefined` vs `void 0`, `['b']` vs `b`, duplicate props), silently changing last-write-wins behavior:

```js
const original = { b: 1, b() { return 2 } };
console.log('original.b:', typeof original.b);  // "function"
console.log('original.b():', original.b());     // 2
```

```js
const autofixed = { b() { return 2 }, b: 1 };
console.log('autofixed.b:', typeof autofixed.b); // "number"
console.log('autofixed.b():', autofixed.b());    // 💥 TypeError: autofixed.b is not a function
```

```js
const originalMap = new Map([[16, 'decimal'], [0x10, 'hex']]);
console.log(originalMap.get(16)); // "hex" (0x10 === 16)

const autofixedMap = new Map([[0x10, 'hex'], [16, 'decimal']]);
console.log(autofixedMap.get(16)); // "decimal"
```

---

### Pre-merge checklist

### DIFF
--- a/utils/compute-static-key-value.ts
+++ b/utils/compute-static-key-value.ts
@@ -1,0 +1,104 @@
+import type { TSESTree } from '@typescript-eslint/types'
+
+import { AST_NODE_TYPES } from '@typescript-eslint/utils'
+
+/**
+ * Computes the statically-known runtime value of a key expression.
+ *
+ * Resolves expressions whose runtime value can be determined without
+ * evaluation: literals (including bigint), template literals without
+ * expressions, unary minus over numeric or bigint literals, unary plus over
+ * numeric literals, `void` expressions and the `undefined` identifier. Regex
+ * literals are excluded because each evaluation produces a distinct object.
+ *
+ * The value is wrapped in an object so that a statically-known `null` or
+ * `undefined` value can be distinguished from "not statically known".
+ *
+ * @example
+ *
+ * ```ts
+ * // Key expression: 0x10
+ * computeStaticKeyValue(node) // Returns: { value: 16 }
+ * ```
+ *
+ * @example
+ *
+ * ```ts
+ * // Key expression: `a${x}`
+ * computeStaticKeyValue(node) // Returns: null
+ * ```
+ *
+ * @param node - Key expression to resolve.
+ * @returns Wrapped static value, or null if the value is not statically known.
+ */
+export function computeStaticKeyValue(
+  node: TSESTree.Node,
+): { value: unknown } | null {
+  switch (node.type) {
+    case AST_NODE_TYPES.TemplateLiteral:
+      return node.expressions.length === 0 ?
+          { value: node.quasis[0]!.value.cooked }
+        : null
+    case AST_NODE_TYPES.UnaryExpression:
+      return computeUnaryStaticKeyValue(node)
+    case AST_NODE_TYPES.Identifier:
+      return node.name === 'undefined' ? { value: undefined } : null
+    case AST_NODE_TYPES.Literal:
+      return 'regex' in node ? null : { value: node.value }
+    default:
+      return null
+  }
+}
+
+/**
+ * Computes the static value of a unary expression.
+ *
+ * `void` expressions always evaluate to `undefined` regardless of their
+ * operand. Unary minus resolves over numeric or bigint literals, and unary plus
+ * over numeric literals (unary plus over a bigint throws at runtime).
+ *
+ * @param node - Unary expression to resolve.
+ * @returns Wrapped static value, or null if the value is not statically known.
+ */
+function computeUnaryStaticKeyValue(
+  node: TSESTree.UnaryExpression,
+): { value: unknown } | null {
+  switch (node.operator) {
+    case 'void':
+      return { value: undefined }
+    case '-':
+      return computeNegatedStaticKeyValue(node)
+    case '+':
+      return (
+          node.argument.type === AST_NODE_TYPES.Literal &&
+            typeof node.argument.value === 'number'
+        ) ?
+          { value: node.argument.value }
+        : null
+    default:
+      return null
+  }
+}
+
+/**
+ * Computes the static value of a unary minus expression over a numeric or
+ * bigint literal.
+ *
+ * @param node - Unary expression to resolve.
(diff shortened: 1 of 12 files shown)


hugop95 (commented):
@azat-io 

I haven't reviewed everything yet: to be honest, I find this fix overkill:

- Typescript reports objects that share the same keys, and so will ESLint.
  - ```ts
    const a = {
      b: 4,
      // TS1119: An object literal cannot have property and accessor with the same name.
      // ESLint: Duplicate key 'b'. (no-dupe-keys)
      b() {},
    }
    ```
  - While not all users use Typescript, I think that the probability a user only uses JS, doesn't have the ESLint rule checking this (but has `perfectionist`), and also has duplicate keys written in different ways in the object to be almost impossible, and not worth the large amount of code this PR adds.
- Similarly to `maps`, I think it's the responsibility of the user not to add two keys that are the same, but written in a different way.
  - I don't see a valid use case where this would happen.

IMO, the fact that we have never got an issue about it speaks for itself: it's not a use case that exists in practice, and while I understand we should push to be "as safe as possible", I think it needs to be done while also looking at the maintenance cost: in this case, it's a lot of code for a situation that wouldn't appear in reality.

Let me know what you think!

azat-io:
@hugop95 Fair points - for objects and JSX props duplicates are already caught by `no-dupe-keys`, TS and `react/jsx-no-duplicate-props`, so I agree the guard there isn't worth the code.

The exception is `sort-maps`: nothing warns about `new Map([[16, x], [0x10, y]])`, and `--fix` silently flips `m.get(16)`.

Proposal: I'll drop the guard from `sort-objects` and `sort-jsx-props` and keep only a simple check for `sort-maps` - skip sorting when entries have equal runtime keys. That removes most of the added code.

utils/compute-static-key-value.ts
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * ```ts
+ * // Key expression: `a${x}`
+ * computeStaticKeyValue(node) // Returns: null
+ * ```
+ *
+ * @param node - Key expression to resolve.
+ * @returns Wrapped static value, or null if the value is not statically known.
+ */
+export function computeStaticKeyValue(
```
hugop95, line 34: @azat-io 

I think we can simplify this: what `computeStaticKeyValue` does is really close to a combination of `computeNodeName` and `computeExpressionNumberValue` from `sort-enums`. Reusing `computeExpressionNumberValue` also fixes a case missed with the current logic: it will allow us to detec `4 * 4` and `16` as colliding too.

We could merge `computeStaticKeyValue` and `computeNodeName` into a new `computeNodeInfos` function (I didn't fully test this):

<details><summary>Details</summary>
<p>

```ts
export function computeNodeInfos({
  sourceCode,
  node,
}: {
  sourceCode: TSESLint.SourceCode
  node: TSESTree.Expression
}): { keyValue: string; name: string } {
  if (node.type !== AST_NODE_TYPES.ArrayExpression) {
    let text = sourceCode.getText(node)
    return { keyValue: `text:${text}`, name: text }
  }

  let [left] = node.elements

  if (!left) {
    return { keyValue: computeValueKey(undefined), name: 'undefined' }
  }

  let leftText = sourceCode.getText(left)

  let numericValue = computeExpressionNumberValue(left)
  if (numericValue !== null) {
    return {
      keyValue: computeValueKey(numericValue),
      name: leftText,
    }
  }

  switch (left.type) {
    case AST_NODE_TYPES.UnaryExpression:
      if (left.operator === 'void') {
        return {
          keyValue: computeValueKey(undefined),
          name: leftText,
        }
      }
      break
    case AST_NODE_TYPES.TemplateLiteral:
      if (left.expressions.length === 0) {
        return {
          keyValue: computeValueKey(left.quasis[0]!.value.cooked),
          name: leftText,
        }
      }
      break
    case AST_NODE_TYPES.Identifier:
      if (left.name === 'undefined') {
        return { keyValue: computeValueKey(undefined), name: left.name }
      }
      break
    case AST_NODE_TYPES.Literal:
      return {
        keyValue:
          'regex' in left ? `text:${left.raw}` : computeValueKey(left.value),
        name: left.raw,
      }
  }

  return { keyValue: `text:${leftText}`, name: leftText }
}

function computeValueKey(
  value: undefined | boolean | string | number | bigint | null,
): string {
  return `value:${typeof value}:${String(value)}`
}
```

</p>
</details>

Then, thinking out loud, I would do the following:
- Create a new `SortMapsSortingNode` object, with a `keyValue: string` field.
- When computing each `SortMapsSortingNode`, populate the `keyValue` and `name` field thanks to this function.
- After having computed all `SortMapsSortingNode`, check if two are equal. If that's the case, stop the sorting now.

I think this would keep new code to a minimum here.

hugop95 (commented):
@azat-io 

Answering [your comment](https://github.com/azat-io/eslint-plugin-perfectionist/pull/746#issuecomment-4886087435) here:

Honestly, I personally would ignore this, as we cannot fix the issue entirely:

```ts
new Map([
  [b, 1],
  [a, 0],
])
``` 

will still get sorted even if `a` is equal to `b` (which we can't know).

So we are only going to fix a few specific static cases, which I believe will never really happen in practice (I don't see a user explicitly setting a map with `16` and `0x10`).

This means that IMO, the guard gives a false sense of safety: it catches the spellings that nobody writes while missing the main case that could really occur.

> [!IMPORTANT]
> I'm not saying that we should never sort non-string/number keys: I still think that the main `sort-maps` strength is to be able to sort maps keyed by variables or enum members.
>
> So it's a case where I think that defering the responsability of the user to **not** write duplicate keys is acceptable, and better for the plugin.

---

With that being said, if you really think that this needs to be fixed, I agree that the simplest thing to do would be to drop sorting if two keys are detected as colliding.

I've made a comment that merges `computeNodeName` with `computeStaticKeyValue` in that case, as they are both quite similar.

azat-io:
Okay, agree with you 👍

PR #747, opened 2026-07-05
fix: detect dependencies in iifes, auto-accessors and computed enum access

azat-io:
### Description

The dependency detectors missed three reference forms, so the dependency-aware autofix could move a binding above its initialization, silently producing code that throws at runtime:                                            

- - references inside immediately invoked `functions ((() => b)()`, `new (function () {})())` were treated as deferred
- auto-accessor initializers (`accessor a = this.z`) were excluded from dependency computation entirely
- computed enum member access (`Enum['B']`) was invisible to the detector

---

### Pre-merge checklist


utils/is-function-immediately-executed.ts
```diff
@@ -0,0 +1,164 @@
...
+ *
+ * Covers directly invoked functions (IIFE), including functions used as the
+ * callee of a `new` expression, methods and accessors of object literals that
+ * are immediately used, and members of class expressions that are immediately
+ * used.
+ *
+ * @param functionNode - The function expression to check.
+ * @returns Whether the function is immediately executed.
+ */
+export function isFunctionImmediatelyExecuted(
```
hugop95: @azat-io 

TBH I don't know if it's worth checking this, it's already quite complex, and there are cases that aren't covered.

For example

```ts
let 
  b = 1,
  a = (function () {
    return () => b
  })()()
```

Or if the function is cast before being executed. There are likely other convoluted edge cases.

I don't really see a valid use case for such syntax!

hugop95 (commented):
@azat-io 

- OK for `sort-classes`.
- OK for `sort-enums`.

I'm wondering about the `sort-variable-declarations` and `sort-objects` use cases as the function checking if a function is immediately executed seems to be missing cases. Maybe that case is overkill?

azat-io:
@hugop95 Done ✅

hugop95 (approved):
@azat-io Nice!

PR #751, opened 2026-07-10
fix(sort-modules): detect decorator metadata dependencies

### LINKED ISSUE #750: Bug: (sorting-class) 5.10.0 broke class sorting if extending
### Describe the bug

New class sorting break `use before initialisation`

### Code example

Before: 5.9.1

```
export class WorkspaceItemConstructorParams {
    .... 
}

@InterfaceType()
export abstract class WorkspaceItem {
  @Field()
  @PropertyCreatedAt()
  createdAt: Date = new Date();

  constructor(params: WorkspaceItemConstructorParams) {
     .....
  }
}
```

After: 5.10.0

```
@InterfaceType()
export abstract class WorkspaceItem {
  @Field()
  @PropertyCreatedAt()
  createdAt: Date = new Date();

  constructor(params: WorkspaceItemConstructorParams) {
    ....
  }
}

export class WorkspaceItemConstructorParams {
  workspace!: Ref<Workspace>;
}
```

That make break at compile - using WorkingItemConstructorParams before his declaration

### ESLint version

10.6.0

5.10.0

### Additional comments

_No response_

### Validations

hugop95:
- Resolves #750.

# Description

When `emitDecoratorMetadata: true`, a decorator on a class/member makes TS emit metadata **at class-declaration time**.

That turns an otherwise-erased type annotation into a real runtime **value** reference, i.e, a genuine declaration-time dependency. 

This PR adds a `tsconfig` option, similar to [`sort-imports`](https://perfectionist.dev/rules/sort-imports#tsconfig), that reads `emitDecoratorMetadata`, and if `true`, handles those newly-detected dependencies.

## `emitDecoratorMetadata` behavior and how the rule handles them

<details><summary>Details</summary>
<p>

`B` below is the dependency;

- *✅ = emitted.*
- *❌ = not emitted.*

## ✅ Creates a dependency

```ts
@Decorator() class A { constructor(b: B) {} }         // ✅ decorated class, constructor param
@Decorator() class A { constructor(private b: B) {} } // ✅ decorated class, constructor parameter property
class A { constructor(@Decorator() b: B) {} }         // ✅ decorated param
class A { @Decorator() b: B }                         // ✅ decorated property
class A { @Decorator() accessor b: B }                // ✅ decorated accessor
class A { @Decorator() m(b: B): void {} }             // ✅ decorated method param
class A { @Decorator() m(): B {} }                    // ✅ decorated method return
class A { @Decorator() get b(): B {} }                // ✅ decorated getter return
class A { @Decorator() set b(v: B) {} }               // ✅ decorated setter param
```

### ℹ️ Siblings parameters count too

The whole parameter list is emitted.

```ts
class A { constructor(@Decorator() b: B, c: C) {} }   // ✅ decorated B and C
class A { @Decorator() m(b: B, c: C) {} }             // ✅ B and C (decorated method)
```
### ⚠️ Unhandled case (rare use case, could be kept for a V2 if truly needed)

```ts
@Decorator() class A { constructor(b: ns.B) {} }      // ❌ namespace
```

## ❌ Does NOT create a dependency

```ts
class A { constructor(b: B) {} }                      // ❌ no decorator
@Decorator() class A { m(b: B) {} }                   // ❌ method not decorated
@Decorator() class A { b: B }                         // ❌ property not decorated
class A { @Decorator() m() { return x as B } }        // ❌ type in body
class A { static { let b: B } }                       // ❌ not a class member
@Decorator() class A { constructor(b: I) {} }         // ❌ interface / type alias
```

### Non-top-level elements

```ts
@Decorator() class A { constructor(b: B[]) {} }       // ❌ array
@Decorator() class A { constructor(b: Array<B>) {} }  // ❌ generic
@Decorator() class A { constructor(b: B | C) {} }     // ❌ union
@Decorator() class A { constructor(b: B & C) {} }     // ❌ intersection
@Decorator() class A { constructor(b: [B, C]) {} }    // ❌ tuple
```

## The `strictNullChecks` edge case

`T | null` / `T | undefined` depend on the flag:

```ts
@Decorator() class A { constructor(b: B | null) {} }
```

- `strictNullChecks: false` ➡️ TS strips `null` and emits `[B]` ➡️ ✅ dependency.
- `strictNullChecks: true`  ➡️ emits `[Object]` ➡️ ❌ no dependency.
  - https://github.com/microsoft/TypeScript/issues/12703
  - https://github.com/microsoft/TypeScript/issues/18509

Same thing for `B | undefined` and `B | null | undefined`.

---

For simplicity, the rule doesn't that into account and considers things as if `strictNullChecks: true` was true: a union is simply ignored.

</p>
</details>


rules/sort-modules.ts
```diff
@@ -347,7 +351,19 @@ function analyzeModule({
...
+        readClosestTsConfigByPath({
+          tsconfigFilename: options.tsconfig.filename ?? 'tsconfig.json',
+          tsconfigRootDir: options.tsconfig.rootDir,
+          filePath: context.physicalFilename,
+          contextCwd: context.cwd,
+        })
+      : null
     let dependenciesBySortingNode = computeDependenciesBySortingNode({
+      emitDecoratorMetadata:
+        tsConfigOutput?.compilerOptions.emitDecoratorMetadata ?? false,
```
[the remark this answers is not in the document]
hugop95: @coderabbitai This is the same pattern as for sort-imports.

test/rules/sort-modules.test.ts
```diff
@@ -2580,6 +2581,630 @@ describe('sort-modules', () => {
...
+              }
+
+              class B {}
+            `,
+            options: [
+              { ...metadataOptions, useExperimentalDependencyDetection: false },
+            ],
+          })
+        })
+      })
```
azat-io, line 3562: Should we add these tests?

```ts
it('ignores a local variable annotation inside a decorated method body', async () => {
  await valid({
    code: dedent`
      class A {
        @Decorator()
        method() {
          let localVar: B
        }
      }

      class B {}
    `,
    options: [options],
  })
})

it('ignores an arrow parameter annotation inside a decorated property', async () => {
  await valid({
    code: dedent`
      class A {
        @Decorator()
        fn = (b: B) => {}
      }

      class B {}
    `,
    options: [options],
  })
})

it('ignores a function-type parameter annotation of a decorated method', async () => {
  await valid({
    code: dedent`
      class A {
        @Decorator()
        method(callback: (b: B) => void): void {}
      }

      class B {}
    `,
    options: [options],
  })
})

it('detects parameter types of a method with a decorated parameter', async () => {
  await invalid({
    errors: [
      {
        data: { nodeDependentOnRight: 'A', right: 'B' },
        messageId: 'unexpectedModulesDependencyOrder',
      },
    ],
    output: dedent`
      class B {}

      class A {
        method(@Inject() b: B): void {}
      }
    `,
    code: dedent`
      class A {
        method(@Inject() b: B): void {}
      }

      class B {}
    `,
    options: [options],
  })
})
```

hugop95: @azat-io 

Fixed, a couple of cases were missing indeed, and the enum part was indeed not accurate 👍

https://github.com/azat-io/eslint-plugin-perfectionist/pull/751/commits/1468d7d065ded5bcb898f5618930afdc3edbd8c4

azat-io (approved):

azat-io:
Thanks!

PR #756, opened 2026-07-19
fix(sort-classes): make `ignoreCallbackDependenciesPatterns` redundant and deprecate it

hugop95:
- Follow-up of https://github.com/azat-io/eslint-plugin-perfectionist/pull/755

# Description

Under experimental dependency detection, `sort-classes` treats a class member as depending on any other member it referenced inside a callback, even when that callback is deferred and never runs during initialization.

We reorder members to satisfy an issue that doesn't exist:

```ts
class User {
  b = computed(() => this.a)
  a = "something"
}
```

Here `b` is forced before `a`, even though the arrow function passed to `computed` is never invoked at init time.

The `ignoreCallbackDependenciesPatterns` option is what allows us to opt out of this behavior.

---

This PR aims to make `ignoreCallbackDependenciesPatterns` redundant:

Here is whether a member referenced **inside a deferred callback** is treated as a dependency (and thus forces a reorder):

| `ignoreCallbackDependenciesPatterns` | `useExperimentalDependencyDetection` | Before                         | After                             |
|--------------------------------------|--------------------------------------|--------------------------------|-----------------------------------|
| off                                  | `true` (default)                     | dependency ➡️ useless reorder | ignored ➡️ no reorder             |
| matches the callback                 | `true` (default)                     | ignored (manual opt-out)       | ignored (option is now redundant) |
| off                                  | `false` (legacy)                     | dependency ➡️ useless reorder | unchanged                         |
| matches the callback                 | `false` (legacy)                     | ignored (manual opt-out)       | unchanged                         |

# Note

Other dependency-aware rules (`sort-variable-declarations` and `sort-objects`) already behave this way: both use `computeDependenciesOutsideFunctionsBySortingNode`, so a reference inside a deferred callback is already ignored:

```ts
let a =
  f(() => b),
  b = 1
```

## Other changes

- Deprecate `useExperimentalDependencyDetection`: it's been enabled by default for a while and seems very stable.


azat-io (approved):

test/rules/sort-classes.test.ts
```diff
@@ -2250,6 +2250,23 @@ describe('sort-classes', () => {
       useExperimentalDependencyDetection: boolean,
```
azat-io, line 2250: Should we add these tests?

```ts
it('.map() invokes its callback synchronously — b must stay before a', async () => {
  await valid({
    code: dedent`
      class Class {
        b = 1
        a = [1, 2].map(() => this.b * 2)
      }
    `,
    options: [{ useExperimentalDependencyDetection: true }],
  })
})

it('zustand-style creator invokes its callback synchronously', async () => {

  await valid({
    code: dedent`
      class Class {
        b = createQueryString();
        a = createState((set) => {
          set('query', this.b.value);
        });
      }
    `,
    options: [
      {
        useExperimentalDependencyDetection: true,
        groups: [['property', 'method']],
      },
    ],
  })
})

it('IIFE calling an inner function declaration reads b at init time', async () => {
  await invalid({
    errors: [
      {
        messageId: 'unexpectedClassesDependencyOrder',
        data: { nodeDependentOnRight: 'a', right: 'b' },
      },
    ],
    output: dedent`
      class Class {
        static b = 1
        static a = (() => {
          function inner() {
            return Class.b
          }
          return inner()
        })()
      }
    `,
    code: dedent`
      class Class {
        static a = (() => {
          function inner() {
            return Class.b
          }
          return inner()
        })()
        static b = 1
      }
    `,
    options: [{ useExperimentalDependencyDetection: true }],
  })
})
```

hugop95: @azat-io 

Thanks for the review, I'm closing this PR in favor of https://github.com/azat-io/eslint-plugin-perfectionist/pull/761.

I realize that this PR would have lowered safety by default (without allowing users to opt-in for a securer approach), and I think we should anyway prioritize the safer approach by default first whenever it's possible and reasonable.

PR #766, opened 2026-08-18
fix: don't fail on tsconfig files without input files

azat-io:
### Description

Ignore TS18002/TS18003 tsconfig diagnostics - they only concern input files and don't affect compiler options resolution.

---

### Pre-merge checklist

### DIFF
--- a/test/utils/tsconfig/read-closest-ts-config-by-path.test.ts
+++ b/test/utils/tsconfig/read-closest-ts-config-by-path.test.ts
@@ -123,6 +123,36 @@ describe('readClosestTsConfigByPath', () => {
       )
     })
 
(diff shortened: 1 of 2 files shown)


hugop95 (approved):
@azat-io 

I have no strong opinion about this, but I initially didn't add it in https://github.com/azat-io/eslint-plugin-perfectionist/pull/759 as I found the use-case convoluted (cf [this comment](https://github.com/azat-io/eslint-plugin-perfectionist/pull/759#discussion_r3699727082)).

My rationale is that from what I understand, `tsc` would crash instantly at all in those situations (at least for 18002), so I don't expect a user to expect linting to work either.

---

But there isn't a lot of added code, so I guess it's OK as well!

PR #768, opened 2026-08-26
feat(sort-modules): add `additionalModuleBlockTypes` option

### LINKED ISSUE #767: `sort-modules` ignores `<script>` content in `.svelte` files
## Description

`sort-modules` rule does not sort top-level declarations (`type`, `function`, `class`, etc.) inside Svelte `<script>` blocks.

## Root cause

`eslint-plugin-svelte` wraps `<script>` content in a custom AST node type `SvelteScriptElement` inside `Program.body`. The `sort-modules` rule iterates `Program.body` and handles standard node types (`TSTypeAliasDeclaration`, `FunctionDeclaration`, `ClassDeclaration`, etc.), but `SvelteScriptElement` falls through to the `default: continue` branch and is silently skipped.

The only recursive descent in `analyzeModule` is for `TSModuleDeclaration` (namespace/module blocks) — there is no equivalent path for `SvelteScriptElement`.

### Code example

```svelte
<script lang="ts">
type Zebra = { a: string };
type Alpha = { b: string };
</script>
```

Run: `eslint --rule 'perfectionist/sort-modules: error' file.svelte`

**Expected:** error — "Alpha" should come before "Zebra"
**Actual:** no error

### ESLint version

v10.9.1

v5.10.1

## Environment

- `eslint-plugin-perfectionist@5.10.1`
- `eslint-plugin-svelte@5.x`
- `@sveltejs/vite-plugin-svelte@5.x`
- Parser: `svelte-eslint-parser` (via `eslint-plugin-svelte`)

## Suggested fix

In `analyzeModule`, when encountering a `SvelteScriptElement` (or any custom wrapper node from framework-specific parsers), recurse into its `.body` the same way `TSModuleDeclaration` recurses into `.moduleBlock.body`.

Alternatively, accept a configurable list of wrapper node types to recurse into.

## Workaround

Sort declarations manually in `.svelte` files — `sort-modules` only works in `.ts`/`.js`.

### Validations

hugop95:
- Resolves #767.

### Description

This PR adds a `additionalModuleBlockTypes` option to `sort-modules` so users can support additional parsers.

For Svelte for example, they can now use

```ts
{
  additionalModuleBlockTypes: ['SvelteScriptElement']
}
```

to parse `<script>` bodies.

My local tests showed that `vue-eslint-parser` and `astro-eslint-parser` already work without it.


azat-io (commented):
Thanks! The mechanism looks solid, but I'd like Svelte to work out of the box - #767 primarily asked for zero-config support, and requiring every Svelte user to discover this option feels wrong.

Proposal: keep a built-in list of known wrapper types (just `'SvelteScriptElement'` for now) that is always merged with `additionalModuleBlockTypes`, rather than making it the option's default value. That way:

- Svelte works with no configuration;
- users who set the option for another parser don't silently lose Svelte support (which would happen if it were just the default);
- the `additional...` name stays accurate - the option remains purely additive on top of the built-ins.

It's behavior-neutral for standard TS/JS ASTs since that node type can't appear there, and the check only activates when body is an array anyway.

Two smaller asks:

1. Add a test with top-level statements before and after a listed block — right now the partition boundary push (`sortingNodeGroupsWithoutOverloadSignature.push([])`) isn't pinned by any test, and it's what keeps the autofix from moving code across a `<script>` block.
2. A test with a genuinely unknown node type would exercise the real path better than BlockStatement — the wrapper-parser pattern from `sort-classes.test.ts` (`unknownClassElementParser`) fits here.

hugop95:
@azat-io

Thanks for the feedback!

> Proposal: keep a built-in list of known wrapper types (just 'SvelteScriptElement' for now) that is always merged with additionalModuleBlockTypes, rather than making it the option's default value. That way:

I was hesitant about it: my understanding was that we really wanted to get away from "officially" supporting other parsers.

And on one side, I do agree/understand that proposing the option but not making `SvelteScriptElement` a default one may look strange.

But on the other side, officially adding `SvelteScriptElement` sort-of means that we are back to "officially" adding Svelte support (as I assume this built-in support would be documented).

---

So I'm OK with supporting it out of the box, it's just something I found worth keeping in mind. Maybe this could open further requests to add support for more parsers/frameworks in the future, with its consequences on maintainability.

---

I'll do the changes by the end of the week!

hugop95:
@azat-io

- Partition-related test: https://github.com/azat-io/eslint-plugin-perfectionist/pull/768/commits/9008045ef4a1caeae062cd0b259871f8ef32da34
- Dedicated testing parser: https://github.com/azat-io/eslint-plugin-perfectionist/pull/768/commits/7fa10f31354ee312712018ce6fd2e4519d25d154
- Adding `SvelteScriptElement` by default: https://github.com/azat-io/eslint-plugin-perfectionist/pull/768/commits/88d5f7f2de7514f82f32103bb223c70cfeeb0608

azat-io (approved):

Issue #770, opened 2026-09-07
Bug: sort-modules no longer detects class dependencies in 5.11.0, demands the reverse of 5.10.1


JakobFelixJulius:
## Describe the bug

In 5.11.0, `sort-modules` no longer detects a dependency between two classes when the reference is a
constructor call inside a method body (plus a property type annotation). It falls back to plain
alphabetical order, and so reports the exact reverse of what 5.10.1 enforces for the same code.

The two versions are mutually exclusive: whichever order you write, one of them errors. That makes
the rule impossible to satisfy across a version bump, and `--fix` will move a dependency *below* its
dependent — which is a runtime hazard, not just a style change, for declarations evaluated at module
scope (TDZ `ReferenceError` on import).

Same two files, both versions, `perfectionist.configs["recommended-natural"]` with
`@typescript-eslint/parser`:

| Declaration order | 5.10.1 | 5.11.0 |
| --- | --- | --- |
| `ZebraConnection`, then `AlphaService` | clean | `Expected "AlphaService" to come before "ZebraConnection"` |
| `AlphaService`, then `ZebraConnection` | `Expected dependency "ZebraConnection" to come before "AlphaService"` | clean |

Note the wording difference — 5.10.1 reports it as a *dependency* violation, 5.11.0 as a plain
ordering one, which is what suggests dependency detection is no longer firing for this shape.

I suspect this is fallout from "sort-modules: Support custom module block nodes" (077f5d4), since the
5.11.0 release notes list no intentional change to dependency detection.

## Code example

Dependency declared first — clean on 5.10.1, errors on 5.11.0:

```ts
class ZebraConnection {
  open() {}
}

class AlphaService {
  private conn: null | ZebraConnection = null;

  start() {
    this.conn = new ZebraConnection();
    this.conn.open();
  }
}

export const service = new AlphaService();
```

```
5.10.1: clean
5.11.0: 6:1  error  Expected "AlphaService" to come before "ZebraConnection"  perfectionist/sort-modules
```

Dependent declared first (what 5.11.0's `--fix` produces) — errors on 5.10.1:

```ts
class AlphaService {
  private conn: null | ZebraConnection = null;

  start() {
    this.conn = new ZebraConnection();
    this.conn.open();
  }
}

class ZebraConnection {
  open() {}
}

export const service = new AlphaService();
```

```
[...]

PR #771, opened 2026-09-07
fix(sort-modules): detect class dependencies in type annotations

azat-io:
### Description

#770 

### Reviewer notes

N/A.

---

### Pre-merge checklist

### DIFF
--- a/test/rules/sort-modules.test.ts
+++ b/test/rules/sort-modules.test.ts
@@ -3620,6 +3620,71 @@ describe('sort-modules', () => {
       })
     })
 
+    describe('decorator metadata dependency detection without tsconfig', () => {
+      let noTsconfigOptions = {
+        ...options,
+        useExperimentalDependencyDetection: true,
+        groups: ['unknown'],
+      } as const
+
+      it('assumes emitDecoratorMetadata when no tsconfig option is given', async () => {
+        await invalid({
+          errors: [
+            {
+              data: { nodeDependentOnRight: 'A', right: 'B' },
+              messageId: 'unexpectedModulesDependencyOrder',
+            },
+          ],
+          output: dedent`
+            class B {}
+
+            class A {
+              @Decorator()
+              b: B
+            }
+          `,
+          code: dedent`
+            class A {
+              @Decorator()
+              b: B
+            }
+
+            class B {}
+          `,
+          options: [noTsconfigOptions],
+        })
+      })
+
+      it('ignores an undecorated type reference when no tsconfig option is given', async () => {
+        await valid({
+          code: dedent`
+            class A {
+              b: B
+            }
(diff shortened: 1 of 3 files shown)


test/rules/sort-modules.test.ts
```diff
@@ -2951,8 +2951,22 @@ describe('sort-modules', () => {
           })
         })
 
-        it('ignores an undecorated property type on a decorated class', async () => {
-          await valid({
+        it('detects an undecorated property type on a decorated class', async () => {
```
hugop95: @azat-io 

Cf https://github.com/azat-io/eslint-plugin-perfectionist/issues/770#issuecomment-5570132697,

But I don't see why the original code would be a problem: there is no dependency to enforce here. If we want to enforce this behavior **by default** (without having to pass `tsconfig`), then I think the test name should reflect that.

hugop95 (commented):
@azat-io 

I'll wait for https://github.com/azat-io/eslint-plugin-perfectionist/issues/770#issuecomment-5570132697 before reviewing further: I don't think these changes are justified today, as so far, I don't see why things would break here.

The purpose of https://github.com/azat-io/eslint-plugin-perfectionist/pull/751 is to accurately detect decorator-related dependencies, and avoid overdetection as much as possible.

This PR would add back some overdetection for no justifyable reason.

---

[thoughts] I haven't gone deeper on that topic, but if we want to make https://github.com/azat-io/eslint-plugin-perfectionist/pull/751 less brutal, we could instead change

https://github.com/azat-io/eslint-plugin-perfectionist/blob/4a57d49812d18a62d715d7e9d1e6795749763d9c/rules/sort-modules.ts#L392

to be `true` when no `tsconfig` option is given, until the next major version. After that, users will have to explicitly set `tsconfig.json` to remove overdetected dependencies related to decorators.

hugop95 (commented):
@azat-io 

I think that the fix that we want is to replace

https://github.com/azat-io/eslint-plugin-perfectionist/blob/4a57d49812d18a62d715d7e9d1e6795749763d9c/rules/sort-modules.ts#L391-L392

with

```ts
emitDecoratorMetadata:
  tsConfigOutput ?
    (tsConfigOutput.compilerOptions.emitDecoratorMetadata ?? false)
  : true,
```

---

This way:
- We don't overdetect things when it's not needed (see your 4 tests failing after that).
- When a user forgets to set the `tsconfig` option when they depend on decorator dependencies, we still sort them correctly.
-  We keep things simple: there is only a single change to do implementation-wise.

azat-io:
@hugop95 Fixed. Rewrote the PR to your suggestion

hugop95 (approved):
@azat-io 

Thanks 🙏

PR #773, opened 2026-09-08
perf: build a sortable list's fix once instead of once per report

azat-io:
### Description

`reportErrors` attached a `fix` callback to every reported pair, and each one rebuilt the fix for the whole list. ESLint calls those eagerly and retains the merged string per message, so `n` misplaced elements cost `O(n)` reports × `O(n)` work and `O(n²)` retained characters — on a plain `eslint` run, no `--fix`.

Now the fix is built at most once per list, lazily, and merged into the single `{ range, text }` ESLint would have built anyway. `cloneFix` copies `text` by reference, so retained text drops to `O(n)`.

Results - `sort-objects`, reversed N-key object

| N | time | heap |
|---|---|---|
| 1000 | 891 → **159 ms** | 177 → **76 MB** |
| 2000 | 3728 → **431 ms** | 593 → **56 MB** |
| 5000 | OOM → **2144 ms** | — → **69 MB** |

`sort-switch-case` at N=800: 567 → **82 ms**.

### Reviewer notes

N/A.

---

### Pre-merge checklist


hugop95 (approved):
🚀

PR #774, opened 2026-09-09
fix: honor eslint-disable directives with a description

azat-io:
### Description

Fixes handling of `--` descriptions in ESLint directives so autofixes preserve the positions of disabled elements.

### Reviewer notes

N/A.

---

### Pre-merge checklist


hugop95 (approved):
@azat-io 

Thanks!

I understand the idea behind only adding tests to 2 rules, it would be quite tedious to update all rules tests.

A part of me still thinks that it contributes to a progressive drift and accumulation of discrepancies between rule tests, with some rules testing common behaviors and others not, and I wonder if we shouldn't at some point go over all rules and unify them as much as possible (common test wordings, unified ordering, `describe` structures when testing options that are common to multiple rules, etc).

azat-io:
Thank you! Added more tests 👍

PR #788, opened 2026-09-19
refactor: share context option matching and comparators

azat-io:
### Description

Moves the context option check that eight rules duplicated and the name/value comparator shared by sort-objects and sort-object-types into shared utils, and builds the four string comparators from one `buildStringComparatorByOptionsComputer`. Behavior is unchanged.

### Reviewer notes

N/A.

---

### Pre-merge checklist


hugop95 (approved):
@azat-io Nice!

